PR #1, opened 2022-04-25
Implement basic enc, decoding

A6GibKm:
I am not very happy with having the decrypt in item.rs and the decrypt in secret.rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,15 @@ tokio = {version = "1.17", features = ["macros", "rt-multi-thread"]}
 futures = "0.3"
 rand = {version = "0.8", default-features = false}
 
+# Crypto
+aes = "0.8"
+cbc = "0.1"
+cipher = {version = "0.4", features = ["rand_core"]}
+hmac = "0.12"
+serde_json = "1.0"
+sha2 = "0.10"
+zeroize = "1.5"
+
 [[bin]]
 name = "basic"
 path = "examples/basic.rs"
--- a/src/dbus/api/secret.rs
+++ b/src/dbus/api/secret.rs
@@ -28,7 +28,7 @@ impl<'a> Secret<'a> {
     ) -> Self {
         let (value, parameters) = match algorithm.as_ref() {
             Algorithm::Plain => (secret.to_vec(), vec![]),
-            Algorithm::Dh(aes_key) => utils::encrypt(secret, aes_key).unwrap(),
+            Algorithm::Encrypted(aes_key) => utils::encrypt(secret, aes_key).unwrap(),
         };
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -11,10 +11,19 @@ serde = "1.0"
...
+zeroize = "1.5"
 
 [[bin]]
 name = "basic"
 path = "examples/basic.rs"
 
 [dev-dependencies]
 serde_json = "1.0"
+
+[dependencies.cipher]
```
bilelmoussaoui, line 27: I would inline it like the other dependencies above

src/utils.rs
```diff
@@ -0,0 +1,55 @@
+use aes::cipher::{
+    block_padding::Pkcs7, BlockDecryptMut, BlockEncryptMut, BlockSizeUser, KeyIvInit,
+};
+use hmac::{digest::OutputSizeUser, Mac};
+use zeroize::Zeroizing;
+
+use crate::error::Result;
```
bilelmoussaoui, line 7:
```suggestion
use crate::Result;
```

src/dbus/api/secret.rs
```diff
@@ -26,12 +26,9 @@ impl<'a> Secret<'a> {
...
     ) -> Self {
-        let (parameters, value) = match algorithm.as_ref() {
-            Algorithm::Plain => (vec![], secret.to_vec()),
-            Algorithm::Dh(_blob) => {
-                // See https://github.com/hwchen/secret-service-rs/blob/d6aaa774f0ec504ff5f26662279e07175b8ef111/src/util.rs#L52
-                unimplemented!()
-            }
+        let (value, parameters) = match algorithm.as_ref() {
+            Algorithm::Plain => (secret.to_vec(), vec![]),
+            Algorithm::Dh(aes_key) => utils::encrypt(secret, aes_key).unwrap(),
```
bilelmoussaoui, line 31: The key should no longer be stored inside Dh variant, I would also maybe just rename it to Encrypted  and we could document what encryption it uses in the future. The key then should be generated and stored as part of src/dbus/service.rs and passed to the src::api::Session when created.

Cargo.toml
A6GibKm: Done.

src/dbus/api/secret.rs
A6GibKm: Renamed to Encrypted, without a inner field. With this one can drop all the `Arc`s required. The key should be retrievable from the session right?

bilelmoussaoui: I moved it to src/dbus, so it probably needs a rebase here

PR #2, opened 2022-04-25
Add create_item test

A6GibKm:


src/dbus/collection.rs
```diff
@@ -176,3 +176,28 @@ impl<'a> Collection<'a> {
...
+        let secret = "a password".as_bytes();
+
+        let collection = service.default_collection().await.unwrap().unwrap();
+        let item = collection
+            .create_item("A secret", attributes, secret, true, "text/plain")
+            .await
+            .unwrap();
+
+        assert_eq!(item.secret().await.unwrap(), secret);
+        assert_eq!(item.attributes().await.unwrap()["type"], "token");
```
bilelmoussaoui, line 201: Ideally we would remove the item afterwards to keep the CI service intact to not influence the results of other tests in the future.

bilelmoussaoui:
Otherwise looks good to me, we can merge it and I will look into making the CI work properly

PR #3, opened 2022-04-25
keyring: Add support for keyring storage

sophie-h:
The format is used by libsecret inside of flatpaks.
This patch adds a general keyring file format API
as well as convencience functions to store/lookup/delete
passwords that are encrypted via a secret from the
Secret portal.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,11 +6,24 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-zbus = "2.1"
+zbus = {version = "2.1", features = ["gvariant"]}
 serde = "1.0"
 tokio = {version = "1.17", features = ["macros", "rt-multi-thread"]}
 futures = "0.3"
 rand = {version = "0.8", default-features = false}
+async-std = {version = "1.11", features = ["attributes"]}
+byteorder = "1.0"
+
+aes = "0.8"
+cbc = "0.1"
+cipher = {version = "0.4", features = ["rand_core"]}
+digest = "0.10"
+hmac = "0.12"
+pbkdf2 = "0.11"
+sha2 = "0.10"
+zeroize = {version = "1.5", features = ["zeroize_derive"]}
+
+dirs = "4.0"
 
 [features]
 default = ["local_tests"]
@@ -21,6 +34,3 @@ local_tests = []
 [[bin]]
 name = "basic"
 path = "examples/basic.rs"
-
-[dev-dependencies]
-byteorder = "1.0"
--- a/src/keyring.rs
+++ b/src/keyring.rs
@@ -1,0 +1,102 @@
+/*!
+Keyring
+
+```ignore
+# std::env::set_var("XDG_DATA_HOME", "/tmp/doctest");
+# use oo7::keyring::{self, Error};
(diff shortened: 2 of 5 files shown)


src/keyring.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Times are stored before reading the file to detect
+    /// file changes before writing
+    mtime: Option<std::time::SystemTime>,
+    key: api::Key,
+}
+
+impl Storage {
+    /// Load from default keyring file
+    pub async fn load_default() -> Result<Self> {
+        // TODO: use secret api here
```
bilelmoussaoui, line 68: Is this TODO comment still needed?

src/keyring.rs
```diff
@@ -0,0 +1,103 @@
...
+        Ok(Storage {
+            keyring,
+            path: path.as_ref().to_path_buf(),
+            mtime,
+            key,
+        })
+    }
+
+    pub async fn write(self) -> Result<()> {
+        self.keyring.dump(self.path, self.mtime).await
```
bilelmoussaoui, line 101: Shouldn't mtime be updated when the keyring is dumped ?

src/keyring/api.rs
```diff
@@ -0,0 +1,500 @@
...
+const MAJOR_VERSION: u8 = 1;
+const MINOR_VERSION: u8 = 0;
+
+type MacAlg = hmac::Hmac<sha2::Sha256>;
+type EncAlg = cbc::Encryptor<aes::Aes128>;
+type DecAlg = cbc::Decryptor<aes::Aes128>;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug)]
```
bilelmoussaoui, line 44: I would move the Error type into a separate file. along with the keyring stuff to the portal module

src/keyring/api.rs
```diff
@@ -0,0 +1,500 @@
...
+pub struct Keyring {
+    salt_size: u32,
+    salt: Vec<u8>,
+    iteration_count: u32,
+    modified_time: u64,
+    pub usage_count: u32,
+    pub items: Vec<EncryptedItem>,
+}
+
+impl Keyring {
```
bilelmoussaoui, line 100: Would split each item into a separate file, to make the code more readable

Cargo.toml
```diff
@@ -6,15 +6,25 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-zbus = "2.1"
+zbus = {version = "2.1", features = ["gvariant"]}
 serde = "1.0"
 tokio = {version = "1.17", features = ["macros", "rt-multi-thread"]}
 futures = "0.3"
 rand = {version = "0.8", default-features = false}
+async-std = {version = "1.11", features = ["attributes"]}
```
bilelmoussaoui, line 14: I would appreciate if you could drop the tokio dependency and replace the tokio::test calls with async_std::test.

sophie-h: Will gladly do it. But other PR, because this one is huge already and I want to touch as little of the existing files with this one.

tests/keyring.rs
```diff
@@ -0,0 +1,66 @@
+use oo7::keyring;
```
bilelmoussaoui, line 1: Those tests don't seems to run

PR #6, opened 2022-04-26
start Implementing basic support for encrypted dbus

A6GibKm:


src/dbus/algorithm.rs
```diff
@@ -1,29 +1,29 @@
...
-                    array
-                        .append(zvariant::Value::U8(*byte))
-                        .expect("Element of valid type");
-                }
-                array.into()
-            }
-        }
-    }
-}
+// impl Algorithm {
```
bilelmoussaoui, line 11: No longer needed?

src/dbus/item.rs
```diff
@@ -128,7 +131,16 @@ impl<'a> Item<'a> {
             Err(Error::Deleted)
         } else {
             let secret = self.inner.secret(&self.session).await?;
-            Ok(secret.value)
+
+            let value = match self.algorithm {
+                Algorithm::Plain => secret.value,
+                Algorithm::Encrypted => {
+                    let iv = secret.parameters;
+                    let aes_key = self.aes_key.as_ref().unwrap();
```
bilelmoussaoui, line 139: Should keep a note here why unwrap is safe (because it is encrpyted), same for the decrypt call. Which i believe shouldn't have an unwrap

src/dbus/service.rs
```diff
@@ -18,24 +18,35 @@ use crate::Result;
...
 }
 
 impl<'a> Service<'a> {
     /// Create a new instance of the Service.
     pub async fn new(algorithm: Algorithm) -> Result<Service<'a>> {
         let cnx = zbus::Connection::session().await?;
         let service = Arc::new(api::Service::new(&cnx).await?);
-        let (service_key, session) = service.open_session(&algorithm).await?;
+
+        let private_key = Key::generate_private_key();
```
bilelmoussaoui, line 32: You don't need this if it Algorithm::Plain, so I would move it to inside the match and make it return a tuple (public_key, private_key)

src/dbus/utils.rs
```diff
@@ -0,0 +1,51 @@
...
+};
+use hmac::{digest::OutputSizeUser, Mac};
+
+use crate::{error::Result, Key};
+
+type MacAlg = hmac::Hmac<sha2::Sha256>;
+type EncAlg = cbc::Encryptor<aes::Aes128>;
+type DecAlg = cbc::Decryptor<aes::Aes128>;
+
+pub fn encrypt(data: &[u8], key: &Key) -> Result<(Vec<u8>, Vec<u8>)> {
```
bilelmoussaoui, line 12:
```suggestion
pub(crate) fn encrypt(data: &[u8], key: &Key) -> Result<(Vec<u8>, Vec<u8>)> {
```

src/key.rs
```diff
@@ -15,3 +37,111 @@ impl AsMut<[u8]> for Key {
...
+impl Key {
+    pub(crate) fn generate_private_key() -> Key {
+        let mut rng = OsRng {};
+        let mut private_key = Key(vec![0_u8; 128]);
+        rng.fill(private_key.as_mut());
+
+        private_key
+    }
+
+    // TODO zeroize
```
bilelmoussaoui, line 50: Is this still a thing?

src/key.rs
```diff
@@ -15,3 +37,111 @@ impl AsMut<[u8]> for Key {
         &mut self.0
     }
 }
+
+impl Key {
+    pub(crate) fn generate_private_key() -> Key {
```
bilelmoussaoui, line 42:
```suggestion
    pub(crate) fn generate_private_key() -> Self {
```

src/key.rs
A6GibKm: Yes, a bunch of operations which should be zeroized, but I think the entire thing should go away, right now it is copy paste from secret-service-rs

bilelmoussaoui: But you are already deriving Zerioize here, but in general we would need for things like Secret/(a potential generic Password wrapper)

A6GibKm: Yes and no. Note that we are turning our safe stuff into biguint and I dont think they take these subtleties into account.

src/key.rs
```diff
@@ -15,3 +38,129 @@ impl AsMut<[u8]> for Key {
...
+        array.into()
+    }
+
+    pub(crate) fn from_value(value: &zvariant::OwnedValue) -> Self {
+        let mut key = zeroize::Zeroizing::new(vec![]);
+        for value in value.downcast_ref::<zvariant::Array>().unwrap().get() {
+            key.push(*value.downcast_ref::<u8>().unwrap());
+        }
+        Key(key.to_vec())
+    }
```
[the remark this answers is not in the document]
A6GibKm: It is not clear to me when From/Into are more appropriate (we generally specify them as `fn from_x` in bindings, but thats probably an artifact of being bindings) Ill turn the from_value to a From though.

PR #7, opened 2022-04-26
meta: Add unstable feature

sophie-h:


sophie-h:
This should need nightly for docs to build

bilelmoussaoui:
Could you switch the docs CI to use nightly instead of stable?

bilelmoussaoui:
> Could you switch the docs CI to use nightly instead of stable?

I will handle it in a future commit actually, as I want to mark the dbus::api as unstable as well.

PR #11, opened 2022-04-27
portal: Do not fail when writing repeatedly

sophie-h:


src/portal/api/mod.rs
```diff
@@ -161,6 +167,24 @@ impl Keyring {
             .collect()
     }
 
+    pub fn lookup_item(
```
bilelmoussaoui, line 170: Would add a one line doc to this function

bilelmoussaoui: Does it make sense to provide a similar function to the oo7::Keyring type as well?

bilelmoussaoui:
I will go ahead and merge this, we could land more docs improvements in the future.

PR #17, opened 2022-06-02
is_sandboxed: Check if app is a running under snap

A6GibKm:
See https://github.com/bilelmoussaoui/ashpd/pull/65


src/utils.rs
```diff
@@ -0,0 +1,80 @@
+pub(crate) async fn is_flatpak() -> bool {
+    #[cfg(feature = "async-std")]
+    let b = async_std::path::PathBuf::from("/.flatpak-info")
+        .exists()
+        .await;
+    #[cfg(not(feature = "async-std"))]
+    let b = std::path::PathBuf::from("/.flatpak-info").exists();
+
+    b
```
bilelmoussaoui, line 9:
```suggestion
    #[cfg(feature = "async-std")]
    async_std::path::PathBuf::from("/.flatpak-info")
        .exists()
        .await
    #[cfg(not(feature = "async-std"))]
    std::path::PathBuf::from("/.flatpak-info").exists()
```

src/utils.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+pub(crate) async fn is_snap() -> bool {
+    #[cfg(feature = "async-std")]
+    use async_std::{fs::File, prelude::*};
+    #[cfg(all(feature = "tokio", not(feature = "async-std")))]
+    use tokio::{fs::File, io::AsyncReadExt};
```
bilelmoussaoui, line 16: Move those to the top of the file please.

bilelmoussaoui: Also rename this file to helpers instead

src/portal/secret.rs
```diff
@@ -99,6 +99,12 @@ impl<'a> SecretProxy<'a> {
 }
 
 pub async fn retrieve() -> Result<Vec<u8>, Error> {
+    #[cfg(not(feature = "async-std"))]
+    use std::{io::Read, os::unix::net::UnixStream};
+
+    #[cfg(feature = "async-std")]
+    use async_std::{os::unix::net::UnixStream, prelude::*};
```
bilelmoussaoui, line 106: Move to the top of the file please

src/utils.rs
A6GibKm: Compiler hates it.

src/dbus/api/secret.rs
```diff
@@ -10,12 +11,15 @@ use crate::{crypto, dbus::Error, Key};
 #[zvariant(signature = "(oayays)")]
 pub(crate) struct SecretInner(pub OwnedObjectPath, pub Vec<u8>, pub Vec<u8>, pub String);
 
-#[derive(Debug, Type)]
+#[derive(Debug, Type, Zeroize, ZeroizeOnDrop)]
 #[zvariant(signature = "(oayays)")]
 pub struct Secret<'a> {
+    #[zeroize(skip)]
     pub(crate) session: Arc<Session<'a>>,
+    #[zeroize(skip)]
```
bilelmoussaoui, line 19: I would keep this zeroized

src/utils.rs
bilelmoussaoui: use {} around those then

Cargo.toml
```diff
@@ -17,9 +17,10 @@ serde = "1.0"
 futures = {version = "0.3", default-features = false, features = ["std", "async-await"]}
 rand = {version = "0.8", default-features = false}
 async-std = {version = "1.11", features = ["attributes"], optional = true}
```
bilelmoussaoui, line 19: should async-std get a default-features=false as well?

bilelmoussaoui:
otherwise lgtm

Cargo.toml
A6GibKm: That would be nice, but we get a lot of errors like
```
  --> src/portal/mod.rs:37:17
   |
37 | use async_std::{fs, io, prelude::*, sync::Mutex};
   |                 ^^  ^^  ^^^^^^^     ^^^^ could not find `sync` in `async_std`
   |                 |   |   |
   |                 |   |   could not find `prelude` in `async_std`
   |                 |   no `io` in the root
   |                 no `fs` in the root
```

ill check if we can set the explicitly

PR #18, opened 2022-06-07
portal: Only derive key if needed

sophie-h:
This allows for searches that find nothing or number of items
operations to complete without deriving the key.

I think with this one I wouldn't need the unstable feature anymore.

### DIFF
--- a/src/portal/mod.rs
+++ b/src/portal/mod.rs
@@ -28,7 +28,6 @@
 //! # }
 //! ```
 
-use once_cell::sync::OnceCell;
 use std::{
     collections::HashMap,
     path::{Path, PathBuf},
@@ -37,10 +36,11 @@ use std::{
 
 #[cfg(feature = "async-std")]
 use async_std::{fs, io, prelude::*, sync::Mutex};
+use futures::{stream, StreamExt, TryStreamExt};
+use once_cell::sync::OnceCell;
 #[cfg(feature = "tokio")]
 #[cfg(not(feature = "async-std"))]
 use tokio::{fs, io, io::AsyncReadExt, sync::Mutex};
-
 #[cfg(feature = "unstable")]
 pub mod api;
 #[cfg(not(feature = "unstable"))]
@@ -107,8 +107,6 @@ impl Keyring {
             }
         };
 
-        //let key = Lazy::new(|| keyring.derive_key(secret));
-
         Ok(Self {
             keyring: Mutex::new(keyring),
             path: path.as_ref().to_path_buf(),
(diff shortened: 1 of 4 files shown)


sophie-h:
Not really finished but wanted to have an opinion first

src/portal/mod.rs
```diff
@@ -104,48 +107,65 @@ impl Keyring {
...
-            key,
+            secret: Zeroizing::new(secret.to_vec()),
+            key: OnceCell::new(),
         })
     }
 
+    /// Retrieve the number of items
+    ///
+    /// This function does not require for a key to be derived.
+    pub async fn num_items(&self) -> usize {
```
bilelmoussaoui, line 124: Would name it `n_items`

src/portal/mod.rs
```diff
@@ -218,6 +238,48 @@ impl Keyring {
...
+                key
+            } else {
+                let keyring = self.keyring.lock().await;
+                self.key.get_or_init(|| keyring.derive_key(&self.secret))
+            }
+        })
+    }
+}
+
+impl KeyExt for &Keyring {
```
bilelmoussaoui, line 258: Do we need to provide an impl for `Keyring` & `&Keyring`? As this is only used internally, I guess not?

src/portal/mod.rs
```diff
@@ -218,6 +238,48 @@ impl Keyring {
...
+                key
+            } else {
+                let keyring = self.keyring.lock().await;
+                self.key.get_or_init(|| keyring.derive_key(&self.secret))
+            }
+        })
+    }
+}
+
+impl KeyExt for crate::Key {
```
bilelmoussaoui, line 271: Same here

src/portal/mod.rs
bilelmoussaoui: Should the dbus side of things have a similar method? if so, please add it and add it to wrapper keyring (src/keyring)

src/portal/mod.rs
```diff
@@ -218,6 +238,48 @@ impl Keyring {
...
+
+impl KeyExt for Keyring {
+    fn get<'a>(&'a self) -> Pin<Box<dyn std::future::Future<Output = &crate::Key> + 'a>> {
+        Box::pin(async {
+            if let Some(key) = self.key.get() {
+                key
+            } else {
+                let keyring = self.keyring.lock().await;
+                self.key.get_or_init(|| keyring.derive_key(&self.secret))
+            }
```
bilelmoussaoui, line 253: Can't this be a self.key.get_or_init directly without the whole condition?

bilelmoussaoui (commented):
Looks good to me otherwise, thanks!

src/portal/mod.rs
bilelmoussaoui: I see. That is not great but it is fine, could use a comment why it is done this to avoid future you or me trying to change it :P

src/portal/mod.rs
```diff
@@ -104,48 +107,65 @@ impl Keyring {
...
+            key: OnceCell::new(),
         })
     }
 
+    /// Force to derive key
+    ///
+    /// Usually the key is only derived when required. This usually happens
+    /// when optaining in element. With this function it will already be
+    /// derived and cached beforehand.
+    pub async fn force_dervice_key(&self) {
```
bilelmoussaoui, line 124: Not really sure we want this api to be public?

bilelmoussaoui: At this point, I would make KeyExt part of the unstable feature so you could call it without having to expose this function

sophie-h: I don't need it. But some apps might want to lookup a key quickly while running and therefore might want to do the derivation on startup. That was my idea behind this function.

bilelmoussaoui: I see. Well unless people need more APIs I wouldn't add them because removing them is harder :P

A6GibKm:
See if https://github.com/bilelmoussaoui/oo7/pull/19 helps with the deadlocks.

src/portal/mod.rs
```diff
@@ -218,6 +238,29 @@ impl Keyring {
     }
 }
 
+pub trait KeyExt {
```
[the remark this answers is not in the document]
bilelmoussaoui: I don't think it is worth it

PR #21, opened 2022-07-25
portal: Only derive key if needed - take 2

sophie-h:
Replaces #18


sophie-h:
I don't think there was a reason to switch to RwLock, to begin with. Mutex is usually faster than RwLock.

sophie-h:
But we can also stick with rwlock

sophie-h:
I will just revert to RwLock and everyone is happy.

A6GibKm:
My reasoning for using RwLock is that the api is nicer and that we do multiple reads but only a few writes. It might be possible that RwLock has too big of an overhead for the small amount of operations that we make use  but I have no clue if thats the case.

A6GibKm:
I don't have a strong opinion on the matter, but  RwLock vs Mutex is outside of the scope of this PR as far as I can tell.

bilelmoussaoui:
> I don't have a strong opinion on the matter, but RwLock vs Mutex is outside of the scope of this PR as far as I can tell.

I agree, it would be faster to review one thing at a time and land it. I would like to do a new beta release to push this crate a bit forward.

sophie-h:
> RwLock vs Mutex is outside of the scope of this PR as far as I can tell.

Well, your change was based on your assumption that it would make the feature in this PR possible. So I reverted it in the implementation that does not need it.

sophie-h:
Ready to go from my side

src/portal/secret.rs
```diff
@@ -10,9 +10,29 @@ use rand::{distributions::Alphanumeric, thread_rng, Rng};
 #[cfg(feature = "tokio")]
 use tokio::{io::AsyncReadExt, net::UnixStream};
 use zbus::zvariant::{Fd, ObjectPath, OwnedValue, SerializeDict, Type};
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
 use super::Error;
 
+#[derive(Debug, Zeroize, ZeroizeOnDrop)]
+pub struct Secret {
```
bilelmoussaoui, line 18: As Secret is now part of the public API, maybe it needs to be re-exported in src/portal/mod.rs? or event src/lib.rs?

bilelmoussaoui: Also could use a small comment about what this type represents

src/portal/mod.rs
```diff
@@ -103,48 +106,60 @@ impl Keyring {
...
-            key,
+            key: Default::default(),
+            secret,
         })
     }
 
+    /// Retrieve the number of items
+    ///
+    /// This function will not trigger a key derivation.
+    pub async fn n_items(&self) -> usize {
```
bilelmoussaoui, line 121: I would be interested in having this implemented also for the dbus implementation & exposed in the "common" generic API if possible.

src/portal/mod.rs
```diff
@@ -217,18 +236,26 @@ impl Keyring {
     }
 }
 
+/// Return key, derive and store it first if not initialized
+async fn derive_key<'a>(
+    key: &'a mut OnceCell<crate::Key>,
+    keyring: &api::Keyring,
+    secret: &[u8],
```
bilelmoussaoui, line 243:
```suggestion
    secret: &Secret,
```
is more accurate no?

src/portal/secret.rs
sophie-h: > maybe it needs to be re-exported in src/portal/mod.rs

It is re-exported in `src/portal/mod.rs` already. I added docs to it.

src/portal/mod.rs
sophie-h: changed

src/portal/mod.rs
sophie-h: Anyways, this PR only touches the `portal` part anyways so the rest can be done in a different PR.

PR #22, opened 2022-09-21
portal: Add ability to modify/remove by index

sophie-h:
This feature is handy for key managers that edit entries

### DIFF
--- a/src/portal/error.rs
+++ b/src/portal/error.rs
@@ -29,7 +29,10 @@ pub enum Error {
     CancelledPortalRequest,
     /// If the portal is not available on the host.
     /// Can happen if the host has an old xdg-desktop-portal
+    /// or no secret service is available to store the secret.
     PortalNotAvailable,
+    /// The addressed index does not exist.
+    InvalidItemIndex(usize),
 }
 
 impl From<zvariant::Error> for Error {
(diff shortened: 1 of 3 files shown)


sophie-h:
Still have to test the delete function with Key Rack

bilelmoussaoui:
Could use a rebase

sophie-h:
> Could use a rebase

I mean yeah, it's based on the mentioned PR and rebases just fine on main now

src/portal/error.rs
```diff
@@ -79,7 +82,37 @@ impl std::fmt::Display for Error {
...
+#[derive(Debug)]
+/// All information that is available about an invalid (not decryptable)
+/// [`Item`](super::Item)
+pub struct InvalidItemError {
+    error: Error,
+    attribute_names: Vec<String>,
+}
+
+impl InvalidItemError {
+    pub fn new(error: Error, attribute_names: Vec<String>) -> Self {
```
bilelmoussaoui, line 100: Should the constructor be public?

src/portal/mod.rs
```diff
@@ -195,6 +206,44 @@ impl Keyring {
         self.write().await
     }
 
+    /// Replaces item at the given index.
+    ///
+    /// The `index` refers to the index of the [`Vec`] returned by
+    /// [`items()`](Self::items). If the index does not exist, the functions
+    /// returns an error.
+    pub async fn replace_item_index(&self, index: usize, item: &Item) -> Result<(), Error> {
```
bilelmoussaoui, line 214: Not a fan of the name of those two methods but I don't have a better suggestion neither

sophie-h: Me neither

bilelmoussaoui (approved):

PR #23, opened 2022-09-23
portal: Spawn key derivation blocking

sophie-h:
This one depends on merging #21 and #22

### DIFF
--- a/src/portal/mod.rs
+++ b/src/portal/mod.rs
@@ -31,6 +31,7 @@
 use std::{
     collections::HashMap,
     path::{Path, PathBuf},
+    sync::Arc,
 };
 
 #[cfg(feature = "async-std")]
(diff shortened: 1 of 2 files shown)


src/portal/mod.rs
```diff
@@ -283,24 +284,45 @@ impl Keyring {
...
+        let newkey = async_global_executor::spawn_blocking(move || {
+            async_std::task::block_on(async { keyring.read().await.derive_key(&secret) })
+        })
+        .await;
+
+        #[cfg(feature = "tokio")]
+        let newkey =
+            tokio::task::spawn_blocking(move || keyring.blocking_read().derive_key(&secret))
+                .await
+                .unwrap();
```
bilelmoussaoui, line 303: is it safe to unwrap here?

sophie-h: I mean it fails if the thread panics which would probably be on out of memory or whatever so I guess it's the best thing we can do?

bilelmoussaoui: I would do an expect then, just to point out to the correct place to check in case of a potential future bug report

sophie-h: I mean, unwraps always print the filename and line, right? Is there a reason this needs to be an expect and the other 56 unwraps not?

```
oo7 $ git g 'unwrap()' | wc -l
57
```

bilelmoussaoui (approved):

PR #25, opened 2022-09-29
portal: Don't encrypt if we deem key to be weak

sophie-h:
Minima are the defaults used by libsecert.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -9,7 +9,7 @@ use sha2::Sha256;
 use zbus::zvariant::{self, Type};
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
-use crate::crypto::EncAlg;
+use crate::{crypto::EncAlg, portal};
 
 // for key exchange
 static DH_GENERATOR: Lazy<BigUint> = Lazy::new(|| BigUint::from_u64(0x2).unwrap());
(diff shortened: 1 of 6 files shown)


src/portal/error.rs
```diff
@@ -116,3 +131,25 @@ impl std::fmt::Display for InvalidItemError {
         )
     }
 }
+#[derive(Debug, Clone)]
+pub enum WeakKeyError {
```
bilelmoussaoui, line 135: Could use a bit of documentation, at least the type itself

src/portal/item.rs
```diff
@@ -93,6 +93,8 @@ impl Item {
     }
 
     pub(crate) fn encrypt(&self, key: &Key) -> Result<EncryptedItem, Error> {
+        key.strength()?;
```
bilelmoussaoui, line 96:
```suggestion
        key.check_strength()?;
```
or something similar maybe?

src/key.rs
```diff
@@ -29,24 +29,43 @@ static DH_PRIME: Lazy<BigUint> = Lazy::new(|| {
...
 
 impl AsMut<[u8]> for Key {
     fn as_mut(&mut self) -> &mut [u8] {
-        &mut self.0
+        &mut self.key
     }
 }
 
 impl Key {
+    pub fn new(key: Vec<u8>) -> Self {
```
bilelmoussaoui, line 51: Does it make sense to keep this public?

src/key.rs
```diff
@@ -29,24 +29,43 @@ static DH_PRIME: Lazy<BigUint> = Lazy::new(|| {
...
+        Self::new_with_strength(key, Err(portal::WeakKeyError::StrengthUnknown))
+    }
+
+    pub fn strength(&self) -> Result<(), portal::WeakKeyError> {
```
bilelmoussaoui, line 55: Should this be public?

bilelmoussaoui:
I took the liberty to rebase this branch btw

src/portal/error.rs
sophie-h: Added the reasons why we do those checks as well.

src/portal/item.rs
sophie-h: Yeah, that explains more about what it's for.

src/key.rs
sophie-h: (crate) now

bilelmoussaoui (approved):

PR #32, opened 2023-01-08
Allow using openssl crate for cryptographic primitives

ueno:
This allows oo7 to selectively use the openssl crate instead of the
RustCrypto crates, when `openssl_crypto` feature is enabled.  The
motivation behind this is to make it FIPS 140 compliant by delegating
the cryptographic primitives implementation to the FIPS-certified
system library (i.e., libcrypto.so in OpenSSL).


bilelmoussaoui:
Thank you for your PR, although I think we should still have an optional feature for opting-out of OpenSSL instead of dropping the current code completely

ueno:
Yes, I agree (pulling in openssl may also mean a license change to ASL-2.0). Let me add a build-time feature.

ueno:
@bilelmoussaoui I think the CI should pass now; could you retrigger the workflow?

bilelmoussaoui (approved):

bilelmoussaoui:
lgtm, thanks a lot!

PR #37, opened 2023-06-05
Use std::sync::OnceLock

A6GibKm:
Its too early to merge this, so ill mark as draft, at least until around the time the dependency is not needed  by gtk-rs-core.


bilelmoussaoui:
Any chance we could land this?

bilelmoussaoui:
No longer needed, thanks nonetheless

PR #44, opened 2023-10-15
Use chrono::DateTime for secrets

A6GibKm:
I am not sure if adding another dependency is desired.

Its somewhat weird to return the number of seconds since the unix epoch in anything but as a date.

### DIFF
--- a/src/portal/item.rs
+++ b/src/portal/item.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, time::Duration};
 
 use serde::{Deserialize, Serialize};
 use zbus::zvariant::{self, Type};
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
I am not a fan of depending on chrono for this very small bit; using Duration is fine and people can convert it to a date using whatever library they are using

A6GibKm:
Ok then ill add the methods as duration

A6GibKm:
Done.

src/portal/item.rs
```diff
@@ -88,6 +88,18 @@ impl Item {
         self.secret = secret.as_ref().to_vec();
     }
 
+    /// The UNIX time when the item was created.
+    pub fn created(&self) -> Duration {
```
bilelmoussaoui, line 92: Can't we have a Duration directly in the struct?

A6GibKm: Those are serialized/deserialized, no idea tbh

bilelmoussaoui: Right, Duration is serialized as a `{"secs": 10, "nanos": 20}` but implementing a custom Serialize/Deserialize for Item should be pretty easy

bilelmoussaoui: Let us land this as is for now

bilelmoussaoui (approved):

Issue #48, opened 2023-11-14
Re-implement `secret-tool`


bilelmoussaoui:
It is a binary provided by libsecret

```
usage: secret-tool store --label='label' attribute value ...
       secret-tool lookup attribute value ...
       secret-tool clear attribute value ...
       secret-tool search [--all] [--unlock] attribute value ...
       secret-tool lock --collection='collection'
```

It should be easy to write something similar in Rust using oo7's dbus API.

PR #50, opened 2023-11-15
Add keyring-util

A6GibKm:
Fixes: https://github.com/bilelmoussaoui/oo7/issues/48

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -16,8 +16,14 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
-      - uses: actions-rs/cargo@v1
+      - name: Check
+        uses: actions-rs/cargo@v1
         with:
+          command: check
+      - name: Check (keyring-util)
+        uses: actions-rs/cargo@v1
+        with:
+          working-directory: ./keyring-util
           command: check
 
   test:
@@ -40,6 +46,10 @@ jobs:
       - name: Build and test (OpenSSL)
         run: |
           dbus-run-session -- cargo test --no-default-features --features async-std --features openssl_crypto
+      - name: Build (keyring-util)
+        working-directory: ./keyring-util
+        run: |
+          cargo build
 
   fmt:
     name: Rustfmt
@@ -52,9 +62,16 @@ jobs:
           toolchain: nightly
           override: true
       - run: rustup component add rustfmt
-      - uses: actions-rs/cargo@v1
+      - name: Rust Format
+        uses: actions-rs/cargo@v1
         with:
           command: fmt
+          args: --all -- --check
+      - name: Rust Format (keyring-util)
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          working-directory: ./keyring-util
           args: --all -- --check
 
   clippy:
@@ -68,7 +85,14 @@ jobs:
           toolchain: stable
           override: true
       - run: rustup component add clippy
-      - uses: actions-rs/cargo@v1
+      - name: Clippy
+        uses: actions-rs/cargo@v1
         with:
           command: clippy
+          args: -- -D warnings
+      - name: Clippy (keyring-util)
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          working-directory: ./keyring-util
           args: -- -D warnings
--- a/keyring-util/Cargo.toml
+++ b/keyring-util/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "keyring-util"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-std = { version = "1.12.0", features = [ "attributes" ] }
+chrono = { version = "0.4.31", default-features = false, features = ["alloc", "clock"] }
+clap = { version = "4.4.6", features = [ "cargo", "derive" ] }
+oo7 = { path = "../" }
+rpassword = "7.2.0"
--- a/keyring-util/src/main.rs
+++ b/keyring-util/src/main.rs
@@ -1,0 +1,291 @@
+use std::{
+    collections::HashMap,
+    fmt,
+    io::Write,
+    process::{ExitCode, Termination},
+};
+
+use clap::{Command, CommandFactory, FromArgMatches, Parser};
+use oo7::{
+    dbus::{Algorithm, Collection, Service},
+    zbus,
+};
+
+struct Error(String);
+
+impl fmt::Debug for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl From<oo7::dbus::Error> for Error {
+    fn from(err: oo7::dbus::Error) -> Error {
+        Error(err.to_string())
+    }
+}
+
+impl Error {
+    fn new(s: &str) -> Self {
+        Self(String::from(s))
+    }
+}
+
+impl Termination for Error {
+    fn report(self) -> ExitCode {
+        ExitCode::FAILURE
+    }
+}
+
+#[derive(Parser)]
+#[command(
+    name = "store",
+    about = "Store a secret",
+    after_help = "The contents of the secret will be asked afterwards.\n\nExample:\n  keyring-util store 'My Personal Mail' smtp-port 1025 imap-port 143"
+)]
+struct StoreArgs {
+    #[clap(help = "Description for the secret")]
+    label: String,
+    #[clap(help = "List of attributes. This is a space separated list of pairs of key value")]
+    attributes: Vec<String>,
+}
+
+#[derive(Parser)]
+#[command(
+    name = "search",
+    about = "Search entries with matching attributes",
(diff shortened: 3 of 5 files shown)


keyring-util/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+                    arg!(--"all")
+                        .help("Whether to list all possible matches or only the first result")
+                )
+                .arg(
+                    arg!(attributes: [ATTRIBUTES] ...)
+                        .help("List of attributes. This is a space separated list of pairs of key value")
+                        .value_parser(value_parser!(String))
+                ),
+        )
+        .subcommand(Command::new("lock").about("Lock the keyring"))
```
warusadura, line 92: I think the new `lock` and `unlock` commands are nice. But, do we really need them here in this tool? I'm asking this because, `secret-tool` only does "store and retrieve passwords" and not lock/unlock keyring. For me personally I always prefer to keep my keyring(s) in a locked state and avoid accidental lock/unlock situations :)

(will update my review after seeing the implementation of this).

keyring-util/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "keyring-util"
```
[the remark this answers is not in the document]
A6GibKm: That tool already exist and this is definitively a new different one, which fills its role, but not necessarily in the same way. I think this more as a debugging tool than as part of the desktop  e.g. one could for example allow giving the tool a keyring file and use that instead of the portal. Apps and other services can just use oo7 or libsecret.

A6GibKm:
Regarding search and lookup, one could do `get` and `list` instead.

keyring-util/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+
+impl Termination for Error {
+    fn report(self) -> ExitCode {
+        ExitCode::FAILURE
+    }
+}
+
+#[async_std::main]
+async fn main() -> Result<(), Error> {
+    let cmd = Command::new("keyring-util")
```
bilelmoussaoui, line 39: I would use the derive feature of clap, makes the commands much easier to parse

keyring-util/src/main.rs
```diff
@@ -0,0 +1,259 @@
+use std::collections::HashMap;
+use std::fmt;
+use std::io::Write;
+use std::process::{ExitCode, Termination};
+
+use clap::parser::ValuesRef;
+use clap::{arg, value_parser, Command};
+use oo7::Keyring;
```
bilelmoussaoui, line 8: it should use the dbus specific backend as the tool is not supposed as is to work with sandboxed apps

keyring-util/Cargo.toml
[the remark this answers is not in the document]
sophie-h: I think it's confusing to use "keyring" as a name. That's already the name of GNOME's secret service implementation and the name of two different file formats used in two different contexts. If you are only accessing keyring files with it okay, but not if it's using secret service or even only using secret service.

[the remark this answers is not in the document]
A6GibKm: I am open to suggestion. Note that this util interacts with secrets, secret service, and keyrings, it will be hard to find a perfect name.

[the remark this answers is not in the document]
bilelmoussaoui: As this is not something we will be releasing any time soon, I would not block on it and we can open a separate issue to come up with a different name.

keyring-util/src/main.rs
bilelmoussaoui: I would keep them for backward compatibility purposes

warusadura (commented):
Looks good to me!
I have one suggestion though and I'm not sure it's possible to implement. It would be great if  `keyring-util` can handle a scenario like the following,

```
 ./target/debug/keyring-util store 'My password' key1 value1
Error: The collection 'default' doesn't exists
```
So, we need a new command to create a keyring or collection. What do you think @A6GibKm ?

keyring-util/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-std = { version = "1.12.0", features = [ "attributes" ] }
+chrono = "0.4.31"
+clap = { version = "4.4.6", features = [ "cargo", "derive" ] }
+oo7 = { path = "../" }
+rpassword = "7.2.0"
```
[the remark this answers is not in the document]
A6GibKm: Not sure if it is a good idea, I wanted to have the binary separated from oo7 initially. cc @bilelmoussaoui.

bilelmoussaoui:
> Looks good to me! I have one suggestion though and I'm not sure it's possible to implement. It would be great if `keyring-util` can handle a scenario like the following,
> 
> ```
>  ./target/debug/keyring-util store 'My password' key1 value1
> Error: The collection 'default' doesn't exists
> ```
> 
> So, we need a new command to create a keyring or collection. What do you think @A6GibKm ?

I would check first what is the behaviour of secret-tool in this case. Collections are not widely used so maybe we should automatically create the default collection if it doesn't exists?

keyring-util/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+        writeln!(&mut result, "schema = {schema} ").unwrap();
+    }
+    writeln!(&mut result, "attributes = {attributes:?} ").unwrap();
+    print!("{result}");
+
+    Ok(())
+}
+
+async fn collection<'a>() -> Result<Collection<'a>, Error> {
+    let service = Service::new(Algorithm::Encrypted)
```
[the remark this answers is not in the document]
A6GibKm: There are two common patterns we use repeatedly fallback the encrypted to a plain service, and load or create+load the default keyring, both used here (after the last force-push).

Regarding the first one, I am not sure how to call such method. Regarding the second, I don't think it is justified to create a helper for this, this might not be something users of the library should be doing often.

A6GibKm:
> > Looks good to me! I have one suggestion though and I'm not sure it's possible to implement. It would be great if `keyring-util` can handle a scenario like the following,
> > ```
> >  ./target/debug/keyring-util store 'My password' key1 value1
> > Error: The collection 'default' doesn't exists
> > ```
> > 
> > So, we need a new command to create a keyring or collection. What do you think @A6GibKm ?
> 
> I would check first what is the behaviour of secret-tool in this case. Collections are not widely used so maybe we should automatically create the default collection if it doesn't exists?

For the moment I will just create one if needed.

A6GibKm:
> Looks good to me but still lacks CI bits :)

Could you please take care of this? I have yet to understand how github CI works.

.github/workflows/CI.yml
```diff
@@ -40,6 +46,9 @@ jobs:
       - name: Build and test (OpenSSL)
         run: |
           dbus-run-session -- cargo test --no-default-features --features async-std --features openssl_crypto
+      - name: Build and test keyring-util
```
bilelmoussaoui, line 49: we don't have tests for keyring-util, so just drop this

Cargo.toml
```diff
@@ -89,3 +89,8 @@ required-features = ["async-std"]
 name = "dbus_service"
 path = "examples/dbus_service.rs"
 required-features = ["async-std"]
+
+[workspace]
```
bilelmoussaoui, line 93: doesn't sound good at all, i would rather move oo7 to a client sub-directory in a separate PR first and setup the workspace infrastructure before getting into such things

keyring-util/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "keyring-util"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-std = { version = "1.12.0", features = [ "attributes" ] }
+chrono = "0.4.31"
```
bilelmoussaoui, line 8: are there features we should disable?

keyring-util/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+clap = { version = "4.4.6", features = [ "cargo", "derive" ] }
+oo7 = { path = "../" }
+rpassword = "7.2.0"
+zbus = { version = "3.5.0", default-features = false}
```
bilelmoussaoui, line 12: why do we need direct zbus dependency when it is re-exported by oo7?

bilelmoussaoui (commented):
otherwise looks good! thanks

keyring-util/Cargo.toml
A6GibKm: It is not, but I added it.

keyring-util/Cargo.toml
A6GibKm: Set to `alloc`+`clock`.

Cargo.toml
A6GibKm: Removed. It was a nice hack to avoid having one extra CI for rust fmt

.github/workflows/CI.yml
A6GibKm: Renamed to `Build` and used `cargo build` instead.

bilelmoussaoui (approved):

PR #53, opened 2023-12-02
Do not take ownership of the attributes

A6GibKm:
This will free us of cloning them.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -106,28 +106,28 @@ async fn main() -> Result<(), Error> {
                 StoreArgs::from_arg_matches(matches).map_err(|e| Error::new(&e.to_string()))?;
             let attributes = parse_attributes(&args.attributes)?;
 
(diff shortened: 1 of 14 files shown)


bilelmoussaoui:
Would need a rebase

bilelmoussaoui (approved):

A6GibKm:
Rebased and tested with the CLI utility.

cli/src/main.rs
```diff
@@ -11,7 +11,7 @@ use oo7::{
     zbus,
 };
 
-const BINARY_NAME: &'static str = env!("CARGO_BIN_NAME");
+const BINARY_NAME: &str = env!("CARGO_BIN_NAME");
```
bilelmoussaoui, line 14: recent rust complains about this

A6GibKm: Removed, if needed can be done elsewhere.

PR #54, opened 2023-12-03
Add a portal implementation

A6GibKm:
There are a few things missing

- [ ] Use meson to build and install the binary together with the .portal and .service files
- [x] Find a better name
- [x] Clean code a bit
- [x] Is `Request` doing the correct thing?

This is compatible with gnome-keyring's stored secrets, but we store them with a more descriptive label, e.g.

    Secret Portal token for com.belmoussaoui.ashpd.demo

See https://github.com/flatpak/xdg-desktop-portal/issues/482.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,4 +4,5 @@ resolver = "2"
 members = [
   "client",
   "cli",
+  "portal",
 ]
--- a/portal/Cargo.toml
+++ b/portal/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "oo7-portal"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+futures-channel = "0.3"
+futures-util = "0.3"
+oo7 = { path = "../client" }
+ring = "0.17.5"
+secrecy = { version = "0.8", features = ["alloc"] }
+serde = { version = "1.0", features = ["derive"] }
+tokio = { version = "1.17", features = ["io-util", "net", "macros", "rt-multi-thread"] }
+tracing = "0.1"
+tracing-subscriber = "0.3"
+zbus = "3.5.0"
+zeroize = { version = "1", features = ["zeroize_derive"] }
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,9 +43,14 @@ jobs:
       - name: Build client (tokio / OpenSSL)
         run: |
           cargo build --manifest-path ./client/Cargo.toml --no-default-features --features tokio --features openssl_crypto
+
       - name: Build CLI
         run: |
           cargo build --manifest-path ./cli/Cargo.toml
+
+      - name: Build Portal
+        run: |
+          cargo build --manifest-path ./portal/Cargo.toml
 
       - name: Test (native)
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
+/portal/target
 Cargo.lock
 cli/target
--- a/portal/data/oo7-portal.portal
+++ b/portal/data/oo7-portal.portal
@@ -1,0 +1,4 @@
+[portal]
+DBusName=org.freedesktop.impl.portal.desktop.oo7
+Interfaces=org.freedesktop.impl.portal.Secret
+UseIn=gnome
--- a/portal/data/oo7-portal.service
+++ b/portal/data/oo7-portal.service
@@ -1,0 +1,3 @@
+[D-BUS Service]
+Name=org.freedesktop.impl.portal.Secret
+Exec=@bindir@/oo7-portal
--- a/portal/src/error.rs
+++ b/portal/src/error.rs
@@ -1,0 +1,35 @@
+use oo7::zbus;
+
+#[derive(zbus::DBusError, Debug)]
+pub enum Error {
+    Owned(String),
+}
+
+impl From<zbus::fdo::Error> for Error {
+    fn from(err: zbus::fdo::Error) -> Self {
+        Self::Owned(err.to_string())
+    }
+}
+impl From<zbus::Error> for Error {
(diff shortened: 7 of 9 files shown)


oo7-portal-impl/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[dependencies]
+async-std = { version = "1.11", features = [ "attributes" ] }
+futures-channel = "0.3"
+oo7 = { path = "../" }
+ring = "0.17.5"
+secrecy = { version = "0.8", features = [ "alloc" ] }
+serde = {version = "1.0", features = ["derive"]}
+tracing = "0.1"
+tracing-subscriber = "0.3"
+zbus = { git = "https://github.com/dbus2/zbus.git" }
```
bilelmoussaoui, line 15: Does it make sense to mix zbus stable & git?

bilelmoussaoui: Well, you can switch oo7 to use that and use oo7 from git as well, that would be a good start

A6GibKm: I think that would be fine... for a separate MR.

oo7-portal-impl/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "oo7-portal-impl"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-std = { version = "1.11", features = [ "attributes" ] }
```
ueno, line 7: Now that there is also the `keyring-util` sub-crate, it might be worth considering a switch to workspace so [dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) can be maintained in a single place?

bilelmoussaoui: We discussed this privately, I agree for switching this repository to a workspace but could be done in a separate pull request

bilelmoussaoui: This is now done, so please make the portal use the workspace feature & rename the project directory to portal. Maybe we should go with `oo7-portal`, `oo7-cli` and so on for now until we decide on something to be used. At least that would be consistent.

oo7-portal-impl/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+}
+
+struct Secret;
+
+#[dbus_interface(name = "org.freedesktop.impl.portal.Secret")]
+impl Secret {
+    #[dbus_interface(property, name = "version")]
+    fn version(&self) -> u32 {
+        0
```
bilelmoussaoui, line 53: Is this correct? I would move it to a const at the very top of the file.

bilelmoussaoui: It doesn't seem to set any version at all. So I changed it to 1 instead

portal/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "oo7-portal"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+async-std = { version = "1.11", features = [ "attributes" ] }
```
bilelmoussaoui, line 7: We have dropped async_std from the main crate, it would be nice to replace it with whetever the macro ends up calling or just using tokio imoh.

bilelmoussaoui:
>  Use meson to build and install the binary together with the .portal and .service files

This is something we can do later imho

portal/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+futures = "0.3"
```
bilelmoussaoui, line 8: Use the specific crate needed instead of the whole futures, i would also double check the various used dependencies and check if we can disable some of the default features to avoid pulling useless deps

A6GibKm: We need FutureExt, I am not sure if that can be loaded without the entire futures crate.

bilelmoussaoui: it comes from futures_util

A6GibKm: Done, thanks for the tip.

portal/Cargo.toml
A6GibKm: Ported to tokio.

portal-impl/src/main.rs
```diff
@@ -0,0 +1,228 @@
...
+    Ok(zeroize::Zeroizing::new(secret.to_vec()))
+}
+
+async fn retrieve_secret_inner(
+    app_id: &str,
+    fd: zvariant::Fd,
+) -> Result<HashMap<&'static str, zvariant::OwnedValue>, Error> {
+    let service = Service::new(Algorithm::Encrypted)
+        .await
+        .or(Service::new(Algorithm::Plain).await)?;
```
bilelmoussaoui, line 124:
```suggestion
    let service = Service::new(Algorithm::Encrypted)
        .await
        .or(Service::new(Algorithm::Plain).await)?;
```
should no longer be needed with https://github.com/bilelmoussaoui/oo7/pull/64

A6GibKm: Done.

Cargo.toml
```diff
@@ -4,4 +4,5 @@ resolver = "2"
 members = [
   "client",
   "cli",
+  "portal-impl",
```
bilelmoussaoui, line 7:
```suggestion
  "portal",
```

portal-impl/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "oo7-portal-impl"
```
bilelmoussaoui, line 2:
```suggestion
name = "oo7-portal"
```

A6GibKm:
From my point of view, the only thing missing here is more testing.

bilelmoussaoui:
We can have more testing once we get the server bits re-implemented. Then we can write a mock implementation and have all the bits tested. For now, let us land it

PR #56, opened 2023-12-11
wip: secret-server backend implementation

warusadura:
# Moved to: https://github.com/bilelmoussaoui/oo7/pull/73

### TODO:

- [ ] Implement the dbus secret service API for oo7 daemon: https://specifications.freedesktop.org/secret-service/latest/ [in progress] 
- [ ]  Add legacy keyring format (version 0) support
- [ ]  Write mock tests based on client oo7 library
- [ ] Test daemon in a real case scenario
- [ ] Squash temporary commits

#### - [ ] Interface implementations:

- [ ] ` org.freedesktop.Secret.Service`:
     #### Service methods:
    - [ ] OpenSession()
    - [x] CreateCollection()
    - [x] SearchItems()
    - [x] Unlock()
    - [x] Lock()
    - [ ] GetSecrets()
    - [x] ReadAlias()
    - [x] SetAlias()
     #### Signals:
   
- [ ] ` org.freedesktop.Secret.Collection`:
    #### Service methods:
    - [x] Delete()
    - [x] SearchItems()
    - [x] CreateItem()
     #### Signals:
    #### Other methods:
    - [ ] load existing keyrings in `.local/share/keyrings/`
    
- [ ] ` org.freedesktop.Secret.Session`:
     #### Service methods:
    - [x] Close()
     #### Signals:

<del> - [ ] ` org.freedesktop.Secret.Prompt`:
  
- [ ] ` org.freedesktop.Secret.Item`:
     #### Service methods:
    - [x] Delete()
    - [x] GetSecret()
    - [x] SetSecret()
    #### Signals:

#### - [x] Interfaces access to Property:
 - [x] `busctl --user get-property`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
   "client",
   "cli",
   "portal",
+  "daemon",
 ]
 
 [workspace.package]
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "oo7-daemon"
+edition.workspace = true
+exclude.workspace = true
+homepage.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
+oo7 = { path = "../client", features = ["unstable"] }
+serde = "1.0"
+tokio = { version = "1", features = ["full"] } # look for specific features
+zbus = "3.14.1"
+zeroize = { version = "1", features = ["zeroize_derive"] }
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -11,4 +11,4 @@ version.workspace = true
 oo7 = { path = "../client", features = ["unstable"] }
 serde = "1.0"
 tokio = { version = "1", features = ["full"] } # look for specific features
-zbus = "3.14.1"
+zbus = "4.0.0"
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -12,4 +12,3 @@ oo7 = { path = "../client", features = ["unstable"] }
 serde = "1.0"
 tokio = { version = "1", features = ["full"] } # look for specific features
 zbus = "3.14.1"
-zeroize = { version = "1", features = ["zeroize_derive"] }
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -1,0 +1,44 @@
+pub mod service;
+
+use std::{future::pending, sync::OnceLock};
+
+use oo7::portal::Keyring;
+use zbus::{ConnectionBuilder, Result};
+
+const SECRET_SERVICE_OBJECTPATH: &str = "/org/freedesktop/secrets_";
+// const SECRET_COLLECTION_OBJECTPATH: &str =
+// "/org/freedesktop/secrets_/collection"; const SECRET_SESSION_OBJECTPATH: &str
+// = "/org/freedesktop/secrets_/session";
+
+use crate::service::service::Service;
+
+pub static KEYRING: OnceLock<Keyring> = OnceLock::new();
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let secret_service = Service {
+        collections: Vec::new(),
+    };
+
+    // let collection = Collection::default();
+    // let session = Session {
+    // path: ObjectPath::default().into(),
+    // };
+
+    let _service = ConnectionBuilder::session()?
+        .name("org.freedesktop.secrets_")?
+        .serve_at(SECRET_SERVICE_OBJECTPATH, secret_service)?
+        //.serve_at(SECRET_COLLECTION_OBJECTPATH, collection)?
+        //.serve_at(SECRET_SESSION_OBJECTPATH, session)?
+        .build()
+        .await?;
+
+    let _keyring = match Keyring::load_default().await {
+        Ok(keyring) => KEYRING.set(keyring),
+        Err(_) => todo!(), // call some init to create default keyring
+    };
+
+    pending::<()>().await;
+
+    Ok(())
+}
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -6,37 +6,21 @@ use oo7::portal::Keyring;
 use zbus::{ConnectionBuilder, Result};
 
 const SECRET_SERVICE_OBJECTPATH: &str = "/org/freedesktop/secrets_";
-// const SECRET_COLLECTION_OBJECTPATH: &str =
-// "/org/freedesktop/secrets_/collection"; const SECRET_SESSION_OBJECTPATH: &str
-// = "/org/freedesktop/secrets_/session";
 
 use crate::service::service::Service;
 
 pub static KEYRING: OnceLock<Keyring> = OnceLock::new();
+// TODO remove this and use service::keyring()
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    let secret_service = Service {
-        collections: Vec::new(),
-    };
-
-    // let collection = Collection::default();
-    // let session = Session {
-    // path: ObjectPath::default().into(),
-    // };
+    let service = Service::new().await;
 
     let _service = ConnectionBuilder::session()?
         .name("org.freedesktop.secrets_")?
-        .serve_at(SECRET_SERVICE_OBJECTPATH, secret_service)?
-        //.serve_at(SECRET_COLLECTION_OBJECTPATH, collection)?
-        //.serve_at(SECRET_SESSION_OBJECTPATH, session)?
+        .serve_at(SECRET_SERVICE_OBJECTPATH, service)?
         .build()
         .await?;
-
-    let _keyring = match Keyring::load_default().await {
-        Ok(keyring) => KEYRING.set(keyring),
-        Err(_) => todo!(), // call some init to create default keyring
-    };
 
     pending::<()>().await;
 
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -3,9 +3,7 @@ pub mod service;
 use std::{future::pending, sync::OnceLock};
 
 use oo7::portal::Keyring;
-use zbus::{ConnectionBuilder, Result};
-
-const SECRET_SERVICE_OBJECTPATH: &str = "/org/freedesktop/secrets_";
+use zbus::Result;
 
 use crate::service::service::Service;
 
@@ -15,12 +13,7 @@ pub static KEYRING: OnceLock<Keyring> = OnceLock::new();
 #[tokio::main]
 async fn main() -> Result<()> {
     let service = Service::new().await;
-
-    let _service = ConnectionBuilder::session()?
-        .name("org.freedesktop.secrets_")?
-        .serve_at(SECRET_SERVICE_OBJECTPATH, service)?
-        .build()
-        .await?;
+    service.run().await?;
 
     pending::<()>().await;
 
--- a/daemon/src/service/service.rs
+++ b/daemon/src/service/service.rs
@@ -63,9 +63,6 @@ impl Service {
         Ok((path, prompt))
     }
 
-    // I have updated the collection interface impl. So, I need to re check everything starting
-    // here
-    /*
     #[zbus(out_args("unlocked", "locked"))]
     pub async fn search_items(
         &self,
@@ -93,6 +90,7 @@ impl Service {
         Ok((unlocked, locked))
     }
 
+    /*
     #[zbus(out_args("unlocked", "prompt"))]
     pub async fn unlock(
         &mut self,
@@ -167,28 +165,36 @@ impl Service {
                 if in_collection.path().as_str() == item.as_str() {}
             }
         }
-    }
+    }*/
 
     pub async fn read_alias(&self, name: &str) -> ObjectPath {
-        let mut ret = ObjectPath::default();
+        let mut path = ObjectPath::default();
         for collection in &self.collections {
-            if collection.label() == name {
-                ret = collection.path().into();
+            if collection.alias() == name {
+                path = collection.path().into();
             }
         }
 
-        ret
+        path
     }
 
-    pub async fn set_alias(&mut self, name: String, collection: OwnedObjectPath) {
-        // WIP: not complete:: handle alias
-        for in_collection in &mut self.collections {
-            if in_collection.path().as_str() == collection.as_str() {
-                in_collection.set_label(name.clone()).await;
+    pub async fn set_alias(&mut self, name: String, collection_path: OwnedObjectPath) {
+        // handle "default" aliais
+        if name == "default" {
+            return; // return Err
+        }
+
+        // handle "/" ObjectPath
+        if collection_path.as_str() == "/" {
+            return; // return Err
+        }
+
+        for collection in &mut self.collections {
+            if collection.path() == collection_path.as_str() {
+                collection.set_alias(name.clone());
             }
         }
     }
-    */
 
     #[zbus(property, name = "Collections")]
     pub fn collections(&self) -> Vec<ObjectPath> {
--- a/daemon/src/service/service.rs
+++ b/daemon/src/service/service.rs
@@ -7,13 +7,15 @@ use oo7::{
     portal::{Item, Keyring},
 };
 use serde::Serialize;
-use zbus::{fdo, interface, zvariant, Error, ObjectServer, SignalContext};
+use zbus::{connection, fdo, interface, zvariant, Error, ObjectServer, SignalContext};
 use zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Value};
 
 use crate::{
     service::{collection::Collection, session::Session},
     KEYRING,
 };
+
+const SECRET_SERVICE_OBJECTPATH: &str = "/org/freedesktop/secrets_";
 
 #[derive(Serialize, Debug)]
 pub struct Service {
@@ -205,5 +207,17 @@ impl Service {
 
     pub fn keyring(&self) -> &Keyring {
         &self.keyring
+    }
+
+    pub async fn run(self) -> Result<(), Error> {
+        match connection::Builder::session()?
+            .name("org.freedesktop.secrets_")?
+            .serve_at(SECRET_SERVICE_OBJECTPATH, self)?
+            .build()
+            .await // .await.map_err(From::from)
+        {
+            Ok(_) => Ok(()),
+            Err(e) => Err(e),
+        }
     }
 }
--- a/daemon/src/service/service.rs
+++ b/daemon/src/service/service.rs
@@ -1,6 +1,6 @@
-//  org.freedesOnceCellktop.Secret.Service
+// org.freedesOnceCellktop.Secret.Service
 
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, sync::Arc, time::SystemTime};
 
 use oo7::{
     dbus::api::Properties,
@@ -46,17 +46,26 @@ impl Service {
         alias: String,
         #[zbus(object_server)] object_server: &ObjectServer,
     ) -> fdo::Result<(ObjectPath, ObjectPath)> {
-        let _attributes = properties.attributes(); // expand and pass these to Collection:new
-        let collection = Collection::new("temp"); // temporarily
-        let path = format!("/org/freedesktop/secrets_/collection/{}", &alias);
+        let collection = Collection::new(
+            properties,
+            alias,
+            SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .unwrap()
+                .as_secs(),
+        );
+
+        let path = collection.path().to_owned();
 
         let _ = object_server.at(path.clone(), collection).await;
-        let new_collection_path = ObjectPath::try_from(path).unwrap(); // temporarily
         let prompt = ObjectPath::default().into(); // temp Prompt
 
-        Ok((new_collection_path, prompt))
+        Ok((path, prompt))
     }
 
+    // I have updated the collection interface impl. So, I need to re check everything starting
+    // here
+    /*
     #[zbus(out_args("unlocked", "locked"))]
     pub async fn search_items(
         &self,
@@ -122,13 +131,14 @@ impl Service {
     ) -> fdo::Result<(Vec<OwnedObjectPath>, OwnedObjectPath)> {
         // manage lock state in memory
         // when do we need to prompt?
+        // WIP
         let mut locked: Vec<OwnedObjectPath> = Vec::new();
 
         for object in objects {
             for collection in &mut self.collections {
                 if collection.path().as_str() == object.as_str() {
                     if !collection.locked() {
-                        collection.set_locked(true).await;
+                        //collection.set_locked(true).await;
                         locked.push(object.clone());
                     }
                 }
@@ -178,6 +188,7 @@ impl Service {
             }
         }
     }
+    */
 
     #[zbus(property, name = "Collections")]
     pub fn collections(&self) -> Vec<ObjectPath> {
--- a/daemon/src/service/service.rs
+++ b/daemon/src/service/service.rs
@@ -1,0 +1,191 @@
+//  org.freedesOnceCellktop.Secret.Service
+
+use std::collections::HashMap;
+
+use oo7::{dbus::api::Properties, portal::Item};
+use serde::Serialize;
+use zbus::{dbus_interface, fdo, zvariant, Error, ObjectServer, SignalContext};
+use zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Type, Value};
+
+use crate::{
+    service::{collection::Collection, session::Session},
+    KEYRING,
+};
+
+#[derive(Serialize, Type, Debug)]
+pub struct Service {
+    pub collections: Vec<Collection>,
+}
+
+#[dbus_interface(name = "org.freedesktop.Secret.Service")]
+impl Service {
+    pub async fn open_session(
+        &self,
+        // algorithm: &str,
+        input: Value<'_>,
+    ) -> fdo::Result<(OwnedValue, OwnedObjectPath)> {
+        // WIP: not complete
+        let session = Session::new(ObjectPath::try_from("/s1").unwrap().into()).await;
+        Ok((
+            OwnedValue::try_from(input).unwrap(), // avoid using unwrap
+            ObjectPath::try_from(session.path().await).unwrap().into(),
+        ))
+    }
+
+    #[dbus_interface(out_args("collection", "prompt"))]
+    pub async fn create_collection(
+        &self,
+        properties: Properties,
+        alias: String,
+        #[zbus(object_server)] object_server: &ObjectServer,
+    ) -> fdo::Result<(ObjectPath, ObjectPath)> {
+        let _attributes = properties.attributes(); // expand and pass these to Collection:new
+        let collection = Collection::new("temp"); // temporarily
+        let path = format!("/org/freedesktop/secrets_/collection/{}", &alias);
+
+        let _ = object_server.at(path.clone(), collection).await;
+        let new_collection_path = ObjectPath::try_from(path).unwrap(); // temporarily
+        let prompt = ObjectPath::default().into(); // temp Prompt
+
+        Ok((new_collection_path, prompt))
+    }
+
+    #[dbus_interface(out_args("unlocked", "locked"))]
+    pub async fn search_items(
+        &self,
+        attributes: HashMap<&str, &str>,
+    ) -> fdo::Result<(Vec<Item>, Vec<Item>)> {
+        let items = match KEYRING.get().unwrap().search_items(&attributes).await {
+            Ok(i) => i,
+            Err(_) => todo!(),
+        };
+
+        let mut unlocked: Vec<Item> = Vec::new();
+        let mut locked: Vec<Item> = Vec::new();
+
+        for item in items {
+            let attributes = item.attributes();
+            if attributes.get("locked").is_some() {
+                // this if condition is probably wrong
+                // how to access &AttributeValue value
+                locked.push(item)
+            } else {
+                unlocked.push(item)
+            }
+        }
+
+        Ok((unlocked, locked))
+    }
+
+    #[dbus_interface(out_args("unlocked", "prompt"))]
+    pub async fn unlock(
+        &mut self,
+        objects: Vec<OwnedObjectPath>,
+    ) -> fdo::Result<(Vec<OwnedObjectPath>, OwnedObjectPath)> {
+        // manage unlock state in memory
+        // when do we need to prompt?
+        let mut unlocked: Vec<OwnedObjectPath> = Vec::new();
+
+        'main: for object in objects {
+            for collection in &mut self.collections {
+                if collection.path().as_str() == object.as_str() {
+                    if collection.locked() {
+                        collection.set_locked(false).await;
+                        unlocked.push(object.clone());
+                    } else {
+                        break 'main;
+                    }
+                }
+            }
+        }
+
+        if unlocked.len() == 0 {
+            unlocked.push(ObjectPath::default().into());
+        }
+
+        let prompt = ObjectPath::default().into(); // temporarily
+
+        Ok((unlocked, prompt))
+    }
+
+    #[dbus_interface(out_args("locked", "prompt"))]
+    pub async fn lock(
+        &mut self,
+        objects: Vec<OwnedObjectPath>,
+    ) -> fdo::Result<(Vec<OwnedObjectPath>, OwnedObjectPath)> {
+        // manage lock state in memory
+        // when do we need to prompt?
+        let mut locked: Vec<OwnedObjectPath> = Vec::new();
+
+        for object in objects {
+            for collection in &mut self.collections {
+                if collection.path().as_str() == object.as_str() {
+                    if !collection.locked() {
+                        collection.set_locked(true).await;
+                        locked.push(object.clone());
+                    }
+                }
+            }
+        }
+
+        if locked.len() == 0 {
+            locked.push(ObjectPath::default().into());
+        }
+
+        let prompt = ObjectPath::default().into(); // temporarily
+
+        Ok((locked, prompt))
+    }
+
+    pub async fn get_secrets(&self, items: Vec<OwnedObjectPath>) {
+        // WIP: not complete
+        if KEYRING.get().unwrap().n_items().await == 0 {
+            return;
+        }
+
+        // call item iface get_secret multiple times ?
+
+        for in_collection in &self.collections {
+            for item in &items {
+                if in_collection.path().as_str() == item.as_str() {}
+            }
+        }
+    }
+
+    pub async fn read_alias(&self, name: &str) -> ObjectPath {
+        let mut ret = ObjectPath::default();
+        for collection in &self.collections {
+            if collection.label() == name {
+                ret = collection.path().into();
+            }
+        }
+
+        ret
+    }
+
+    pub async fn set_alias(&mut self, name: String, collection: OwnedObjectPath) {
+        // WIP: not complete:: handle alias
+        for in_collection in &mut self.collections {
+            if in_collection.path().as_str() == collection.as_str() {
+                in_collection.set_label(name.clone()).await;
+            }
+        }
+    }
+
+    #[dbus_interface(property, name = "Collections")]
+    pub fn collections(&self) -> Vec<ObjectPath> {
+        self.collections
+            .iter()
+            .map(|collection| collection.path())
+            .collect()
+    }
+
+    #[dbus_interface(signal)]
+    pub async fn collection_created(ctxt: &SignalContext<'_>) -> Result<(), Error>;
+
+    #[dbus_interface(signal)]
+    pub async fn collection_deleted(ctxt: &SignalContext<'_>) -> Result<(), Error>;
+
+    #[dbus_interface(signal)]
+    pub async fn collection_changed(ctxt: &SignalContext<'_>) -> Result<(), Error>;
+}
--- a/daemon/src/service/service.rs
+++ b/daemon/src/service/service.rs
@@ -7,7 +7,7 @@ use oo7::{
     portal::{Item, Keyring},
 };
 use serde::Serialize;
-use zbus::{dbus_interface, fdo, zvariant, Error, ObjectServer, SignalContext};
+use zbus::{fdo, interface, zvariant, Error, ObjectServer, SignalContext};
 use zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Value};
 
 use crate::{
@@ -22,7 +22,7 @@ pub struct Service {
     keyring: Arc<Keyring>,
 }
 
-#[dbus_interface(name = "org.freedesktop.Secret.Service")]
+#[interface(name = "org.freedesktop.Secret.Service")]
 impl Service {
     pub async fn open_session(
         &self,
@@ -37,7 +37,7 @@ impl Service {
         ))
     }
 
-    #[dbus_interface(out_args("collection", "prompt"))]
+    #[zbus(out_args("collection", "prompt"))]
     pub async fn create_collection(
         &self,
         properties: Properties,
@@ -55,7 +55,7 @@ impl Service {
         Ok((new_collection_path, prompt))
     }
 
-    #[dbus_interface(out_args("unlocked", "locked"))]
+    #[zbus(out_args("unlocked", "locked"))]
     pub async fn search_items(
         &self,
         attributes: HashMap<&str, &str>,
@@ -82,7 +82,7 @@ impl Service {
         Ok((unlocked, locked))
     }
 
-    #[dbus_interface(out_args("unlocked", "prompt"))]
+    #[zbus(out_args("unlocked", "prompt"))]
     pub async fn unlock(
         &mut self,
         objects: Vec<OwnedObjectPath>,
@@ -113,7 +113,7 @@ impl Service {
         Ok((unlocked, prompt))
     }
 
-    #[dbus_interface(out_args("locked", "prompt"))]
+    #[zbus(out_args("locked", "prompt"))]
     pub async fn lock(
         &mut self,
         objects: Vec<OwnedObjectPath>,
@@ -177,7 +177,7 @@ impl Service {
         }
     }
 
-    #[dbus_interface(property, name = "Collections")]
+    #[zbus(property, name = "Collections")]
     pub fn collections(&self) -> Vec<ObjectPath> {
         self.collections
             .iter()
@@ -185,13 +185,13 @@ impl Service {
             .collect()
     }
 
-    #[dbus_interface(signal)]
+    #[zbus(signal)]
     pub async fn collection_created(ctxt: &SignalContext<'_>) -> Result<(), Error>;
 
-    #[dbus_interface(signal)]
+    #[zbus(signal)]
     pub async fn collection_deleted(ctxt: &SignalContext<'_>) -> Result<(), Error>;
 
-    #[dbus_interface(signal)]
+    #[zbus(signal)]
     pub async fn collection_changed(ctxt: &SignalContext<'_>) -> Result<(), Error>;
 }
 
--- a/daemon/src/service/service.rs
+++ b/daemon/src/service/service.rs
@@ -1,20 +1,25 @@
 //  org.freedesOnceCellktop.Secret.Service
 
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
-use oo7::{dbus::api::Properties, portal::Item};
+use oo7::{
+    dbus::api::Properties,
+    portal::{Item, Keyring},
+};
 use serde::Serialize;
 use zbus::{dbus_interface, fdo, zvariant, Error, ObjectServer, SignalContext};
-use zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Type, Value};
+use zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Value};
 
 use crate::{
     service::{collection::Collection, session::Session},
     KEYRING,
 };
 
-#[derive(Serialize, Type, Debug)]
+#[derive(Serialize, Debug)]
 pub struct Service {
-    pub collections: Vec<Collection>,
+    collections: Vec<Collection>,
+    #[serde(skip_serializing)]
+    keyring: Arc<Keyring>,
 }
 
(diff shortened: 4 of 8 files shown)


server/src/service/service.rs
```diff
@@ -0,0 +1,135 @@
...
+#[dbus_interface(name = "org.freedesktop.Secret.Service")]
+impl Service {
+    pub async fn open_session(
+        &self,
+        algorithm: &str,
+        input: Value<'_>,
+    ) -> zbus::fdo::Result<(OwnedValue, OwnedObjectPath)> {
+        let session = Session::new(ObjectPath::try_from("/s1").unwrap().into()).await;
+        Ok((
+            OwnedValue::try_from(input).unwrap(),
```
TTWNO, line 28: Errors that can happen due to invalid user input should not use `.unwrap()`. I'd recommend using a `Result<T, E>` type here to make error handling easier. While working on improving the code, I'd recommend turning on `clippy::unwrap_used` and `clippy::expect_used` to find other places where this may be an issue.

server/src/service/service.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub async fn set_alias(&self, name: String, collection: OwnedObjectPath) {
+        for collection_ in self.collections().await {
+            if collection_.path().await.as_str() == collection.as_str() {
+                //collection_.set_label(name.clone()).await;
+            }
+        }
+    }
+
+    /*
+    #[dbus_interface(property)]
```
TTWNO, line 100: I also think I helped you fix this earlier, but maybe I'm thinking of something else.

server/src/service/service.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+impl Service {
+    pub async fn collections(&self) -> &Vec<Collection> {
+        &self.collections
+    }
+}
+
+pub async fn init_secret_service() -> Service {
+    let mut service = Service {
```
TTWNO, line 122: `mut` is not necessary here. You do not change the value.

I'd recommend adding the `unused_mut` clippy lint :)

server/src/service/session.rs
```diff
@@ -0,0 +1,35 @@
...
+        Self { path: path.into() }
+    }
+
+    pub async fn path(&self) -> &OwnedObjectPath {
+        &self.path
+    }
+}
+
+pub async fn init_session_service() -> Session {
+    Session::new(ObjectPath::try_from("/").unwrap().into()).await
```
TTWNO, line 34: If you're looking for an (Owned)ObjectPath that is just "/", you should use [(Owned)ObjectPath::default()](https://docs.rs/zvariant/3.15.0/src/zvariant/object_path.rs.html#118-122)

This also eliminates some more unnecessary `.unwrap()`s

TTWNO (commented):
Make `cargo clippy` happy first. In general, that is a great place to start in terms of code review. It looks like this repo even [has a clippy job](https://github.com/bilelmoussaoui/oo7/actions/runs/7514920869/job/20458313090), so you'd never be able to merge before cleaning that up anyway.

If you have any problems, ping me here or on Matrix. I can help out with a stuck clippy lint/borrow checker problem anytime.

server/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "server"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+zbus = "3.14.1"
+tokio = { version = "1", features = ["full"] }
```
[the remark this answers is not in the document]
warusadura: Got it, thanks. I'll make a note of this to fix at later stage.

server/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+oo7 = { path = "../client", features = ["unstable"] }
+serde = "1.0.193"
```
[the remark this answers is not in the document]
warusadura: Done, thanks!

server/src/service/collection.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+#[dbus_interface(name = "org.freedesktop.Secret.Collection")]
+impl Collection {
+    pub async fn delete(&self, #[zbus(object_server)] object_server: &ObjectServer) -> ObjectPath {
+        let _ = object_server.remove::<Collection, _>(&self.path).await;
+        ObjectPath::try_from("/").unwrap()
+    }
+
+    pub async fn search_items(&self, attributes: HashMap<&str, &str>) -> Vec<Item> {
```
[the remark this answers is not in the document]
warusadura: Yeah make sense. Actually this should return `Vec<OwnedObjectPath>`. I'm keeping `Vec<Item>` temporarily.

server/src/service/service.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    pub async fn get_secrets(&self, items: Vec<OwnedObjectPath>) {
+        let n_secrets = KEYRING.get().unwrap().n_items().await;
+        if n_secrets == 0 {
+            return;
+        }
+    }
+
+    pub async fn read_alias(&self, name: &str) -> ObjectPath {
```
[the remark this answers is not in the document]
warusadura: Yes, I think so. `gnome-keyring-daemon`: the current secret service provider returns "/" in this situation.

server/src/service/service.rs
warusadura: Yeah, I agree.

server/src/service/session.rs
warusadura: This is much cleaner approach. Thanks!

server/src/service/service.rs
warusadura: Nope :) I got an explanation why the trait bound error pops here.

server/src/service/service.rs
warusadura: Got it, thanks!

server/src/service/item.rs
```diff
@@ -0,0 +1,57 @@
...
+use oo7::dbus::api::Secret;
+use serde::{Deserialize, Serialize};
+use zbus::{dbus_interface, fdo, zvariant, ObjectServer, SignalContext};
+use zvariant::{ObjectPath, OwnedObjectPath};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Item {
+    locked: bool,
+    attributes: Properties,
+    label: &'static str,
```
bilelmoussaoui, line 15: The properties already contain the label, so you can just store that

warusadura: done, thanks!

client/src/dbus/api/mod.rs
```diff
@@ -13,6 +13,8 @@ impl Unlockable for &zbus::zvariant::OwnedObjectPath {}
 mod collection;
 mod item;
 mod prompt;
+pub mod properties;
+#[cfg(not(feature = "unstable"))]
```
bilelmoussaoui, line 17: This change is not needed, would drop the whole commit

server/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "server"
+version = "0.1.0"
+edition = "2021"
```
bilelmoussaoui, line 4: use `version.workspace` & others, see what we have done in other crates

server/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+
+[dependencies]
+zbus = "3.14.1"
```
bilelmoussaoui, line 7: Keep those sorted please

server/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+tokio = { version = "1", features = ["full"] } # look for specific features
+oo7 = { path = "../client", features = ["unstable"] }
+serde = "1.0"
+once_cell = "1.19.0"
```
bilelmoussaoui, line 11: Avoid using once_cell, we are trying to get rid of it as what we usually need it for can be covered with std nowadays

server/src/main.rs
```diff
@@ -0,0 +1,41 @@
+// WIP: not ready for reviews
+
+pub mod service;
+
+use std::collections::HashMap;
+
+use once_cell::sync::OnceCell;
+use oo7::portal::Keyring;
+use std::future::pending;
```
bilelmoussaoui, line 9: I would run rustfmt from nightly, it should sort the imports here for you

server/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+use zbus::{ConnectionBuilder, Result};
+
+use crate::service::collection::init_collection_service;
+use crate::service::service::init_secret_service;
+use crate::service::session::{init_session_service, Session};
```
bilelmoussaoui, line 14: This should not be done here at all; you just need to initialize the service & wait until an interaction happens before initializing anything else

server/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+async fn main() -> Result<()> {
+    let secret_service = init_secret_service().await;
+    let collection = init_collection_service().await;
+    let session = init_session_service().await;
+
+    let _service = ConnectionBuilder::session()?
+        .name("org.freedesktop.secrets_")?
+        .serve_at("/org/freedesktop/secrets_", secret_service)?
+        .serve_at("/org/freedesktop/secrets_/collection", collection)?
+        .serve_at("/org/freedesktop/secrets_/session", session)?
```
bilelmoussaoui, line 29: could use some consts for the name & the various paths.

server/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+        .build()
+        .await?;
+
+    let _keyring = match Keyring::load_default().await {
+        Ok(i) => KEYRING.set(i),
```
bilelmoussaoui, line 34: don't use variable names like this please
```suggestion
        Ok(keyring) => KEYRING.set(keyring),
```
or instance

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+use crate::service::item;
+use crate::KEYRING;
+
+#[derive(Default, Debug, Deserialize)]
+pub struct Collection {
+    pub items: Vec<item::Item>,
+    label: String,
+    locked: bool,
+    created: u64, // how to represent date
+    modified: u64,
```
bilelmoussaoui, line 20: Can't those use std types? see the client library

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+use oo7::dbus::api::properties::Properties;
+use oo7::portal::Item;
+use serde::{Deserialize, Serialize, Serializer};
+use zbus::{dbus_interface, zvariant, Error, ObjectServer, SignalContext};
+use zvariant::{ObjectPath, OwnedObjectPath, Signature, Type};
+
+use crate::service::item;
+use crate::KEYRING;
+
+#[derive(Default, Debug, Deserialize)]
```
bilelmoussaoui, line 14: Does this needs to implement Deserialize at all?

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+        items // currently returns Vec<oo7::portal::Item>, this should be a path
+    }
+
+    pub async fn create_item(
+        &mut self,
+        properties: Properties,
+        secret: &str,
+        replace: bool,
+    ) -> (ObjectPath, ObjectPath) {
+        let label = properties.attributes().unwrap().get("label").unwrap();
```
bilelmoussaoui, line 46: There is a getter for the label

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+        let mut attributes: HashMap<&str, &str> = Default::default();
+
+        for (key, value) in properties.attributes().unwrap().into_iter() {
+            attributes.insert(key.as_str(), value.as_str());
+        }
```
bilelmoussaoui, line 51: Why you need this?

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        let _collection = KEYRING
```
bilelmoussaoui, line 53: This doesn't return anything so instead handle the error

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+            .get()
+            .unwrap()
+            .create_item(label, attributes.clone(), secret, replace)
+            .await;
+
+        // lookup just created item
+        let lookup_item = KEYRING
```
bilelmoussaoui, line 60: Not sure why you need this for? you already have the attributes above, if you need the portal::Item itself, you can make the create_item method returns it for example.

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+            attributes_for_new.insert(key.as_str(), value);
+        }
+
+        // TODO map portal::Item to crate::Item and push it to items attribute
+        let item = item::Item::new(
+            &attributes_for_new,
+            lookup_item.secret().to_vec(),
+            lookup_item.label(),
+            lookup_item.created().as_secs(),
+            lookup_item.modified().as_secs(),
```
bilelmoussaoui, line 83: I would avoid any conversions here, see what I mentioned above

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        (created_item_path, prompt)
+    }
+
+    #[dbus_interface(property, name = "Items")]
+    async fn items(&self) -> Vec<ObjectPath> {
+        let mut paths: Vec<ObjectPath> = Vec::new();
+
+        for path in &self.items {
+            paths.push(path.path().await.into());
```
bilelmoussaoui, line 97: I don't think this should be async at all. then you could just write `self.items.iter().map(|item| item.path()).collect()`

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        paths
+    }
+
+    #[dbus_interface(property, name = "Label")]
+    pub async fn label(&self) -> &String {
```
bilelmoussaoui, line 104: not async & 
```suggestion
    pub async fn label(&self) -> &str {
```

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+        &self.label
+    }
+
+    #[dbus_interface(property, name = "Locked")]
+    pub async fn locked(&self) -> bool {
```
bilelmoussaoui, line 109: not async

bilelmoussaoui: same for all other properties

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+        let item = item::Item::new(
+            &attributes_for_new,
+            lookup_item.secret().to_vec(),
+            lookup_item.label(),
+            lookup_item.created().as_secs(),
+            lookup_item.modified().as_secs(),
+            self.path().await,
+        )
+        .await;
+        self.items.push(item);
```
bilelmoussaoui, line 87: You need to call `item_created` so the signal is emitted.

server/src/service/collection.rs
bilelmoussaoui: Also the label is readwrite, so you need a setter as well

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+    {
+        OwnedObjectPath::serialize(&self.path, serializer)
+    }
+}
+
+impl Type for Collection {
+    fn signature() -> Signature<'static> {
+        ObjectPath::signature()
+    }
+}
```
bilelmoussaoui, line 146: you can replace this with 
```
#[derive(zvariant::Type)]
#[zvariant(signature = "o")]
```
on the struct itself

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl Collection {
+    // temporarily creates a generic Collection object
+    pub async fn new(label: String) -> Self {
```
bilelmoussaoui, line 150:
```suggestion
    pub async fn new(label: &str) -> Self {
```

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+        Collection {
```
bilelmoussaoui, line 151:
```suggestion
        Self {
```

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+            items: Vec::new(),
+            label: label.clone(),
```
bilelmoussaoui, line 153:
```suggestion
            label: label.to_owned(),
```
Use ToOwned trait when you are taking ownership of the passed data instead of clone.

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+            locked: false,
```
bilelmoussaoui, line 154: Is this a sane default? what current gnome-keyring do here?

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+            created: 23123,
+            modified: 23123,
```
bilelmoussaoui, line 156: The data has passed to the constructor

server/src/service/collection.rs
bilelmoussaoui, line 150: No reason to be async as well

server/src/service/collection.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    pub async fn path(&self) -> &OwnedObjectPath {
+        &self.path
+    }
+
+    pub async fn set_label(&mut self, label: String) {
+        self.label = label;
+    }
+
+    pub async fn set_locked(&mut self, locked: bool) {
```
bilelmoussaoui, line 173: The lock property is read only

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+use std::collections::HashMap;
+
+use oo7::dbus::api::properties::Properties;
+use serde::{Deserialize, Serialize};
+use zbus::{dbus_interface, fdo, zvariant, ObjectServer};
+use zvariant::{ObjectPath, OwnedObjectPath};
+
+use crate::KEYRING;
+
+#[derive(Debug, Serialize, Deserialize)]
```
bilelmoussaoui, line 12: Does it make sense for this type to implement Serialize/Deserialize?

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+pub struct Item {
+    locked: bool,
+    attributes: Properties,
```
bilelmoussaoui, line 15: Rename to properties

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+    created: u64,
+    modified: u64,
+    secret: Vec<u8>,
```
bilelmoussaoui, line 18: Use zeroize here, see the client library

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+
+#[dbus_interface(name = "org.freedesktop.Secret.Item")]
+impl Item {
+    pub async fn delete(
+        &self,
+        #[zbus(object_server)] object_server: &ObjectServer,
+    ) -> fdo::Result<ObjectPath> {
+        let mut properties: HashMap<&str, &str> = Default::default();
+
+        for (key, value) in self.attributes().await.attributes().unwrap().into_iter() {
```
bilelmoussaoui, line 30: The attributes getter is already giving you just the attributes. This whole thing is not needed

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+            properties.insert(key.as_str(), value.as_str());
+        }
+
+        let _ = KEYRING.get().unwrap().delete(&properties).await;
+        let _ = object_server.remove::<Item, _>(self.path().await).await;
```
bilelmoussaoui, line 35: You also need to call item_deleted in the collection that contains this item, somehow

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+        Ok(ObjectPath::default().into())
+    }
+
+    pub async fn get_secret(&self /*session: Session*/) -> fdo::Result<Vec<u8>> {
```
bilelmoussaoui, line 39: The session should be an object path, as creating one would require some extra-logic that you can't do in the Deserialize implementation.

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+        let item = KEYRING
+            .get()
+            .unwrap()
+            .lookup_item(&self.lookup_item_attributes().await)
+            .await
+            .unwrap()
+            .unwrap();
```
bilelmoussaoui, line 46: I am not happy about those unwraps here.

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        item.set_secret(secret)
+    }
+
+    pub async fn locked(&self) -> &bool {
+        &self.locked
+    }
+
+    pub async fn attributes(&self) -> &Properties {
```
bilelmoussaoui, line 67: this & label are readwrite properties

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+            .unwrap()
+            .lookup_item(&self.lookup_item_attributes().await)
+            .await
+            .unwrap()
+            .unwrap();
+
+        item.set_secret(secret)
+    }
+
+    pub async fn locked(&self) -> &bool {
```
bilelmoussaoui, line 63:
```suggestion
    pub async fn locked(&self) -> bool {
```
& not async

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub async fn attributes(&self) -> &Properties {
+        &self.attributes
+    }
+
+    pub async fn label(&self) -> &str {
+        let attributes = self.attributes().await;
+        attributes.label()
+    }
+
+    pub async fn created(&self) -> &u64 {
```
bilelmoussaoui, line 76: don't pass a copy type by reference

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl Item {
+    pub async fn new(
+        attributes: &HashMap<&str, &str>,
+        secret: Vec<u8>,
+        label: &str,
+        created: u64,
+        modified: u64,
+        collection_path: &OwnedObjectPath,
```
bilelmoussaoui, line 96:
```suggestion
        collection_path: OwnedObjectPath,
```

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+            path: OwnedObjectPath::try_from(format!(
+                "{}/items/{}",
+                collection_path.as_str(),
+                label
+            ))
+            .unwrap(),
+        }
+    }
+
+    pub async fn lookup_item_attributes(&self) -> HashMap<&str, &str> {
```
bilelmoussaoui, line 115: I am not sure about the purpose of this function

server/src/service/item.rs
```diff
@@ -0,0 +1,124 @@
...
+        attributes: &HashMap<&str, &str>,
+        secret: Vec<u8>,
+        label: &str,
+        created: u64,
+        modified: u64,
+        collection_path: &OwnedObjectPath,
+    ) -> Self {
+        // maps oo7::portal::Item to crate Item
+        let properties = Properties::for_item(label, attributes);
+        Item {
```
bilelmoussaoui, line 100:
```suggestion
        Self {
```

server/src/service/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod collection;
+pub mod item;
+pub mod service;
+pub mod session;
```
bilelmoussaoui, line 4: Should all of those be public? I don't think so

server/src/service/service.rs
```diff
@@ -0,0 +1,210 @@
...
+        alias: String,
+        #[zbus(object_server)] object_server: &ObjectServer,
+    ) -> fdo::Result<(ObjectPath, ObjectPath)> {
+        let _attributes = properties.attributes(); // expand and pass these to Collection:new
+        let collection = Collection::new(String::from("temp")).await; // temporarily
+        let path = format!("/org/freedesktop/secrets_/collection/{}", &alias);
+
+        let _ = object_server.at(path.clone(), collection).await;
+        let new_collection_path = ObjectPath::try_from(path).unwrap(); // temporarily
+        let prompt = ObjectPath::default().into(); // temp Prompt
```
bilelmoussaoui, line 48: I would suggest already creating a basic struct called Prompt, make it serialize as an ObjectPath and return it here. This way you can do `Prompt::default`, it would simplify finding out places where you still need to handle prompt later

server/src/service/service.rs
```diff
@@ -0,0 +1,210 @@
...
+            Ok(i) => i,
+            Err(_) => todo!(),
+        };
+
+        let mut unlocked: Vec<Item> = Vec::new();
+        let mut locked: Vec<Item> = Vec::new();
+
+        for item in items {
+            let attributes = item.attributes();
+            if attributes.get("locked").is_some() {
```
bilelmoussaoui, line 68: I am not sure you get "locked" in the attributes of the item. You should use the getter in the item.

server/src/service/service.rs
```diff
@@ -0,0 +1,210 @@
...
+        &mut self,
+        objects: Vec<OwnedObjectPath>,
+    ) -> fdo::Result<(Vec<OwnedObjectPath>, OwnedObjectPath)> {
+        // manage unlock state in memory
+        // when do we need to prompt?
+        let mut unlocked: Vec<OwnedObjectPath> = Vec::new();
+
+        'main: for object in objects {
+            for collection in &mut self.collections {
+                if collection.path().await.as_str() == object.as_str() {
```
bilelmoussaoui, line 91: No need to convert to a string to do a comparison

server/src/service/service.rs
```diff
@@ -0,0 +1,210 @@
...
+            for item in &items {
+                if in_collection.path().await.as_str() == item.as_str() {}
+            }
+        }
+    }
+
+    pub async fn read_alias(&self, name: &str) -> ObjectPath {
+        let mut ret = ObjectPath::default();
+        for collection in &self.collections {
+            if collection.label().await == name {
```
bilelmoussaoui, line 158: this is the alias of the collection, not it is label

server/src/service/service.rs
```diff
@@ -0,0 +1,210 @@
...
+        for in_collection in &mut self.collections {
+            if in_collection.path().await.as_str() == collection.as_str() {
+                in_collection.set_label(name.clone()).await;
+            }
+        }
+    }
+
+    #[dbus_interface(property, name = "Collections")]
+    pub async fn collections(&self) -> Vec<ObjectPath> {
+        let mut paths: Vec<ObjectPath> = Vec::new();
```
bilelmoussaoui, line 177: Same as what i said in the collection.items() case

server/src/service/service.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        for collection in &self.collections {
+            paths.push(collection.path().await.into())
+        }
+
+        paths
+    }
+
+    #[dbus_interface(signal)]
+    pub async fn collection_created(ctxt: &SignalContext<'_>) -> Result<(), Error>;
```
bilelmoussaoui, line 187: All of those signals should take an object path; and you should call them in the code where appropriate

server/src/service/service.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    #[dbus_interface(signal)]
+    pub async fn collection_deleted(ctxt: &SignalContext<'_>) -> Result<(), Error>;
+
+    #[dbus_interface(signal)]
+    pub async fn collection_changed(ctxt: &SignalContext<'_>) -> Result<(), Error>;
+}
+
+pub async fn init_secret_service() -> Service {
```
bilelmoussaoui, line 196: Could be a constructor of the struct itself if needed

server/src/service/session.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub async fn new(path: OwnedObjectPath) -> Self {
+        Self { path: path.into() }
+    }
+
+    pub async fn path(&self) -> &OwnedObjectPath {
+        &self.path
+    }
+}
+
+pub async fn init_session_service() -> Session {
```
bilelmoussaoui, line 35: Not needed

server/src/service/session.rs
```diff
@@ -0,0 +1,37 @@
...
+        object_server
+            .remove::<Self, &OwnedObjectPath>(&self.path)
+            .await?;
+        Ok(())
+    }
+}
+
+impl Session {
+    pub async fn new(path: OwnedObjectPath) -> Self {
+        Self { path: path.into() }
```
bilelmoussaoui, line 27:
```suggestion
        Self { path }
```

server/src/service/session.rs
```diff
@@ -0,0 +1,37 @@
+// org.freedesktop.Secret.Session
+
+use serde::{Deserialize, Serialize};
+use zbus::{dbus_interface, fdo, zvariant};
+use zvariant::{ObjectPath, OwnedObjectPath, Type};
+
+#[derive(Debug, Serialize, Deserialize, Type)]
```
bilelmoussaoui, line 7: Should it implement Serialize/Desrialize/Type?

server/src/main.rs
warusadura: Done, thanks!

server/src/main.rs
warusadura: Yeah, I agree.

server/src/service/collection.rs
warusadura: got it, thanks!

server/src/service/collection.rs
warusadura: Done.

server/src/service/collection.rs
warusadura: got it, thanks!

server/src/service/item.rs
warusadura: Done.

server/src/service/item.rs
warusadura: Done, thanks!

server/src/service/session.rs
warusadura: Done.

server/src/service/collection.rs
warusadura: understood, thanks!

server/src/service/collection.rs
warusadura: I guess not :)

server/src/service/collection.rs
warusadura: I tried replacing `u64` with `std::time::Duration` and also returning `u64`s as `std::time::Duration`s in getters. Both times I'm getting the following error,
```
error[E0277]: the trait bound `zbus::zvariant::Structure<'_>: From<Duration>` is not satisfied
........
#[dbus_interface(name = "org.freedesktop.Secret.Collection")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<Duration>` is not implemented for `zbus::zvariant::Structure<'_>`
.........
```

server/src/service/collection.rs
warusadura: got it, thanks!

server/src/service/collection.rs
warusadura: `properties.attributes().unwrap()` returns `&HashMap<String, String>` and `.create_item()` expects `attributes` parameter to be `HashMap<&str, &str>` type. So, that's why I'm doing this. Is this wrong?

server/src/service/collection.rs
warusadura: Since, `create_item` doesn't return anything, what's best way to handle/match `Err` variant here? I currently have,
```
match KEYRING
            .get()
            .unwrap()
            .create_item(label, attributes.clone(), secret, replace)
            .await
        {
            Ok(_) => (), 
            Err(_) => (), // best approach to handle this error?
        }
```

server/src/service/collection.rs
warusadura: you mean update [this](https://github.com/bilelmoussaoui/oo7/blob/83d44804dcffa79e4f823ceae320a276c4d8d985/client/src/portal/mod.rs#L202) to return portal::Item?

server/src/service/collection.rs
warusadura: Done!

server/src/main.rs
warusadura: I'm a bit unclear about this :) Is this only for `init_session_service`

server/src/service/item.rs
warusadura: Yeah I agree. But, the attributes getter returns `&HashMap<String, String>` and I need `HashMap<&str, &str>`

server/src/service/item.rs
warusadura: got it, thanks!

server/src/service/session.rs
warusadura: Yeah! Removed.

server/src/service/mod.rs
warusadura: `pub(crate)` perhaps?

server/src/service/collection.rs
warusadura: Done, thanks!

server/src/service/service.rs
warusadura: Done.

server/src/service/collection.rs
bilelmoussaoui: Possibly.

client/src/dbus/api/mod.rs
warusadura: Done, thanks!

server/src/main.rs
bilelmoussaoui: All of them, there shouldn't be any initialization done from the functions.

bilelmoussaoui: I mean there would be, but indirectly through constructing a Service would construct the types it needs to function.

server/src/main.rs
```diff
@@ -0,0 +1,46 @@
+// WIP: not ready for reviews
+
+pub mod service;
```
bilelmoussaoui, line 3: As mentioned before, maybe this should be renamed to daemon?

server/src/service/item.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+impl Item {
+    pub async fn new(
+        attributes: &HashMap<&str, &str>,
+        secret: Vec<u8>,
+        label: &str,
+        created: u64,
+        modified: u64,
```
bilelmoussaoui, line 100: I wonder why not just take a `portal::Item` here and avoid duplicating this data + avoid re-searching for the item everytime

server/src/service/service.rs
```diff
@@ -0,0 +1,205 @@
...
+#[derive(Serialize, Type, Debug)]
+pub struct Service {
+    pub(self) collections: Vec<Collection>,
+}
+
+#[dbus_interface(name = "org.freedesktop.Secret.Service")]
+impl Service {
+    pub async fn open_session(
+        &self,
+        // algorithm: &str,
```
bilelmoussaoui, line 24: We have an Alogrithm enum in the dbus API, use that

client/src/dbus/api/properties.rs
```diff
@@ -4,7 +4,7 @@ use serde::{
     ser::{Serialize, SerializeMap},
     Deserialize,
 };
-use zbus::zvariant::{self, OwnedValue, Type, Value};
```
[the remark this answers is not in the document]
warusadura: Apologies if the commit message is unclear.

In order to access `oo7::dbus::api::Properties` I have enabled the `unstable` feature. With this change in place I got a few new clippy warnings,

```
warning: unused import: `OwnedValue`
 --> client/src/dbus/api/properties.rs:7:28
  |
7 | use zbus::zvariant::{self, OwnedValue, Type, Value};
  |                            ^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: variable does not need to be mutable
  --> client/src/dbus/api/service.rs:40:13
   |
40 | ...   let mut stream = self.inner().receive_signal("CollectionCreated").awa...
   |           ----^^^^^^
   |           |
   |           help: remove this `mut`
   |
   = note: `#[warn(unused_mut)]` on by default

warning: variable does not need to be mutable
  --> client/src/dbus/api/service.rs:52:13
   |
52 | ...   let mut stream = self.inner().receive_signal("CollectionDeleted").awa...
   |           ----^^^^^^
   |           |
   |           help: remove this `mut`

warning: variable does not need to be mutable
  --> client/src/dbus/api/service.rs:64:13
   |
64 | ...   let mut stream = self.inner().receive_signal("CollectionChanged").awa...
   |           ----^^^^^^
   |           |
   |           help: remove this `mut`

```

So, I have introduced a few changes to fix to them through a separate [commit](https://github.com/bilelmoussaoui/oo7/pull/56/commits/5a9ac97c7bd1378030d576622d749e488f4847d6)

server/Cargo.toml
warusadura: Done, thanks!

server/src/main.rs

server/src/service/mod.rs
warusadura: Done, thanks!

server/src/service/session.rs
warusadura: Removed, thanks!

server/src/main.rs
warusadura: Done, thanks!

server/src/service/service.rs
warusadura: Done, thanks!

client/src/dbus/api/properties.rs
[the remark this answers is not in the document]
bilelmoussaoui: Indeed. Would you mind pulling this commit out of this PR so we can land it already?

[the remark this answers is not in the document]
warusadura: Yeah, sure.

server/src/service/daemon.rs
```diff
@@ -0,0 +1,191 @@
+//  org.freedesOnceCellktop.Secret.Service
+
```
bilelmoussaoui, line 2: I meant renaming the folder `server`, not the module name btw

warusadura: Done :)

daemon/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "server"
```
bilelmoussaoui, line 2:
```suggestion
name = "oo7-daemon"
```

server/src/service/item.rs
warusadura: The issue is `oo7::portal::Item` doesn't have a `path:OwnedObjectPath,` field. Since we're dealing with ObjectPaths, for ex: for `CreateItem` we have to return item's ObjectPath. So, that's why I'm doing this mapping (`oo7::portal::Item` to `crate::service::item`).

bilelmoussaoui: Sure, what I meant is that `crate::service::Item` could be 
```rust
struct Item {
   inner: portal::Item,
   path: OwnedObjectPath,
}
```

warusadura: Understood clearly, thank you :)

daemon/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+        .serve_at(SECRET_SERVICE_OBJECTPATH, secret_service)?
+        //.serve_at(SECRET_COLLECTION_OBJECTPATH, collection)?
+        //.serve_at(SECRET_SESSION_OBJECTPATH, session)?
+        .build()
+        .await?;
+
+    let _keyring = match Keyring::load_default().await {
+        Ok(keyring) => KEYRING.set(keyring),
+        Err(_) => todo!(), // call some init to create default keyring
+    };
```
bilelmoussaoui, line 39: All of this should be done inside `Service::new()` except the `serve_at` bit, which you would do at a later stage by calling `service.run().await?;`

bilelmoussaoui: By doing so, the KEYRING, is something that no longer would be a Lazy lock, instead you would add it as a field to the `Service`, like `Arc<Keyring>` and pass it down the line to the collections or items if they need access to it.

daemon/Cargo.toml
warusadura: done.

Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
   "client",
   "cli",
   "portal",
+  "oo7-daemon",
```
bilelmoussaoui, line 8: I meant renaming the project in Cargo.toml, not the directory itself. The name was good there.

oo7-daemon/src/service/item.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    pub fn path(&self) -> ObjectPath {
+        self.path.to_owned().into()
+    }
+}
+
+impl Item {
+    pub async fn new(
+        attributes: &HashMap<&str, &str>,
```
bilelmoussaoui, line 96: Use the new AsAttributes trait

oo7-daemon/src/service/service.rs
```diff
@@ -0,0 +1,191 @@
...
+            }
+        }
+
+        Ok((unlocked, locked))
+    }
+
+    #[dbus_interface(out_args("unlocked", "prompt"))]
+    pub async fn unlock(
+        &mut self,
+        objects: Vec<OwnedObjectPath>,
```
bilelmoussaoui, line 83: You can use IsUnlockable here I guess. Also the function currently assumes the objects are collections which is not true. They can be items as well.

Cargo.toml
warusadura: Done.

daemon/src/main.rs
warusadura: > All of this should be done inside `Service::new()` except the `serve_at` bit, which you would do at a later stage by calling `service.run().await?;`

Hey,

I have implemented the `Service::new()`: https://github.com/bilelmoussaoui/oo7/blob/10c73b94103ff0b9fb592bf0a3a765ceee9804fc/daemon/src/service/service.rs#L199 To implement the `service.run()` I have included `connection: ConnectionBuilder` field in `Service` struct. To this point everything worked well. But, `serve_at` requires the ownership of `service` (or whatever we pass it). So, I think `run()` can't be implemented as a instance method :)

> By doing so, the KEYRING, is something that no longer would be a Lazy lock, instead you would add it as a field to the `Service`, like `Arc<Keyring>` and pass it down the line to the collections or items if they need access to it.

What's the best approach to pass down `service`'s `keyring: Arc<Keyring>`. The only approach I can think and tried (it didn't go well) is use the return value of `service.keyring()` as `OnceLock` :)

bilelmoussaoui: > To implement the service.run() I have included connection: ConnectionBuilder field in Service struct

You don't need to, you should setup the ConnectionBuilder inside `service.run()` + make the function take `self` and not `&self` and serve_at would work by passing it `self`

server/src/service/collection.rs
```diff
@@ -0,0 +1,192 @@
...
+use serde::{Serialize, Serializer};
+use zbus::{dbus_interface, zvariant, Error, ObjectServer, SignalContext};
+use zvariant::{ObjectPath, OwnedObjectPath, Type};
+
+use crate::{service::item, KEYRING};
+
+#[derive(Default, Debug, Type)]
+#[zvariant(signature = "o")]
+pub struct Collection {
+    pub items: Vec<item::Item>,
```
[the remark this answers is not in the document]
warusadura: done, thanks!

oo7-daemon/src/service/service.rs
warusadura: With this change: https://github.com/bilelmoussaoui/oo7/blob/004f98a69f98456d86083b0b5f9264c4471c6c1e/daemon/src/service/service.rs#L212 
oo7-daemon is no longer responds to `busctl`. I usually run `busctl --user tree org.freedesktop.secrets_` after a major change. It's no longer there. I'm not sure what's causing this :)

daemon/src/main.rs
warusadura: With this change: https://github.com/bilelmoussaoui/oo7/blob/004f98a69f98456d86083b0b5f9264c4471c6c1e/daemon/src/service/service.rs#L212 
oo7-daemon is no longer responds to `busctl`. I usually run `busctl --user tree org.freedesktop.secrets_` after a major change. It's no longer there. I'm not sure what's causing this :)

daemon/src/service/service.rs
```diff
@@ -0,0 +1,226 @@
...
+            .expect("Failed to request name")
+            .serve_at(SECRET_SERVICE_OBJECTPATH, self)
+            .expect("Failed to serve the interface")
+            .build()
+            .await
+        {
+            Ok(_) => Ok(()),
+            Err(e) => Err(e),
+        }
+    }
```
bilelmoussaoui, line 225:
```suggestion
        zbus::connection::Builder::session()?
            .name("org.freedesktop.secrets_")?
            .serve_at(SECRET_SERVICE_OBJECTPATH, self)?
            .build()
            .await.map_err(From::from)
    }
```
this should be better

warusadura: thanks :)

daemon/src/main.rs
bilelmoussaoui: I am not sure what would cause that, the code changes looks sane to me.

warusadura: yeah, I agree :) I'll look into this again.

daemon/src/service/service.rs
warusadura: `.await.map_err(From::from)` is introducing some wired errors,

```
.....
error[E0790]: cannot call associated function on trait without specifying the corresponding `impl` type
   --> daemon/src/service/service.rs:217:28
    |
217 |             .await.map_err(From::from)
    |                            ^^^^^^^^^^ cannot call associated function of trait

....
```

I'll get back to this later.

updated run(),

```
    pub async fn run(self) -> Result<(), Error> {
        match connection::Builder::session()?
            .name("org.freedesktop.secrets_")?
            .serve_at(SECRET_SERVICE_OBJECTPATH, self)?
            .build()
            .await // .await.map_err(From::from)
        {
            Ok(_) => Ok(()),
            Err(e) => Err(e),
        }
    }   
```

server/src/service/collection.rs
warusadura: done.

server/src/service/collection.rs
warusadura: this was just temporarily. I have re-implemented the Collection::new.

server/src/service/service.rs
warusadura: got it, done.

server/src/service/service.rs
warusadura: updated the entire method.

bilelmoussaoui:
Closing in favour of #73

PR #58, opened 2023-12-31
Stop using the async-std crate

sophie-h:
There hasn't been updates for 1 1/2 years and it potentially pulls dependencies that are not needed.

Replace it with the maintained parts that async-std is made of.


bilelmoussaoui:
Thanks!

PR #59, opened 2024-01-02
re-org: Use cargo workspaces

bilelmoussaoui:
Move oo7 into client subdirectory, rename keyring-util to cli and move examples to a separate project

Especially that we will soon have a server & portal implementations in-tree.


bilelmoussaoui:
The CI tests failure is unrelated to this change, so I merged it.

PR #63, opened 2024-01-05
client/dbus: Support signals in public API

### LINKED ISSUE #57: Connecting to SecretService signals
I would like to connect to secret service signals. With what we have the naive implementation would probably look like

```
    pub async fn receive_item_changed(
        &self,
    ) -> Result<impl futures_util::Stream<Item = Item<'_>>, Error> {
        self.inner.receive_item_changed().await.map(|stream| {
            stream.map(|item| {
                Item::new(
                    Arc::clone(&self.service),
                    Arc::clone(&self.session),
                    self.algorithm,
                    item,
                    self.aes_key.as_ref().map(Arc::clone),
                )
            })
        })
    }
```

However, oo7 only deals with `Item<'static>` which we don't get here. I think it's a zbus limitation at the moment.

To work around this, I would like to add `Collection::item_for_path` and `Item::path` functions, since together they would allow us to get a static version of the item.

Any thoughts? Otherwise, I would implement this.

sophie-h:
Closes #57

### DIFF
--- a/client/src/dbus/item.rs
+++ b/client/src/dbus/item.rs
@@ -4,6 +4,7 @@ use std::{collections::HashMap, sync::Arc, time::Duration};
 use async_lock::RwLock;
 #[cfg(feature = "tokio")]
 use tokio::sync::RwLock;
+use zbus::zvariant::ObjectPath;
 use zeroize::Zeroizing;
 
 use super::{api, Algorithm, Error};
@@ -194,5 +195,10 @@ impl<'a> Item<'a> {
             self.service.lock(&[self.inner.inner().path()]).await?;
             Ok(())
         }
+    }
+
+    /// Returns item path
+    pub fn path(&self) -> &ObjectPath<'_> {
+        self.inner.inner().path()
     }
 }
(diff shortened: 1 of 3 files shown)


client/src/dbus/item.rs
```diff
@@ -195,4 +196,9 @@ impl<'a> Item<'a> {
             Ok(())
         }
     }
+
+    /// Returns item path
+    pub fn path(&self) -> &ObjectPath<'_> {
+        self.inner.inner().path()
+    }
```
bilelmoussaoui, line 203: If we are exposing the path in the dbus::api::Item, I would expect the same thing to be done collections as well. Or maybe just making it possible to compare an `Item<'_>` and `OwnedObjectPath` without exposing the internals as API.

bilelmoussaoui (commented):
Otherwise lgtm, thanks a lot!

client/src/dbus/item.rs
sophie-h: I'm not sure if obscuring the compare operation is a great idea. I'm also not sure if it would work properly since neither Eq nor ObjectPath are defined in oo7.

About Collection: The Item path is added in the MR since it also gets exposed elsewhere with API introduced in this MR. Therefore I would suggest moving the discussion for Collection elsewhere since the Collection path doesn't appear in any API yet.

PR #68, opened 2024-02-09
portal: update create_item() to return portal::Item

warusadura:
This change updates the `oo7::portal::Keyring::create_item()` to return
`oo7::portal::Item` instead of `()`

We need this change for the server side implementation. In server/daemon
side we are wrapping the return value of `create_item()` inside a struct.


bilelmoussaoui:
Lgtm

for future commits, please describe why the change is needed not what the change itself. In this case, I would have instead mentioned the use case of wrapping the portal::Item` in the server side implementation.

PR #69, opened 2024-02-11
portal: updates oo7::portal::Keyring API "attributes" parameter

warusadura:
This change updates the following `oo7::portal::Keyring` API's
"attributes" parameter type,
`search_items()`
`lookup_item()`
`delete()`
`create_item()`

We need this change for server side implementation.
See: https://github.com/bilelmoussaoui/oo7/pull/56#discussion_r1482705886

### DIFF
--- a/client/src/portal/mod.rs
+++ b/client/src/portal/mod.rs
@@ -160,7 +160,13 @@ impl Keyring {
     }
 
     /// Search items matching the attributes.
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
This is no longer needed with #70, thanks nonetheless ! Please confirm that it works correctly for the server use case as well

PR #72, opened 2024-02-15
client/dbus: make Algorithm enum public under unstable feature

warusadura:
We need this change for server side implementation.
See: https://github.com/bilelmoussaoui/oo7/pull/56#discussion_r1482719151

### DIFF
--- a/client/src/dbus/mod.rs
+++ b/client/src/dbus/mod.rs
@@ -55,7 +55,11 @@ pub mod api;
 mod api;
 
 mod algorithm;
+#[cfg(not(feature = "unstable"))]
 pub(crate) use algorithm::Algorithm;
+#[cfg(feature = "unstable")]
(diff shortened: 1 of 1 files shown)


client/src/dbus/mod.rs
```diff
@@ -55,7 +55,7 @@ pub mod api;
 mod api;
 
 mod algorithm;
-pub(crate) use algorithm::Algorithm;
+pub use algorithm::Algorithm;
```
bilelmoussaoui, line 58: It is not used as public API, make it behind the unstable flag similar to `mod api;` / `pub mod api;`

warusadura: done, thanks!

bilelmoussaoui (approved):

bilelmoussaoui:
thanks!

PR #73, opened 2024-02-15
New gnome keyring

bilelmoussaoui:
Switches #56 to use a branch from this repo so we can easily collaborate on it.

progress to-date:

- [x] D-Bus Secret Service API implementation - complete
- [x] Default keyring (login.keyring) support - complete
- [ ] Multiple keyring support - not done


warusadura:
Does it make sense to store the `session` along with an Item. As in,
```
pub struct Item {
    inner: RwLock<portal::Item>,
    path: OwnedObjectPath,
    keyring: Arc<Keyring>,
    locked: bool,
    session: OwnedObjectPath,
}
```

and then, store the same session inside Collection,
```
pub struct Collection {
    keyring: Arc<Keyring>,
    pub(crate) items: RwLock<Vec<super::item::Item>>,
    alias: RwLock<String>,
    label: String,
    locked: AtomicBool,
    created: Duration,
    modified: Duration,
    path: OwnedObjectPath,
    sessions: Vec<OwnedObjectPath>,
}

with this way we get access to all the available sessions and we can retrieve a session based on an objectpath. wdyt?

**Comment by bilelmoussaoui** [comment]
> Does it make sense to store the `session` along with an Item. As in,
> 
> ```
> pub struct Item {
>     inner: RwLock<portal::Item>,
>     path: OwnedObjectPath,
>     keyring: Arc<Keyring>,
>     locked: bool,
>     session: OwnedObjectPath,
> }
> ```
> 
> and then, store the same session inside Collection,
> 
> ```
> pub struct Collection {
>     keyring: Arc<Keyring>,
>     pub(crate) items: RwLock<Vec<super::item::Item>>,
>     alias: RwLock<String>,
>     label: String,
>     locked: AtomicBool,
>     created: Duration,
>     modified: Duration,
>     path: OwnedObjectPath,
>     sessions: Vec<OwnedObjectPath>,
> }
> ```
> 
> with this way we get access to all the available sessions and we can retrieve a session based on an objectpath. wdyt?

Storing the session is required yes, but not just it ObjectPath. Instead you need to store the Sesssion itself which would contain more information than just the object path. Ideally we can store it as `HashMap<OwnedObjectPath, Session>`, same thing for collections & items I would say.

**Comment by warusadura** [comment]
> > Does it make sense to store the `session` along with an Item. As in,
> > ```
> > pub struct Item {
> >     inner: RwLock<portal::Item>,
> >     path: OwnedObjectPath,
> >     keyring: Arc<Keyring>,
> >     locked: bool,
> >     session: OwnedObjectPath,
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then, store the same session inside Collection,
> > ```
> > pub struct Collection {
> >     keyring: Arc<Keyring>,
> >     pub(crate) items: RwLock<Vec<super::item::Item>>,
> >     alias: RwLock<String>,
> >     label: String,
> >     locked: AtomicBool,
> >     created: Duration,
> >     modified: Duration,
> >     path: OwnedObjectPath,
> >     sessions: Vec<OwnedObjectPath>,
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with this way we get access to all the available sessions and we can retrieve a session based on an objectpath. wdyt?
> 
> Storing the session is required yes, but not just it ObjectPath. Instead you need to store the Sesssion itself which would contain more information than just the object path. Ideally we can store it as `HashMap<OwnedObjectPath, Session>`, same thing for collections & items I would say.

Understood. Yeah I agree using a HashMap make sense. I think it make sense store all these inside `struct Service`

File: server/src/daemon/service.rs [hunk 1]
```diff
@@ -0,0 +1,265 @@
...
+
+use super::{
+    collection::Collection, error::ServiceError, prompt::Prompt, session::Session, Result,
+};
+
+#[derive(Debug)]
+pub struct Service {
+    collections: Vec<Collection>,
+    keyring: Arc<Keyring>,
+    cnx: Mutex<Option<zbus::Connection>>,
```
warusadura, line 31: I was thinking,

```
diff --git a/server/src/daemon/service.rs b/server/src/daemon/service.rs
index d757670..e221b42 100644
--- a/server/src/daemon/service.rs
+++ b/server/src/daemon/service.rs
@@ -21,14 +21,22 @@ use zbus::{
 };
 
 use super::{
-    collection::Collection, error::ServiceError, prompt::Prompt, session::Session, Result,
+    collection::Collection, error::ServiceError, item, prompt::Prompt, session::Session, Result,
 };
 
+#[derive(Debug)]
+pub struct ServiceManager {
+    collections: HashMap<OwnedObjectPath, Collection>,
+    items: HashMap<OwnedObjectPath, item::Item>,
+    sessions: HashMap<OwnedObjectPath, Session>,
+}
+
 #[derive(Debug)]
 pub struct Service {
     collections: Vec<Collection>,
     keyring: Arc<Keyring>,
     cnx: Mutex<Option<zbus::Connection>>,
+    manager: Arc<ServiceManager>,
 }
 
 #[zbus::interface(name = "org.freedesktop.Secret.Service")]
```

bilelmoussaoui: This might make sense yes, although, I guess the collections are not supposed to be stored there or the items. As you don't want to mix items from different collections. 

So maybe just store the sessions there for now and see what requirements you end up needing with time.

warusadura: got it, thanks!

server/src/daemon/collection.rs
```diff
@@ -123,20 +125,16 @@ impl Collection {
 }
 
 impl Collection {
-    // temporarily creates a generic Collection object
-    pub fn new(label: &str, alias: &str, keyring: Arc<Keyring>) -> Self {
+    pub fn new(label: &str, alias: &str, created: u64, keyring: Arc<Keyring>) -> Self {
```
bilelmoussaoui, line 128: There is no point in having DateTime, converting to seconds just convert it back here. So please use `DateTime` type here instead

bilelmoussaoui: The thing is this constructor should be shared with collections you just created and collections that were already available

warusadura: done.

warusadura: > There is no point in having DateTime, converting to seconds just convert it back here. So please use `DateTime` type here instead

done.

warusadura: > The thing is this constructor should be shared with collections you just created and collections that were already available

fixed via: https://github.com/bilelmoussaoui/oo7/pull/73/commits/d759844facc8ecf68d34b19c57254b38fd96e2ec

server/src/daemon/service.rs
```diff
@@ -24,11 +24,33 @@ use super::{
...
+
+    pub fn sessions(&self) -> &HashMap<OwnedObjectPath, Session> {
+        &self.sessions
+    }
+
+    pub fn set_sessions(&mut self, path: OwnedObjectPath, session: Session) {
+        self.sessions.insert(path, session);
+    }
+}
+
```
warusadura, line 47: Does it make sense to put this inside a new file: `src/daemon/service_manager.rs`

server/src/daemon/service.rs
```diff
@@ -24,11 +24,33 @@ use super::{
...
 #[derive(Debug)]
 pub struct Service {
     collections: Vec<Collection>,
     keyring: Arc<Keyring>,
     cnx: Mutex<Option<zbus::Connection>>,
+    manager: Arc<ServiceManager>,
```
warusadura, line 53: Is this approach looks clean to you? For `Service` it make sense. But for `Collection` and `Item`, I'm not so sure.

server/src/daemon/service.rs
```diff
@@ -24,11 +24,33 @@ use super::{
     collection::Collection, error::ServiceError, prompt::Prompt, session::Session, Result,
 };
 
+#[derive(Debug)]
+pub struct ServiceManager {
+    sessions: HashMap<OwnedObjectPath, Session>,
```
warusadura, line 29: `Session` already has a `path` filed attached to it. Can't we use that when we need to access `path` and store `sessions` as a `Vec<Session>`?

server/src/daemon/service.rs
bilelmoussaoui: > For Service it make sense. But for Collection and Item, I'm not so sure

You might need it there as well. You get the session path in get secrets so you have to match the path back to the already open session.

Although, not useful right now without making use of it.

server/src/daemon/service.rs
bilelmoussaoui: Getting a specific item from the hashmap through it path would be faster than iterating the vec i would say

warusadura: understood, thanks!

server/src/daemon/service.rs
warusadura: yeah, got it. Thanks!

server/src/daemon/service.rs
```diff
@@ -24,11 +24,33 @@ use super::{
...
+}
+
+impl ServiceManager {
+    pub fn new() -> Self {
+        Self {
+            sessions: HashMap::new(),
+        }
+    }
+
+    pub fn sessions(&self) -> &HashMap<OwnedObjectPath, Session> {
```
bilelmoussaoui, line 39: Don't add a getter that is not used, yet. You would just need an API to get a Option<&Session> from a session path

server/src/daemon/service.rs
warusadura: yeah, I agree :)

server/src/daemon/service.rs
```diff
@@ -24,11 +24,33 @@ use super::{
...
+        &self.sessions
+    }
+
+    pub fn set_sessions(&mut self, path: OwnedObjectPath, session: Session) {
```
bilelmoussaoui, line 43: I would call it insert_session and the method can't be mutable otherwise you would need mutable access everywhere which would block the async tasks. Instead use a mutex/rwlock

server/src/daemon/service.rs
```diff
@@ -261,4 +268,8 @@ impl Service {
             .await?;
         Ok(())
     }
+
+    pub fn set_collections(&mut self, collection: Collection) {
```
bilelmoussaoui, line 298: This method is not really useful

server/src/daemon/service.rs
```diff
@@ -52,7 +52,7 @@ impl Service {
 
     #[zbus(out_args("collection", "prompt"))]
     pub async fn create_collection(
-        &self,
+        &mut self,
```
bilelmoussaoui, line 78: This shouldn't become mut, instead make the collections field a mutex

server/src/daemon/service.rs
warusadura: understood, will do.

server/src/daemon/service.rs
```diff
@@ -66,6 +66,13 @@ impl Service {
                 .unwrap(),
             Arc::clone(&self.keyring),
         );
+        self.set_collections(Collection::new(
+            collection.label(),
+            alias,
+            Duration::from_secs(collection.created()),
```
bilelmoussaoui, line 96: I would rename the current created/modified methods to return a Duration type and have variants that are exposed as dbus properties that are called `created_as_secs` / `modified_as_secs`

server/src/daemon/service.rs
warusadura: done.

server/src/daemon/service.rs
warusadura: Removed, thanks!

server/src/daemon/service.rs
warusadura: done, thanks!

server/src/daemon/service.rs
warusadura: done, thanks!

server/src/daemon/service.rs
warusadura: wip.

server/src/daemon/service.rs
```diff
@@ -0,0 +1,279 @@
...
+                        let secret = item.secret(session.clone()).await?;
+                        secrets.insert(path.to_owned(), secret);
+                    }
+                }
+            }
+        }
+        Ok(secrets)
+    }
+
+    pub fn read_alias(&self, name: &str) -> ObjectPath {
```
warusadura, line 197: need to update. will push a new commit to fix this.

server/src/daemon/service.rs
```diff
@@ -0,0 +1,279 @@
...
+                if c.label() == name {
+                    Some(c.path())
+                } else {
+                    None
+                }
+            })
+            .unwrap_or_default()
+    }
+
+    pub fn set_alias(
```
warusadura, line 210: need to update. will push a new commit to fix this.

server/src/daemon/service.rs
```diff
@@ -0,0 +1,279 @@
...
+            }
+            None => {
+                tracing::error!("Collection {path} not found");
+                Err(ServiceError::NoSuchObject)
+            }
+        }
+    }
+
+    #[zbus(property, name = "Collections")]
+    pub async fn collections(&self) -> Vec<ObjectPath> {
```
warusadura, line 229: need to update. will push a new commit to fix this.

server/src/daemon/service_manager.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+impl ServiceManager {
+    pub fn new() -> Self {
+        Self {
+            sessions: RwLock::new(HashMap::new()),
+        }
+    }
+
+    pub async fn session(&self, path: ObjectPath<'_>) -> Option<&Session> {
```
warusadura, line 20: > Don't add a getter that is not used, yet. You would just need an API to get a Option<&Session> from a session path

done.

How can I fix the following error,
```
error[E0515]: cannot return value referencing temporary value
  --> server/src/daemon/service_manager.rs:21:9
   |
21 |         self.sessions.read().await.get(&path.into())
   |         --------------------------^^^^^^^^^^^^^^^^^^
   |         |
   |         returns a value referencing data owned by the current function
   |         temporary value created here
```

server/src/daemon/service.rs
```diff
@@ -0,0 +1,287 @@
...
+        &self,
+        algorithm: Algorithm,
+        input: Value<'_>,
+    ) -> Result<(OwnedValue, OwnedObjectPath)> {
+        let client_public_key = match algorithm {
+            Algorithm::Plain => None,
+            Algorithm::Encrypted => Some(Key::from(input)),
+        };
+        let (session, key) = Session::new(client_public_key);
+        // TODO: clean up the default generated key
```
warusadura, line 49: what do you mean by this :) ?

server/src/daemon/service_manager.rs
bilelmoussaoui: Not if you wrap the session key in Arc<Key>

warusadura: I have updated  `client_public_key: Arc<Option<Key>>`. So, cloning a `Session` works now  But, I'm still getting the following error at line: [21](https://github.com/bilelmoussaoui/oo7/blob/95ebf467092569d14dc9d1cb9d1909822bf87adc/server/src/daemon/service_manager.rs#L21),
`returns a value referencing data owned by the current function`.

server/src/daemon/service.rs
```diff
@@ -48,6 +48,11 @@ impl Service {
         let (session, key) = Session::new(client_public_key, Arc::clone(&self.manager));
         // TODO: clean up the default generated key
         // TODO call self.manager.set_sessions();
+        self.manager
+            .lock()
+            .unwrap()
+            .insert_session(session.path().to_owned(), session.to_owned())
+            .await;
```
warusadura, line 55: Ran in to a new error here. Error: https://github.com/bilelmoussaoui/oo7/actions/runs/7987306810/job/21809515211?pr=73#step:4:369

warusadura: Looks like, I "can't hold the lock while you call .await. You have to release it before. If you want to hold if during an .await" (StackOverflow)[1]

[1]: [answer](https://stackoverflow.com/a/75080655) second comment.

warusadura: So, I think the only way to fix this is to remove `async` from `insert_session`. That means I have to remove `RwLock` [here](https://github.com/bilelmoussaoui/oo7/blob/01d1fa7d10c99f3cc9946e6144a0f6571619d619/server/src/daemon/service_manager.rs#L10). Could this be an issue?

warusadura: I think this won't be an issue, since `ServiceManager` is wrapped around `Mutex`.

server/src/daemon/item.rs
```diff
@@ -52,7 +52,9 @@ impl Item {
     pub async fn secret(&self, session: ObjectPath<'_>) -> Result<SecretInner> {
         let inner = self.inner.read().await;
         let secret = inner.secret();
-        // TODO: get the session based on it object path and figure out the parameters
+        // TODO: get the session based on it object path and figure out the parameters -
+        // done
+        let session = self.manager.lock().unwrap().session(session);
         // to pass, also, what we should do about the content-type?
```
warusadura, line 58: any thoughts on this?

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,199 @@
...
+
+        let secret = if aes_key.is_none() {
+            Secret::new(Arc::new(session), value, content_type.as_str()) // private function
+        } else {
+            Secret::new_encrypted(
+                Arc::new(session), // expected `Session<'_>`, found `Session`
+                value,
+                content_type.as_str(),
+                &aes_key.unwrap(),
+            )
```
warusadura, line 90: is this approach make sense to encrypt (or not) the data using session? @bilelmoussaoui 
one issue:
- `Secret::new_encrypted()` expecting a `Session<'_>` and I'm passing a `Session` :) is adding a lifetime to [Session](https://github.com/bilelmoussaoui/oo7/blob/1419233f39ef1807b7ffd5aa62ce40ab97f0c035/server/src/daemon/session.rs#L14) could fix this mismatched types issue?

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,199 @@
...
+        };
+
+        let item = self
+            .keyring
+            .create_item(label, &attributes, secret, replace) // AsRef<[u8]>` is not implemented
```
warusadura, line 95: how can I fix the following error generated here,
`the trait bound oo7::dbus::api::Secret<'_>: AsRef<[u8]> is not satisfied`

server/src/daemon/service.rs
```diff
@@ -0,0 +1,311 @@
...
+            Arc::clone(&self.keyring),
+            Arc::clone(&self.manager),
+        );
+        self.collections.write().await.push(Collection::new(
+            collection.label(),
+            alias,
+            *collection.created(),
+            Arc::clone(&self.keyring),
+            Arc::clone(&self.manager),
+        ));
```
bilelmoussaoui, line 103: You already create the same collection above? this is wrong.

server/src/daemon/service.rs
```diff
@@ -0,0 +1,311 @@
...
+        ctxt: &SignalContext<'_>,
+        collection: ObjectPath<'_>,
+    ) -> zbus::Result<()>;
+}
+
+impl Service {
+    pub async fn new() -> Self {
+        Self {
+            collections: RwLock::new(Vec::new()),
+            keyring: Arc::new(Keyring::load_default().await.unwrap()),
```
bilelmoussaoui, line 292: The unwrap here makes me nervous, should fallback to create the default if not available.

server/src/daemon/service.rs
```diff
@@ -0,0 +1,311 @@
...
+            cnx: Default::default(),
+            manager: Arc::new(Mutex::new(ServiceManager::new())),
+            sessions_counter: RwLock::new(0),
+        }
+    }
+
+    pub async fn run(self) -> Result<()> {
+        let cnx = zbus::connection::Builder::session()?
+            .name(oo7::dbus::api::Service::DESTINATION.unwrap())?
+            //.name("org.freedesktop.secrets.Devel")?
```
bilelmoussaoui, line 302: Instead, define all those values in main.rs as constants, and make them change based on whether you are doing a release or debug build I would say. It would make testing much simpler

server/src/daemon/service_manager.rs
```diff
@@ -0,0 +1,30 @@
...
+pub struct ServiceManager {
+    sessions: HashMap<OwnedObjectPath, Session>,
+}
+
+impl ServiceManager {
+    pub fn new() -> Self {
+        Self {
+            sessions: HashMap::new(),
+        }
+    }
```
bilelmoussaoui, line 17: This is basically `Derive(Default)`

server/src/daemon/session.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub fn new(
+        aes_key: Option<Key>,
+        manager: Arc<Mutex<ServiceManager>>,
+        sessions_counter: i32,
+    ) -> Self {
+        Self {
+            aes_key: Arc::new(aes_key),
+            path: OwnedObjectPath::try_from(format!(
+                "{}{}{}",
+                SECRET_SESSION_PREFIX, "s", sessions_counter
```
bilelmoussaoui, line 42: What is `s` here?

server/src/daemon/session.rs
```diff
@@ -0,0 +1,56 @@
...
+            ))
+            .unwrap(),
+            manager,
+        }
+    }
+
+    pub fn aes_key(&self) -> &Option<Key> {
+        self.aes_key.as_ref()
+    }
```
bilelmoussaoui, line 51:
```suggestion
    pub fn aes_key(&self) -> Option<&Key> {
        self.aes_key.as_deref()
    }
```

bilelmoussaoui (commented):
Please also fix the clippy warnings. The implementation starts to look decent but we still don't handle things like Prompt.

server/src/daemon/session.rs
warusadura: `s` comes from [here](https://gitlab.gnome.org/GNOME/gnome-keyring/-/blob/master/daemon/dbus/gkd-secret-session.c?ref_type=heads#L313). As in `/org/freedesktop/secrets/session/s1`

bilelmoussaoui: Then there is no point in not writing it directly in `{}s{}`

client/src/crypto/mod.rs
```diff
@@ -2,6 +2,9 @@
 mod native;
 #[cfg(feature = "native_crypto")]
 pub(crate) use native::*;
+#[cfg(feature = "unstable")]
+#[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
+pub use native::*;
```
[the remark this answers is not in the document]
bilelmoussaoui: No, duplicate what you did for openssl.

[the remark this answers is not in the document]
warusadura: Got it, thanks!

client/src/key.rs
```diff
@@ -24,7 +24,7 @@ impl AsMut<[u8]> for Key {
 }
 
 impl Key {
-    pub(crate) fn new(key: Vec<u8>) -> Self {
+    pub fn new(key: Vec<u8>) -> Self {
```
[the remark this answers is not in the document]
warusadura: Removed, thanks!

server/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
+oo7 = { path = "../client", features = ["unstable"] }
+serde = "1.0"
+tokio = { version = "1", features = ["full"] } # look for specific features
+tracing = "0.1"
+tracing-subscriber.workspace = true
+zbus = "4.0.0"
```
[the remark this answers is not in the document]
warusadura: Done.

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,199 @@
...
+        Ok((path, prompt))
+    }
+
+    #[zbus(property, name = "Items")]
+    pub async fn items(&self) -> Vec<OwnedObjectPath> {
+        self.items
+            .read()
+            .await
+            .iter()
+            .map(|item| item.path().into())
```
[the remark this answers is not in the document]
warusadura: Done, thanks!

server/src/daemon/service_manager.rs
warusadura: Done, thanks!

server/src/daemon/item.rs
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+
+    pub(crate) fn path(&self) -> ObjectPath {
+        self.path.as_ref()
+    }
+
+    pub(crate) fn parameters(&self) -> Vec<u8> {
+        self.parameters.to_owned()
+    }
```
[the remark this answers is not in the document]
warusadura: Done, thanks!

server/src/daemon/service.rs
warusadura: So, the reason why I'm doing this is, I have to update the `self.collections` [Vec](https://github.com/bilelmoussaoui/oo7/blob/b2abf3c42aa967882e3e34172c82ff06e3bbb608/server/src/daemon/service.rs#L31) when a new collection is created right. Since we need this just created `collection` later at `object_server.at()`, we can't just passed to it `push()`. So, we have to clone it. But, again `Collection` can't implement the `Clone` trait (Clone is not implemented for tokio ...). So, I can't do, `self.collections.write().await.push(collection.clone())`.

To fix this I have tried to clone the just created collection by hand using the values retrieved from getters :)

server/src/daemon/service.rs
warusadura: Fixed with https://github.com/bilelmoussaoui/oo7/pull/73/commits/b2abf3c42aa967882e3e34172c82ff06e3bbb608

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    pub async fn set_alias(&self, ctxt: &zbus::SignalContext<'_>, alias: &str) -> Result<()> {
+        *self.alias.write().await = alias.to_owned();
+        Service::collection_changed(ctxt, self.path.as_ref()).await?;
+        Ok(())
+    }
+
+    pub async fn set_locked(&self, locked: bool) {
+        self.locked
```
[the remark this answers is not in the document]
warusadura: I couldn't find collection_changed signal in the dbus specs: https://specifications.freedesktop.org/secret-service/latest/re02.html

server/src/daemon/service.rs
```diff
@@ -289,7 +289,10 @@ impl Service {
     pub async fn new() -> Self {
         Self {
             collections: RwLock::new(Vec::new()),
-            keyring: Arc::new(Keyring::load_default().await.unwrap()),
+            keyring: Arc::new(match Keyring::load_default().await {
+                Ok(keyring) => keyring,
+                Err(_) => todo!(),
```
warusadura, line 294: I'm not sure how to create the default keyring here :)

server/src/daemon/service.rs
bilelmoussaoui: The todo! is not a fix.

server/src/daemon/session.rs
bilelmoussaoui: ah it is wrapped with an Arc... leave it as is for now I guess.

server/src/daemon/service.rs
bilelmoussaoui: Even load_default call is actually wrong. You need to call `.load` instead as you don't use the portal's key to encrypt the keyring.

server/src/daemon/service.rs
warusadura: So, here is the full error,

```
error[E0277]: the trait bound `tokio::sync::RwLock<Vec<item::Item>>: Clone` is not satisfied
  --> server/src/daemon/collection.rs:27:5
   |
24 | #[derive(Clone, Debug)]
   |          ----- in this derive macro expansion
...
27 |     pub(crate) items: RwLock<Vec<super::item::Item>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `tokio::sync::RwLock<Vec<item::Item>>`
   |
   = note: this error originates in the derive macro `Clone` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `tokio::sync::RwLock<std::string::String>: Clone` is not satisfied
  --> server/src/daemon/collection.rs:28:5
   |
24 | #[derive(Clone, Debug)]
   |          ----- in this derive macro expansion
...
28 |     alias: RwLock<String>,
   |     ^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `tokio::sync::RwLock<std::string::String>`
   |
   = note: this error originates in the derive macro `Clone` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `AtomicBool: Clone` is not satisfied
  --> server/src/daemon/collection.rs:30:5
   |
24 | #[derive(Clone, Debug)]
   |          ----- in this derive macro expansion
...
30 |     locked: AtomicBool,
   |     ^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `AtomicBool`
```

server/src/daemon/collection.rs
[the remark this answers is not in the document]
bilelmoussaoui: https://specifications.freedesktop.org/secret-service/latest/re01.html.

server/src/daemon/service.rs
warusadura: Understood, thanks!

warusadura: > The todo! is not a fix.

Please see https://github.com/bilelmoussaoui/oo7/pull/73#discussion_r1521541858 :)

server/src/daemon/service.rs
bilelmoussaoui: Wrap those in Arcs. Nothing new here.

warusadura: Ha! :) will do, thanks!

warusadura: Done.

server/src/daemon/item.rs
```diff
@@ -0,0 +1,151 @@
...
+        object_server.remove::<Item, _>(self.path()).await?;
+        Collection::item_deleted(&ctxt, self.path()).await?;
+        Ok(Prompt::default())
+    }
+
+    #[zbus(name = "GetSecret")]
+    pub async fn secret(&self, session: ObjectPath<'_>) -> Result<SecretInner> {
+        let inner = self.inner.read().await;
+        let secret = inner.secret();
+        let session = self.manager.lock().unwrap().session(session).unwrap();
```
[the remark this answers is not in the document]
warusadura: Fixed with https://github.com/bilelmoussaoui/oo7/pull/73/commits/6fdfac2fa1027cf0d1ba14f9aee7ee53cf4e8faa, thanks!

server/src/daemon/item.rs
```diff
@@ -54,16 +54,20 @@ impl Item {
...
+        match self.manager.lock().unwrap().session(session.clone()) {
+            Some(session) => Ok(SecretInner(
+                session.path().into(),
+                parameters.to_vec(),
+                secret.to_vec(),
+                content_type.to_owned(),
+            )),
+            None => {
+                tracing::error!("Session {session} not found");
+                Err(ServiceError::NoSuchObject)
```
bilelmoussaoui, line 68: This should be NoSession instead. The NoSuchObject is only used for items/collections. See the docs.

warusadura: Got it, thanks!

server/src/daemon/service.rs
warusadura: Implemented via https://github.com/bilelmoussaoui/oo7/pull/73/commits/fe27cd8381955f67709f8bdeb89d16ada70f0f83

server/src/daemon/service.rs
```diff
@@ -286,7 +288,8 @@ impl Service {
     pub async fn new() -> Self {
         Self {
             collections: RwLock::new(Vec::new()),
-            keyring: Arc::new(Keyring::load_default().await.unwrap()),
+            keyring: Arc::new(Keyring::load(DEFAULT_KEYRING_PATH, todo!()).await.unwrap()),
```
warusadura, line 291: > Even load_default call is actually wrong. You need to call .load instead as you don't use the portal's key to encrypt the keyring.

Yeah, I agree that we should call `load` here. And `load` returns a new keyring if the one we request is not found: https://github.com/bilelmoussaoui/oo7/blob/5b73cfb7b90a93501b140f745586eb11b1c5332e/client/src/portal/mod.rs#L110

But, I'm really not sure how to create and pass the `Secret` argument to `load`. I think this is one of the places that we need the Prompt.

server/src/daemon/service.rs
warusadura: I have started a new discussion [here](https://github.com/bilelmoussaoui/oo7/pull/73#discussion_r1523242017)

client/src/crypto/mod.rs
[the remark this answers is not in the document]
warusadura: Sorry. This could be trivial. But, I don't get it :)

I tried the following and the compiler is not happy,
```
+#[cfg(feature = "unstable")]
+#[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
+pub use self::openssl::*;
```

```
error[E0432]: unresolved import `self::openssl`
  --> client/src/crypto/mod.rs:15:15
   |
15 | pub use self::openssl::*;
   |               ^^^^^^^ could not find `openssl` in `self`

```

server/src/daemon/secret.rs
```diff
@@ -0,0 +1,41 @@
+use oo7::{crypto, Key};
+
+use super::session::Session;
+
+pub struct Secret {
+    session: Session,
+    parameters: Vec<u8>,
+    value: Vec<u8>,
+    content_type: String,
```
warusadura, line 9: Except the `value` field others are causing the clippy warnings. What's the best approach to fix them?

```
warning: fields `session`, `parameters`, and `content_type` are never read
 --> server/src/daemon/secret.rs:6:5
  |
5 | pub struct Secret {
  |            ------ fields in this struct
6 |     session: Session,
  |     ^^^^^^^
7 |     parameters: Vec<u8>,
  |     ^^^^^^^^^^
8 |     value: Vec<u8>,
9 |     content_type: String,
  |     ^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default
```

server/src/daemon/service.rs
bilelmoussaoui: Isn't the key something you get from tpm?

client/src/dbus/algorithm.rs
```diff
@@ -20,11 +20,24 @@ impl Serialize for Algorithm {
         S: serde::Serializer,
```
bilelmoussaoui, line 20: The changes on this file can be pulled out of this PR and be merged already

client/src/crypto/native.rs
```diff
@@ -19,7 +19,7 @@ type EncAlg = cbc::Encryptor<aes::Aes128>;
 type DecAlg = cbc::Decryptor<aes::Aes128>;
 type MacAlg = hmac::Hmac<sha2::Sha256>;
 
-pub(crate) fn encrypt(data: impl AsRef<[u8]>, key: &Key, iv: impl AsRef<[u8]>) -> Vec<u8> {
+pub fn encrypt(data: impl AsRef<[u8]>, key: &Key, iv: impl AsRef<[u8]>) -> Vec<u8> {
```
bilelmoussaoui, line 22: The changes on this file can be pulled out of this PR and be merged already

client/src/dbus/api/mod.rs
```diff
@@ -27,5 +27,8 @@ pub(crate) use properties::Properties;
 #[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
 pub use properties::Properties;
 pub use secret::Secret;
+#[cfg(feature = "unstable")]
+#[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
+pub use secret::SecretInner;
```
bilelmoussaoui, line 32: Same here

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,206 @@
...
+};
+use tokio::sync::RwLock;
+use zbus::{interface, zvariant, ObjectServer, SignalContext};
+use zvariant::{ObjectPath, OwnedObjectPath};
+
+use super::{
+    error::ServiceError, prompt::Prompt, secret::Secret, service_manager::ServiceManager, Result,
+    Service,
+};
+use crate::SECRET_COLLECTION_PREFIX;
```
bilelmoussaoui, line 21: this is not a prefix, please name the variables according to their actual content

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,206 @@
...
+            item,
+            parameters,
+            content_type,
+            *self.item_counter.read().await,
+            self.path(),
+            Arc::clone(&self.keyring),
+            Arc::clone(&self.manager),
+        )
+        .await;
+        let path = OwnedObjectPath::from(item.path());
```
bilelmoussaoui, line 110: Why create an Owned variant if you need a ref anyways ?

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,206 @@
...
+        Service::collection_changed(ctxt, self.path.as_ref())
+            .await
+            .unwrap();
+    }
+
+    pub async fn item_counter(&self) -> i32 {
+        *self.item_counter.read().await
+    }
+
+    pub async fn set_item_counter(&self) {
```
bilelmoussaoui, line 203: the method doesn't set the item counter, it increments it

server/src/daemon/item.rs
```diff
@@ -0,0 +1,161 @@
...
+        Ok(Prompt::default())
+    }
+
+    #[zbus(name = "GetSecret")]
+    pub async fn secret(&self, session: ObjectPath<'_>) -> Result<SecretInner> {
+        let inner = self.inner.read().await;
+        let secret = inner.secret();
+        let parameters = self.parameters();
+        let content_type = self.content_type();
+        match self.manager.lock().unwrap().session(session.clone()) {
```
bilelmoussaoui, line 59: instead of clonning, make the session method take a ref, you only need it for comparison no?

server/src/daemon/service.rs
```diff
@@ -0,0 +1,368 @@
...
+                    item.clone(),
+                    item.secret().to_vec(),
+                    "text/plain".to_string(),
+                    login.item_counter().await,
+                    login.path(),
+                    service.keyring.clone(),
+                    service.manager.clone(),
+                )
+                .await;
+                login.set_item_counter().await;
```
bilelmoussaoui, line 361: I don't think you should export all the items until the collection is unlocked, no? also exposing the item counter & updating the counter here sounds meh

server/src/daemon/service.rs
```diff
@@ -0,0 +1,368 @@
...
+                let path = OwnedObjectPath::from(item.path());
+                login.items.write().await.push(item.clone());
```
bilelmoussaoui, line 363: why are you writing to items directly here?

server/src/daemon/service.rs
```diff
@@ -0,0 +1,368 @@
...
+                object_server.at(&path, item).await.unwrap();
```
bilelmoussaoui, line 364: the item should have a method to export itself on an ObjectServer, this way you avoid exposing the path itself

server/src/daemon/service_manager.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use super::session::Session;
+
+#[derive(Debug, Default)]
+pub struct ServiceManager {
+    sessions: HashMap<OwnedObjectPath, Session>,
+}
+
+impl ServiceManager {
+    pub fn session(&self, path: ObjectPath<'_>) -> Option<Session> {
```
bilelmoussaoui, line 13: yes, as i said above, take a ref here and if you need an OwnedObjectPath, take one. Or take a generic param & avoid calling into() in the codebase.

server/src/daemon/session.rs
```diff
@@ -0,0 +1,55 @@
...
+        manager: Arc<Mutex<ServiceManager>>,
+        sessions_counter: i32,
+    ) -> Self {
+        Self {
+            aes_key: Arc::new(aes_key),
+            path: OwnedObjectPath::try_from(format!(
+                "{}s{}",
+                SECRET_SESSION_PREFIX, sessions_counter
+            ))
+            .unwrap(),
```
bilelmoussaoui, line 43: you don't use the counter for anything other than constructing the path, just make the Session type take the path itself and construct it where you store the counter.

server/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[tokio::main]
+async fn main() -> daemon::Result<()> {
+    let args = Args::parse();
+    let mut password = String::new();
+    tracing_subscriber::fmt::init();
+
+    if args.login {
+        password = rpassword::prompt_password("Enter the login password: ").unwrap();
+    }
```
bilelmoussaoui, line 35: So otherwise you are passing an empty string, make the Service::new take a Option<String> maybe? Also this sounds very much like it shouldn't be a string. At worst, it should be a Zerioized<String>

client/src/dbus/algorithm.rs
warusadura: Done, thanks!

client/src/crypto/native.rs
warusadura: Done, thanks!

server/src/daemon/collection.rs
warusadura: I was following the gnome-keyring naming style: https://gitlab.gnome.org/GNOME/gnome-keyring/-/blob/master/daemon/dbus/gkd-secret-types.h#L38. This constant defines the "pre" part of an ObjectPath.

server/src/daemon/collection.rs
warusadura: I agree. Will change it to `increment_item_counter`

client/src/crypto/mod.rs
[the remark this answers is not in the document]
bilelmoussaoui: Handled that for you in https://github.com/bilelmoussaoui/oo7/pull/92

server/src/main.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    tracing::info!("Starting {}", BINARY_NAME);
+
+    let service = daemon::Service::new(Some(password)).await;
```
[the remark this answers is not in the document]
warusadura: Thank you!

warusadura:
Note: dropped following commits due to [this](https://github.com/bilelmoussaoui/oo7/pull/92) change and rebased.
 - client: make key generation methods public and encrypt(): 8dcc58f5af2de7b48fe6005e96c7bcd4f3bb6ad6
 - client: make crypto crate public under unstable feature: 145f9644db950635be7a125be47be48b629d0db4
 - client: Expose native_crypto API generate_iv() and encrypt(): e694e2e05c8d0696288ad66826038498ef2746ce

client/src/crypto/mod.rs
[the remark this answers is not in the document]
warusadura: Thank you :)

server/src/daemon/secret_exchange.rs
```diff
@@ -20,55 +22,58 @@ pub struct SecretExchange {
...
 
+    // Send Secret and perform encryption
     pub fn secret_exchange_send(
         &self,
         secret: &str,
-        exchange: HashMap<&str, String>,
-    ) -> Result<HashMap<&str, String>, base64::DecodeError> {
-        let public_key = BASE64_STANDARD.decode(exchange.get(PUBLIC).unwrap())?;
-        let public_key = Key::new(public_key);
+        exchange: String,
```
ueno, line 61: Where this `exchange` used?

server/src/daemon/secret_exchange.rs
```diff
@@ -20,55 +22,58 @@ pub struct SecretExchange {
...
         let aes_key = Key::generate_aes_key(&self.private_key, &public_key);
         let iv = crypto::generate_iv();
         let secret = crypto::encrypt(secret, &aes_key, &iv);
 
-        let mut exchange: HashMap<&str, String> = HashMap::new();
-        exchange.insert(SECRET, BASE64_STANDARD.encode(secret));
-        exchange.insert(PUBLIC, BASE64_STANDARD.encode(&self.public_key));
-        exchange.insert(IV, BASE64_STANDARD.encode(iv));
+        let map = HashMap::from([
+            (PUBLIC, BASE64_STANDARD.encode(&self.public_key)),
```
ueno, line 71: Given we only use this KeyFile format with binary blob as values, it might be simpler to move the base64 encoding/decoding logic to the `encode`/`decode` functions, like `fn encode(map: &HashMap<&str, &[u8]>) -> Result<String, base64::DecodeError>`?

server/src/daemon/secret_exchange.rs
```diff
@@ -20,55 +22,58 @@ pub struct SecretExchange {
...
-        exchange_in: HashMap<&str, String>,
-    ) -> Result<HashMap<&str, String>, base64::DecodeError> {
+    // Decrypt and retrieve secret
+    pub fn secret_exchange_receive(&self, exchange: String) -> Result<String, base64::DecodeError> {
+        let map = decode(&exchange);
         let mut exchange_out: HashMap<&str, String> = HashMap::new();
-        let secret = exchange_in.get(SECRET);
+
+        let secret = map.get(SECRET);
         if !secret.is_none() {
```
ueno, line 38: `secret.is_some()` ?

server/src/daemon/secret_exchange.rs
```diff
@@ -20,55 +22,58 @@ pub struct SecretExchange {
...
-        exchange.insert(PUBLIC, BASE64_STANDARD.encode(&self.public_key));
-        exchange.insert(IV, BASE64_STANDARD.encode(iv));
+        let map = HashMap::from([
+            (PUBLIC, BASE64_STANDARD.encode(&self.public_key)),
+            (SECRET, BASE64_STANDARD.encode(secret)),
+            (IV, BASE64_STANDARD.encode(iv)),
+        ]);
 
-        Ok(exchange)
+        Ok(encode(&map))
```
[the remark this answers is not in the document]
warusadura: yes, I have included the protocol prefix just now.

server/src/daemon/secret_exchange.rs
```diff
@@ -82,8 +87,32 @@ impl SecretExchange {
     }
 }
 
-pub fn get_secret(exchange: HashMap<&str, String>) -> String {
-    exchange.get(SECRET).unwrap().to_owned()
+// Parse payload HashMap into a String
```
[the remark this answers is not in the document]
warusadura: yeah, I agree, done.

server/src/daemon/secret_exchange.rs
warusadura: done, thanks!

server/src/daemon/secret_exchange.rs
warusadura: make sense, thanks! Moved base64 encoding/decoding out of SecretExchange API and into the encode/decode functions.

server/src/daemon/secret_exchange.rs
warusadura: this is the payload right?

server/src/daemon/secret_exchange.rs
```diff
@@ -20,55 +23,52 @@ pub struct SecretExchange {
...
-    pub fn secret_exchange_send(
-        &self,
-        secret: &str,
-        exchange: HashMap<&str, String>,
-    ) -> Result<HashMap<&str, String>, base64::DecodeError> {
-        let public_key = BASE64_STANDARD.decode(exchange.get(PUBLIC).unwrap())?;
-        let public_key = Key::new(public_key);
+    // Send Secret and perform encryption
+    pub fn secret_exchange_send(&self, secret: &str, exchange: String) -> String {
+        let decoded = decode(&exchange).unwrap();
```
ueno, line 57: It's better not to take an owned value: `exchange: &str`?

warusadura: done, thanks :)

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,217 @@
...
+
+#[interface(name = "org.freedesktop.Secret.Collection")]
+impl Collection {
+    pub async fn delete(
+        &self,
+        #[zbus(signal_context)] ctxt: SignalContext<'_>,
+        #[zbus(object_server)] object_server: &ObjectServer,
+    ) -> Result<Prompt> {
+        let _ = object_server.remove::<Self, _>(&self.path).await;
+        Service::collection_deleted(&ctxt, self.path.as_ref()).await?;
```
bilelmoussaoui, line 45: Log the deleted collection path

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,217 @@
...
+            .await
+            .map_err::<ServiceError, _>(From::from)?;
+        *self.item_counter.write().await += 1;
+
+        let item = super::item::Item::new(
+            item,
+            parameters,
+            content_type,
+            *self.item_counter.read().await,
+            self.path(),
```
bilelmoussaoui, line 103: I think you should generate the path here and only pass that info.

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,217 @@
...
+            Arc::clone(&self.keyring),
+            Arc::clone(&self.manager),
+        )
+        .await;
+        let path = OwnedObjectPath::from(item.path());
```
bilelmoussaoui, line 108: Log the path

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,217 @@
...
+        self.items.write().await.push(item.clone());
+        object_server.at(&path, item).await.unwrap();
+
+        // perform prompt
+        self.manager.lock().unwrap().update_prompts_counter();
+        let prompt = Prompt::new(
+            Arc::clone(&self.manager),
```
bilelmoussaoui, line 115: You already pass the manager, just do the prompts counting bump there?

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,217 @@
...
+    ) -> Self {
+        Self {
+            items: Default::default(),
+            label: label.to_owned(),
+            alias: Arc::new(RwLock::new(alias.to_owned())),
+            locked: Arc::new(AtomicBool::new(false)),
+            modified: created,
+            item_counter: Arc::new(RwLock::new(0)),
+            path: OwnedObjectPath::try_from(format!("{}{}", SECRET_COLLECTION_PREFIX, label))
+                .unwrap(),
```
bilelmoussaoui, line 185: Same, those two should be handled when creating the collection

server/src/daemon/collection.rs
```diff
@@ -0,0 +1,217 @@
...
+            .unwrap();
+    }
+
+    pub async fn item_counter(&self) -> i32 {
+        *self.item_counter.read().await
+    }
+
+    pub async fn incr_item_counter(&self) {
+        *self.item_counter.write().await += 1;
+    }
```
bilelmoussaoui, line 216:
```suggestion
```
Shouldn't be exposed.

server/src/daemon/error.rs
```diff
@@ -0,0 +1,21 @@
+/// DBus Secret Service specific errors.
+/// <https://specifications.freedesktop.org/secret-service/latest/ch15.html>
+#[derive(zbus::DBusError, Debug)]
+#[zbus(prefix = "org.freedesktop.Secret.Error")]
+pub enum ServiceError {
```
bilelmoussaoui, line 5: We have this error type defined in oo7 already, so one should make use of that and create a new error type that cover server side specificities

server/src/daemon/item.rs
```diff
@@ -0,0 +1,163 @@
...
+        #[zbus(signal_context)] ctxt: SignalContext<'_>,
+    ) -> Result<ObjectPath> {
+        let inner = self.inner.read().await;
+        let attributes = inner.attributes();
+        self.keyring
+            .delete(attributes)
+            .await
+            .map_err::<ServiceError, _>(From::from)?;
+        object_server.remove::<Item, _>(self.path()).await?;
+        Collection::item_deleted(&ctxt, self.path()).await?;
```
bilelmoussaoui, line 49: Log the path

server/src/daemon/item.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        // returning an empty objectpath: "/" is enough here
+        Ok(ObjectPath::default())
```
bilelmoussaoui, line 52: Is it? or does it depend on something else? needs a reason why it is enough anyways

server/src/daemon/prompt.rs
```diff
@@ -0,0 +1,139 @@
...
+        #[zbus(connection)] connection: &zbus::Connection,
+        #[zbus(header)] header: Header<'_>,
+    ) -> fdo::Result<()> {
+        println!("prompt: {}", header.path().unwrap());
+
+        let post_fix = if header.path().unwrap().as_str().to_string().contains("/u") {
+            Some("u")
+        } else {
+            None
+        };
```
bilelmoussaoui, line 54: This is horrible, just have two different kinds of Prompts.

server/src/daemon/prompt.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+impl Prompt {
+    pub fn new(
+        manager: Arc<Mutex<ServiceManager>>,
+        prompts_counter: i32,
+        post_fix: Option<&str>,
+    ) -> Self {
+        // if the Prompt::new() is coming from Unlock, we use a postfix 'u'
```
bilelmoussaoui, line 106: create two different constructors.

server/src/daemon/prompter.rs
```diff
@@ -0,0 +1,210 @@
...
+
+use super::{secret_exchange, service_manager::ServiceManager};
+
+// May be change this to /org/oo7_daemon/Prompt
+const SECRET_PROMPTER_PREFIX: &str = "/org/gnome/keyring/Prompt/";
+
+#[derive(Clone, Default, DeserializeDict, Debug, Type, SerializeDict)]
+#[zvariant(signature = "dict")]
+pub struct Properties {
+    title: Option<String>,
```
bilelmoussaoui, line 28: Where those properties are defined? what uses them at the end? are all of them still used?

server/src/daemon/prompter.rs
```diff
@@ -0,0 +1,210 @@
...
+    path: OwnedObjectPath,
+}
+
+#[interface(name = "org.gnome.keyring.internal.Prompter.Callback")]
+impl PrompterCallback {
+    pub async fn prompt_ready(
+        &self,
+        reply: &str,            // the purpose of this?
+        properties: Properties, // this should probably be &mut Properties
+        exchange: &str,         // this should probably be &mut String
```
bilelmoussaoui, line 62: That is not possible, you return values, not take &mut ....

server/src/daemon/prompter.rs
```diff
@@ -0,0 +1,210 @@
...
+        #[zbus(connection)] connection: &zbus::Connection,
+        #[zbus(header)] header: Header<'_>,
+    ) -> fdo::Result<()> {
+        println!("prompter: {}", header.path().unwrap());
+
+        let mut properties = properties.clone();
+
+        if header.path().unwrap().as_str().to_string().contains("/u") {
```
bilelmoussaoui, line 70: By having different constructors, you can easily find out this. Even create a different Properties type based on that.

I would suggest defining an enum like this
```rust
pub enum PromptSource {
  Unlock,
  NewCollection,
}
```
You can then add a `Properties::for_source` that generates this. Or a `PromptSource::path(counter)`

server/src/daemon/prompter.rs
```diff
@@ -0,0 +1,210 @@
...
+            // setting properties related to Secret.Service.Unlock
+            properties.continue_label = Some(String::from("Unlock"));
+            properties.warning = Some(String::new());
```
bilelmoussaoui, line 73: Empty strings, skip them.

server/src/daemon/prompter.rs
```diff
@@ -0,0 +1,210 @@
...
+            ));
+            properties.title = Some(String::new());
+            properties.message = Some(String::from("Choose password for new keyring"));
+            properties.choice_label = Some(String::new());
+            properties.caller_window = Some(String::new());
+            properties.cancel_label = Some(String::from("Cancel"));
+            properties.password_new = Some(true);
+        }
+
+        *self.properties.lock().unwrap() = properties.clone();
```
bilelmoussaoui, line 101: Why you need to store the properties here?

server/src/daemon/prompter.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        let path = Arc::new(header.path().unwrap().to_owned());
```
bilelmoussaoui, line 103: Why? just clone the path

server/src/daemon/prompter.rs
```diff
@@ -0,0 +1,210 @@
...
+        let exchange = exchange.to_string();
+        let connection = Arc::new(connection.to_owned());
+
+        tokio::spawn(async move {
+            let prompter = PrompterProxy::new(&connection).await.unwrap();
```
bilelmoussaoui, line 108: Don't unwrap please

server/src/daemon/prompter.rs
```diff
@@ -0,0 +1,210 @@
...
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> fdo::Result<()> {
+        object_server.remove::<Self, _>(&self.path).await?;
+        Ok(())
+    }
+}
+
+impl PrompterCallback {
+    pub fn new(post_fix: Option<&str>, prompts_counter: i32) -> Self {
+        let path = if post_fix.is_some() {
```
bilelmoussaoui, line 129: Same thing, two constructors or take the `PromptSource` argument.

server/src/daemon/secret.rs
```diff
@@ -0,0 +1,42 @@
+use oo7::{crypto, Key};
+
+use super::session::Session;
+
+#[allow(dead_code)]
+pub struct Secret {
```
bilelmoussaoui, line 6: Zeroize please

server/src/daemon/secret.rs
```diff
@@ -0,0 +1,42 @@
...
+            session,
+            parameters: iv,
+            value: secret,
+            content_type: content_type.to_owned(),
+        }
+    }
+
+    pub fn value(&self) -> Vec<u8> {
+        self.value.to_owned()
+    }
```
bilelmoussaoui, line 41:
```suggestion
    pub fn value(&self) -> &[u8] {
        &self.value
    }
```

server/src/daemon/service.rs
```diff
@@ -0,0 +1,391 @@
...
+        object_server
+            .at(prompt.path().to_owned(), prompt.to_owned())
+            .await?;
+
+        Ok((unlocked, prompt.path().to_owned()))
+    }
+
+    #[zbus(out_args("locked", "prompt"))]
+    pub async fn lock(
+        &mut self,
```
bilelmoussaoui, line 194: this is bad, don't use `&mut self` in any dbus function handler

server/src/daemon/secret_exchange.rs
```diff
@@ -0,0 +1,139 @@
...
+
+#[derive(Debug)]
+pub struct SecretExchange {
+    private_key: Key,
+    public_key: Key,
+}
+
+impl SecretExchange {
+    // Creates the initial payload containing caller public_key
+    pub fn secret_exchange_begin(&self) -> String {
```
bilelmoussaoui, line 27: Drop the prefix from the function names

server/src/daemon/secret_exchange.rs
```diff
@@ -0,0 +1,139 @@
...
+        .iter()
+        .map(|(k, v)| format!("{}={}", k, BASE64_STANDARD.encode(v)))
+        .collect::<Vec<_>>()
+        .join("\n");
+    exchange.insert_str(0, PROTOCOL); // to add PROTOCOL prefix
+
+    exchange
+}
+
+// Convert a payload String into a HashMap
```
bilelmoussaoui, line 97: Please plink to a specification or the original C implementation

server/src/daemon/secret_exchange.rs
```diff
@@ -0,0 +1,139 @@
...
+        let (key, value) = pair.split_once("=").unwrap();
+        encoded = BASE64_STANDARD.decode(value)?;
+        map.insert(key, encoded);
+    }
+
+    Ok(map)
+}
+
+// Retrieve secret from the payload
+pub fn get_secret(exchange: &str) -> Result<String, std::str::Utf8Error> {
```
bilelmoussaoui, line 114: Not sure this has to be public or this whole secret exchange thingy. Make sure only what is needed outside is what is public.

server/src/daemon/service.rs
```diff
@@ -0,0 +1,391 @@
...
+            *self.sessions_counter.read().await,
+        );
+        self.manager
+            .lock()
+            .unwrap()
+            .insert_session(session.path().to_owned(), session.to_owned());
+
+        object_server
+            .at(session.path().to_owned(), session.to_owned())
+            .await?;
```
bilelmoussaoui, line 75: If you create the path here, you would avoid a bit of dancing here with the path.

server/src/daemon/collection.rs
warusadura: Done, thanks!

server/src/daemon/collection.rs
warusadura: done.

server/src/daemon/prompter.rs
warusadura: it was there just temporarily, now removed.

server/src/daemon/secret_exchange.rs
warusadura: done, thanks!

server/src/daemon/secret_exchange.rs
warusadura: done.

server/src/daemon/service.rs
warusadura: understood, thanks!

server/src/daemon/secret.rs
warusadura: Done, thanks!

server/src/daemon/secret_exchange.rs
warusadura: yeah, I'll look into it :)

server/src/daemon/prompter.rs
warusadura: Done

server/src/daemon/item.rs
warusadura: Yes it is :) The reason is, gnome-keyring-daemon returns an empty objectpath: '/' here

server/src/daemon/collection.rs
warusadura: Moved incrementing the counter to Prompt constructors.

server/src/daemon/prompt.rs
warusadura: yeah :) implemented two Prompt constructors.

server/src/daemon/prompter.rs
warusadura: Done, thanks!

server/src/daemon/prompt.rs
warusadura: Done, thanks!

server/src/main.rs

server/src/daemon/collection.rs
warusadura: Done, thanks!

server/src/daemon/prompter.rs
warusadura: Done, thanks!

server/src/daemon/service.rs
warusadura: yeah, I agree. Removed.

server/src/daemon/service.rs
warusadura: thanks, I just removed this part.

server/src/daemon/prompter.rs
warusadura: This is now removed :)

bilelmoussaoui:
I think most of the stuff in this PR is merged into main already, if anything left it can be done separately

PR #75, opened 2024-02-29
Write to stdout

A6GibKm:


A6GibKm:
In theory one should be able to do the same with `println` instead of directly writing to stdout, but no idea how to use it to print the bytes as is

cli/src/main.rs
```diff
@@ -147,12 +155,23 @@ async fn store(label: &str, attributes: &impl AsAttributes) -> Result<(), Error>
     Ok(())
 }
 
-async fn lookup(attributes: &impl AsAttributes) -> Result<(), Error> {
+async fn lookup(attributes: &impl AsAttributes, as_hex: bool) -> Result<(), Error> {
     let collection = collection().await?;
     let items = collection.search_items(attributes).await?;
 
     if let Some(item) = items.first() {
-        print_item(item).await?;
```
bilelmoussaoui, line 155: I think we should still use print_item and make it take a as_hex bool.

bilelmoussaoui: Also not a fan of adding an option to control the output, we should just use hex representation in case the secret is a blob instead of text

A6GibKm: But hex makes it useless for the purposes of piping into other utils

A6GibKm: I do use it for this purpose. Most binaries that requiere some sort of identification with a config file allow setting passwords either in plaintext format or by providing a command to execute.  

E.g. mbsync

bilelmoussaoui:
I added two separate options, one that prints secrets as hex. This option would be also useful for a future `json` that allows you to parse the output programmatically pretty easily afterwards and then the `--secret-only` option that would help your use case, it is also passed to the `search` command as it can also display only one item if the `all` argument is not passed.

PR #76, opened 2024-03-20
portal: Support migration from legacy keyring format

ueno:
This adds support for transparent data migration from legacy keyring format (version 0) used by gnome-keyring.  If the legacy version is found, it tries to decrypt items in memory, encrypt and store them in a new keyring in the version 1 format.


client/src/portal/api/mod.rs
```diff
@@ -342,4 +562,20 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn legacy_decrypt() -> Result<(), Error> {
+        let path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("fixtures")
+            .join("legacy.keyring");
+        let blob = tokio::fs::read(&path).await?;
+        let keyring = LegacyKeyring::try_from(blob.as_slice())?;
+        let password = b"test";
+        let secret: Secret = Vec::from(password).into();
```
bilelmoussaoui, line 574:
```suggestion
        let secret = Secret::from(password);
```
instead?

client/src/portal/api/mod.rs
```diff
@@ -342,4 +562,20 @@ mod tests {
...
+        let items = keyring.decrypt_items(&secret)?;
+
+        assert_eq!(items.len(), 1);
```
bilelmoussaoui, line 577: Can we have the tests also check that we are reading the item properly?

client/src/portal/mod.rs
```diff
@@ -399,4 +426,19 @@ mod tests {
...
+    async fn migrate_from_legacy() -> Result<(), Error> {
+        let path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("fixtures")
+            .join("legacy.keyring");
+
+        let password = b"test";
+        let secret: Secret = Vec::from(password).into();
+        let keyring = Keyring::load(&path, secret).await?;
+
+        assert_eq!(keyring.n_items().await, 1);
```
bilelmoussaoui, line 440: Same as above

client/src/portal/api/mod.rs
```diff
@@ -279,11 +286,224 @@ impl TryFrom<&[u8]> for Keyring {
     }
 }
 
-#[cfg(test)]
+#[derive(Debug)]
```
bilelmoussaoui, line 289: Can you instead move this to a new separate module? to keep the legacy bits separate from the reset.

Cargo.toml
```diff
@@ -20,6 +20,7 @@ rust-version = "1.75"
 exclude = ["org.freedesktop.Secrets.xml"]
 
 [workspace.dependencies]
+byteorder = "1.5.0"
```
bilelmoussaoui, line 23: zbus itself uses https://crates.io/crates/endi, it would be nice if we can use that to avoid pulling the same "code" twice.

bilelmoussaoui (commented):
Passing by initial review

client/src/portal/api/mod.rs
ueno: Sure.

client/src/portal/api/mod.rs
ueno: Yes, expanded the test.

client/src/portal/api/mod.rs
ueno: Moved to `legacy_keyring.rs`.

Cargo.toml
ueno: Switched to using endi instead.

ueno:
Thank you for the prompt review!

A6GibKm:
For the sake of documenting the behavior, gnome-keyring won't be able to read the keyring back right?

ueno:
Thinking a bit more about this, I've come up with the following proposal (from the oo7-daemon perspective):

- Let's start with a new directory, say `~/.local/share/oo7/keyrings/` instead of `~/.local/share/keyrings/`
- Make `portal::Keyring::load` (possibly need a new function) take a keyring name, instead of a path
- The function first looks at `~/.local/share/oo7/keyrings/<name>.keyring`. If it's found, use it anyway
- Otherwise look at `~/.local/share/keyrings/<name>.keyring`. If it's found, decrypt it and migrate it to the new location (`~/.local/share/oo7/keyrings/<name>.keyring`) when writing
- Otherwise, just create a new keyring at `~/.local/share/oo7/keyrings/<name>.keyring`

How does it sound @A6GibKm @bilelmoussaoui?

bilelmoussaoui:
> * Let's start with a new directory, say `~/.local/share/oo7/keyrings/` instead of `~/.local/share/keyrings/`
> 
> * Make `portal::Keyring::load` (possibly need a new function) take a keyring name, instead of a path
> 
> * The function first looks at `~/.local/share/oo7/keyrings/<name>.keyring`. If it's found, use it anyway
> 
> * Otherwise look at `~/.local/share/keyrings/<name>.keyring` (which can be either version 0 or 1). If it's found, migrate it to the new location (`~/.local/share/oo7/keyrings/<name>.keyring`) when writing
> 
> * Otherwise, just create a new keyring at `~/.local/share/oo7/keyrings/<name>.keyring`

That sounds sane to me, not sure about using oo7 in the path but we can figure out that later.

client/src/portal/api/attribute_value.rs
```diff
@@ -5,7 +5,7 @@ use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
 use crate::{crypto, Key};
 
 /// An encrypted attribute value.
-#[derive(Deserialize, Serialize, Type, Clone, Debug, Zeroize, ZeroizeOnDrop)]
+#[derive(Deserialize, Serialize, Type, Clone, Debug, PartialEq, Zeroize, ZeroizeOnDrop)]
```
bilelmoussaoui, line 8: Clippy would complain if Eq is not implemented as well

client/src/portal/api/legacy_keyring.rs
```diff
@@ -0,0 +1,272 @@
...
+        let pos = cursor.position() as usize;
+        if size > cursor.get_ref()[pos..].len() {
+            return Err(Error::NoData);
+        }
+        if size % 16 != 0 {
+            size = (size / 16) * 16;
+        }
+        let encrypted_content = Vec::from(&cursor.get_ref()[pos..pos + size]);
+
+        Ok(Keyring {
```
bilelmoussaoui, line 209:
```suggestion
        Ok(Self {
```

client/src/portal/api/legacy_keyring.rs
```diff
@@ -0,0 +1,272 @@
...
+            ));
+        }
+
+        let version = value.get(FILE_HEADER_LEN..(FILE_HEADER_LEN + 2));
+        if version != Some(&[MAJOR_VERSION, MINOR_VERSION]) {
+            return Err(Error::VersionMismatch(version.map(|x| x.to_vec())));
+        }
+
+        if let Some(data) = value.get((FILE_HEADER_LEN + 2)..) {
+            Keyring::parse(data)
```
bilelmoussaoui, line 235:
```suggestion
            Self::parse(data)
```

client/src/portal/mod.rs
```diff
@@ -399,4 +426,31 @@ mod tests {
...
+            .join("fixtures")
+            .join("legacy.keyring");
+
+        let password = b"test";
+        let secret = Secret::from(password.to_vec());
+        let keyring = Keyring::load(&path, secret).await?;
+
+        assert_eq!(keyring.n_items().await, 1);
+        let items: Result<Vec<_>, _> = keyring.items().await.into_iter().collect();
+        assert!(items.is_ok());
```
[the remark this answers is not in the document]
ueno: I don't think `keyring.items().await?` works here, as `keyring.items().await` returns `Vec<Result<Item, _>>` not `Result<Vec<Item>, _>` (the above `collect` is to flatten it).

client/src/portal/mod.rs
```diff
@@ -399,4 +426,31 @@ mod tests {
...
+        let items = items.unwrap();
+        assert_eq!(items.len(), 1);
+        assert_eq!(items[0].label(), "foo");
+        assert_eq!(items[0].secret().as_ref(), b"foo".to_vec());
+        let attributes = items[0].attributes();
+        assert_eq!(attributes.len(), 1);
+        assert_eq!(
+            attributes.get("xdg:schema"),
+            Some(&AttributeValue::from("org.gnome.keyring.Note"))
```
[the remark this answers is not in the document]
ueno: It doesn't, but using `.map(|v| v.as_ref())` instead of `.as_deref()` works; thanks for the suggestion.

client/src/portal/api/mod.rs
```diff
@@ -224,16 +227,23 @@ impl Keyring {
         Ok(blob)
     }
 
-    pub fn default_path() -> Result<PathBuf, Error> {
+    pub fn path(name: &str, version: u8) -> Result<PathBuf, Error> {
```
bilelmoussaoui, line 230: Does this has to be public? I don't think so, so maybe just `pub(crate)`

client/src/portal/api/legacy_keyring.rs
```diff
@@ -0,0 +1,272 @@
...
+        let items = keyring.decrypt_items(&secret)?;
+
+        assert_eq!(items.len(), 1);
+        assert_eq!(items[0].label(), "foo");
+        assert_eq!(items[0].secret().as_ref(), b"foo".to_vec());
+        let attributes = items[0].attributes();
+        assert_eq!(attributes.len(), 1);
+        assert_eq!(
+            attributes.get("xdg:schema"),
+            Some(&AttributeValue::from("org.gnome.keyring.Note"))
```
bilelmoussaoui, line 267: Same as i mentioned before

client/src/portal/api/legacy_keyring.rs
```diff
@@ -0,0 +1,272 @@
...
+        let (key, iv) = crypto::legacy_derive_key_and_iv(
+            &**secret,
+            self.key_strength(secret),
+            &self.salt,
+            self.iteration_count.try_into().unwrap(),
+        );
+        let decrypted = crypto::decrypt_no_padding(&self.encrypted_content, &key, iv);
+        let (digest, content) = decrypted.split_at(16);
+        if !crypto::verify_checksum_md5(digest, content) {
+            // This not a MAC but anyway
```
bilelmoussaoui, line 46: Would be nice to have a specific error for this instead of using the wrong one, would make it easier to track down any potential issues in the future

bilelmoussaoui (commented):
otherwise this looks pretty decent to me

client/src/portal/mod.rs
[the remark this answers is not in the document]
bilelmoussaoui: I was sure there is a way to go from Vec<Result<T>> to Result<Vec<T>> without iterating/collecting manually

client/src/portal/api/legacy_keyring.rs
bilelmoussaoui: the other option would be to rename the used error to something more generic and maybe make it take a `&'static str` for the algorithm

ueno: Added `ChecksumMismatch` error.

bilelmoussaoui:
> Thinking a bit more about this, I've come up with the following proposal (from the oo7-daemon perspective):
> 
>     * Let's start with a new directory, say `~/.local/share/oo7/keyrings/` instead of `~/.local/share/keyrings/`
> 
>     * Make `portal::Keyring::load` (possibly need a new function) take a keyring name, instead of a path
> 
>     * The function first looks at `~/.local/share/oo7/keyrings/<name>.keyring`. If it's found, use it anyway
> 
>     * Otherwise look at `~/.local/share/keyrings/<name>.keyring` (which can be either version 0 or 1). If it's found, migrate it to the new location (`~/.local/share/oo7/keyrings/<name>.keyring`) when writing
> 
>     * Otherwise, just create a new keyring at `~/.local/share/oo7/keyrings/<name>.keyring`
> 
> 
> How does it sound @A6GibKm @bilelmoussaoui?

Is this something you think we should do as part of this PR? I think it is something that can be done at a later stage in a different PR :)

client/src/portal/api/legacy_keyring.rs
```diff
@@ -0,0 +1,271 @@
...
+
+#[cfg(feature = "async-std")]
+use std::io;
+use std::{
+    collections::HashMap,
+    io::{Cursor, Read},
+};
+
+use endi::{Endian, ReadBytes};
+#[cfg(feature = "tokio")]
```
bilelmoussaoui, line 11: This seems to be used only for errors, so let us just use std::io for that

ueno: Right, switched to using std::io.

ueno:
> Is this something you think we should do as part of this PR? I think it is something that can be done at a later stage in a different PR :)

I think it's feature complete now. I can follow up in a separate PR if any issue arises. Thank you for the review!

bilelmoussaoui (approved):

Issue #47, opened 2023-11-14
Rekeying support for oo7::portal::Keyring


ueno:
The [oo7](https://docs.rs/oo7/latest/oo7/index.html)::[portal](https://docs.rs/oo7/latest/oo7/portal/index.html)::[Keyring](https://docs.rs/oo7/latest/oo7/portal/struct.Keyring.html#) API currently doesn't provide a way to change the key. It would be nice if there is an API, something like:
```rust
pub async fn change_secret(&mut self, secret: Secret) -> Result<(), Error>;
...
let mut keyring = oo7::portal::Keyring::load_path(..., secret).await?;
keyring.change_secret(new_secret).await?;
keyring.write().await?;
```
This would be particularly useful for implementing a gnome-keyring compatible server backed by this file format (#46) and supporting password change through PAM.

sophie-h:
> The [oo7](https://docs.rs/oo7/latest/oo7/index.html)::[portal](https://docs.rs/oo7/latest/oo7/portal/index.html)::[Keyring](https://docs.rs/oo7/latest/oo7/portal/struct.Keyring.html#) API currently doesn't provide a way to change the key

You can use [`replace_item_index()`](https://docs.rs/oo7/latest/oo7/portal/struct.Keyring.html#method.replace_item_index) for that already.

> It would be nice if there is an API, something like

This would need some rework of how the library works right now. Currently the `portal::Item` does not know anything about the keyring file.

ueno:
I think there is a confusion; this issue is about master password to encrypt the items, not the items themselves.

sophie-h:
Oh sorry, I was totally off.

warusadura:
@bilelmoussaoui I'm interested in working on this issue as well :)

PR #77, opened 2024-03-24
portal: Add rekeying support for oo7::portal::Keyring

warusadura:
This change adds `change_secret()` to `oo7::portal::Keyring` API.
`change_secret()` change/update the secret associated with a keyring.
Note: we need this change for the sever side implementation.

Fixes: #47

### DIFF
--- a/client/src/portal/api/mod.rs
+++ b/client/src/portal/api/mod.rs
@@ -252,6 +252,16 @@ impl Keyring {
             self.iteration_count.try_into().unwrap(),
         )
     }
+
+    // Reset Keyring content
+    pub(crate) fn reset(&mut self) {
+        let salt = rand::thread_rng().gen::<[u8; DEFAULT_SALT_SIZE]>().to_vec();
+        self.salt_size = salt.len() as u32;
+        self.salt = salt;
+        self.iteration_count = DEFAULT_ITERATION_COUNT;
+        self.usage_count = 0;
+        self.items = Vec::new();
+    }
 }
 
 impl TryFrom<&[u8]> for Keyring {
--- a/client/src/portal/mod.rs
+++ b/client/src/portal/mod.rs
@@ -33,7 +33,7 @@ use std::io;
 use std::{
     collections::HashMap,
     path::{Path, PathBuf},
-    sync::{Arc, OnceLock},
+    sync::Arc,
 };
 
 #[cfg(feature = "async-std")]
@@ -81,8 +81,8 @@ pub struct Keyring {
     /// Times are stored before reading the file to detect
     /// file changes before writing
     mtime: Mutex<Option<std::time::SystemTime>>,
-    key: OnceLock<Key>,
-    secret: Arc<Secret>,
+    key: Mutex<Option<Arc<Key>>>,
+    secret: Mutex<Arc<Secret>>,
 }
 
 impl Keyring {
@@ -129,7 +129,7 @@ impl Keyring {
             path: path.as_ref().to_path_buf(),
             mtime: Mutex::new(mtime),
             key: Default::default(),
-            secret: Arc::new(secret),
+            secret: Mutex::new(Arc::new(secret)),
         })
     }
 
@@ -147,7 +147,7 @@ impl Keyring {
                 path: path.as_ref().to_path_buf(),
                 mtime: Default::default(),
                 key: Default::default(),
-                secret: Arc::new(secret),
+                secret: Mutex::new(Arc::new(secret)),
             }),
             Err(Error::VersionMismatch(Some(version)))
                 if version[0] == api::LEGACY_MAJOR_VERSION =>
@@ -169,7 +169,7 @@ impl Keyring {
                     path: path.as_ref().to_path_buf(),
                     mtime: Default::default(),
                     key: Default::default(),
-                    secret: Arc::new(secret),
+                    secret: Mutex::new(Arc::new(secret)),
                 })
             }
             Err(err) => Err(err),
@@ -205,7 +205,7 @@ impl Keyring {
                 path: v1_path,
                 mtime: Default::default(),
                 key: Default::default(),
-                secret: Arc::new(secret),
+                secret: Mutex::new(Arc::new(secret)),
             })
         }
     }
@@ -229,7 +229,7 @@ impl Keyring {
             .items
             .iter()
             .map(|e| {
-                (*e).clone().decrypt(key).map_err(|err| {
+                (*e).clone().decrypt(&key).map_err(|err| {
                     InvalidItemError::new(
                         err,
                         e.hashed_attributes.keys().map(|x| x.to_string()).collect(),
@@ -243,14 +243,14 @@ impl Keyring {
     pub async fn search_items(&self, attributes: &impl AsAttributes) -> Result<Vec<Item>, Error> {
         let key = self.derive_key().await;
         let keyring = self.keyring.read().await;
-        keyring.search_items(attributes, key)
+        keyring.search_items(attributes, &key)
     }
 
     /// Find the first item matching the attributes.
     pub async fn lookup_item(&self, attributes: &impl AsAttributes) -> Result<Option<Item>, Error> {
         let key = self.derive_key().await;
         let keyring = self.keyring.read().await;
-        keyring.lookup_item(attributes, key)
+        keyring.lookup_item(attributes, &key)
     }
 
     /// Delete an item.
@@ -258,7 +258,7 @@ impl Keyring {
         {
             let key = self.derive_key().await;
             let mut keyring = self.keyring.write().await;
-            keyring.remove_items(attributes, key)?;
+            keyring.remove_items(attributes, &key)?;
         };
         self.write().await
     }
@@ -284,10 +284,10 @@ impl Keyring {
             let key = self.derive_key().await;
             let mut keyring = self.keyring.write().await;
             if replace {
-                keyring.remove_items(attributes, key)?;
+                keyring.remove_items(attributes, &key)?;
             }
             let item = Item::new(label, attributes, secret);
-            let encrypted_item = item.encrypt(key)?;
+            let encrypted_item = item.encrypt(&key)?;
             keyring.items.push(encrypted_item);
             item
         };
@@ -308,7 +308,7 @@ impl Keyring {
             let mut keyring = self.keyring.write().await;
 
             if let Some(item_store) = keyring.items.get_mut(index) {
-                *item_store = item.encrypt(key)?;
+                *item_store = item.encrypt(&key)?;
             } else {
                 return Err(Error::InvalidItemIndex(index));
             }
@@ -340,10 +340,10 @@ impl Keyring {
         let mut keyring = self.keyring.write().await;
         for (label, attributes, secret, replace) in items {
             if replace {
-                keyring.remove_items(&attributes, key)?;
+                keyring.remove_items(&attributes, &key)?;
             }
             let item = Item::new(label, &attributes, &*secret);
-            let encrypted_item = item.encrypt(key)?;
+            let encrypted_item = item.encrypt(&key)?;
             keyring.items.push(encrypted_item);
         }
 
@@ -374,12 +374,52 @@ impl Keyring {
     }
 
     /// Return key, derive and store it first if not initialized
-    async fn derive_key(&self) -> &Key {
-        if self.key.get().is_none() {
+    async fn derive_key(&self) -> Arc<Key> {
+        if self.key.lock().await.is_none() {
             let key = self.keyring.read().await.derive_key(&self.secret);
-            self.key.set(key).unwrap();
+            *self.key.lock().await = Some(Arc::new(key));
         }
-        self.key.get().unwrap()
+
+        Arc::clone(self.key.lock().await.as_ref().unwrap())
+    }
+
+    /// Change keyring secret
+    ///
+    /// # Arguments
+    ///
+    /// * `secret` - The new secret to store.
+    pub async fn change_secret(&self, secret: Secret) -> Result<(), Error> {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("Changing keyring secret and key");
+
+        let keyring = self.keyring.read().await;
+        let key = self.derive_key().await;
+
+        let mut items = Vec::with_capacity(keyring.items.len());
+        for item in &keyring.items {
+            items.push(item.clone().decrypt(&key)?);
+        }
+        drop(keyring);
+
+        let mut secret_lock = self.secret.lock().await;
+        *secret_lock = Arc::new(secret);
+        drop(secret_lock);
+
+        let mut key_lock = self.key.lock().await;
+        // Unset the old key
+        *key_lock = None;
+        drop(key_lock);
+        let key = self.derive_key().await;
+
+        let mut keyring = self.keyring.write().await;
+
+        for item in items {
+            let encrypted_item = item.encrypt(&key)?;
+            keyring.items.push(encrypted_item);
+        }
+        drop(keyring);
+
+        self.write().await
     }
 }
 
@@ -606,6 +646,32 @@ mod tests {
         keyring.write().await?;
 
         assert!(v1_dir.join("default.keyring").exists());
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn change_secret() -> Result<(), Error> {
+        let path = PathBuf::from("../../tests/test_rekeying.keyring");
+
+        let keyring = Keyring::load(&path, strong_key()).await?;
+        let attributes = HashMap::from([("attr", "value")]);
+        let item_before = keyring
+            .create_item("test", &attributes, "password", false)
+            .await?;
+
+        let new_secret = Secret::from(b"password".to_vec());
+        keyring.change_secret(new_secret).await?;
+
+        let new_secret = Secret::from(b"password".to_vec());
+        let keyring = Keyring::load(&path, new_secret).await?;
+        let item_now = keyring.lookup_item(&attributes).await?.unwrap();
+
+        assert_eq!(item_before.label(), item_now.label());
+        assert_eq!(item_before.secret(), item_now.secret());
+        assert_eq!(item_before.attributes(), item_now.attributes());
+
+        fs::remove_file(path).await?;
 
         Ok(())
     }
--- a/client/src/portal/mod.rs
+++ b/client/src/portal/mod.rs
@@ -375,12 +375,28 @@ impl Keyring {
 
     /// Return key, derive and store it first if not initialized
     async fn derive_key(&self) -> Arc<Key> {
-        if self.key.lock().await.is_none() {
-            let key = self.keyring.read().await.derive_key(&self.secret);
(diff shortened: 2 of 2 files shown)


client/src/dbus/collection.rs
```diff
@@ -133,7 +133,7 @@ impl<'a> Collection<'a> {
                         Arc::clone(&self.session),
                         self.algorithm,
                         item,
-                        self.aes_key.as_ref().map(Arc::clone),
+                        self.aes_key.clone(),
```
bilelmoussaoui, line 136: Please drop this commit, fixed more clippy warnings + those in main just now

warusadura: will do, thanks!

client/src/portal/mod.rs
```diff
@@ -304,6 +304,19 @@ impl Keyring {
         }
         self.key.get().unwrap()
     }
+
+    /// Change keyring secret
+    ///
+    /// # Arguments
+    ///
+    /// * `secret` - The new secret to store.
+    pub async fn change_secret(&mut self, secret: Secret) -> Result<(), Error> {
```
A6GibKm, line 313: I wonder if this should accept the more general `&[u8]` and use Secret as an implementation detail inside.

warusadura: Yeah it make sense. But, I kind of like to follow the `Keyring::load()` approach :)

client/src/key.rs
```diff
@@ -4,7 +4,7 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 use crate::{crypto, portal};
 
 /// A key.
-#[derive(Debug, Zeroize, ZeroizeOnDrop)]
+#[derive(Clone, Debug, Zeroize, ZeroizeOnDrop)]
```
bilelmoussaoui, line 7: The clone here is wrong

client/src/portal/mod.rs
```diff
@@ -399,4 +415,22 @@ mod tests {
...
+    #[tokio::test]
+    async fn change_secret() -> Result<(), Error> {
+        let path = PathBuf::from("../../tests/change_secret.keyring");
+
+        let old_secret = Secret::from(vec![1, 2]);
+        let new_secret = Secret::from(vec![3, 4]);
+        let mut keyring = Keyring::load(&path, old_secret).await?;
+
+        keyring.change_secret(new_secret).await?;
+        keyring.write().await?;
```
ueno, line 428: I might be missing something, but don't this logic (`load` + `write`), also enumerating all items and re-encrypting them, need to be moved to the `change_secret` method; otherwise you will no longer be able to decrypt the items?

client/src/key.rs
warusadura: Removed, thanks!

client/src/portal/mod.rs
bilelmoussaoui: Agree, just writing again is not enough as a test case anywsays as it would never fail. Instead, I would like to see the test checking that the items before and after changing the secret are still readable.

haecker-felix:
Since this also changes the `derive_key()` function, a note that it is currently blocking despite being async (which was fine in 0.2). Might be related to https://github.com/bilelmoussaoui/oo7/commit/86c2a8f41bece33882358c795f0f06111329be95.

client/src/portal/mod.rs
warusadura: Done, thanks!

client/src/portal/mod.rs
```diff
@@ -375,8 +375,21 @@ impl Keyring {
 
     /// Return key, derive and store it first if not initialized
     async fn derive_key(&self) -> Arc<Key> {
+        let keyring = Arc::clone(&self.keyring);
+        let secret = Arc::clone(&self.secret);
+
         if self.key.lock().await.is_none() {
```
bilelmoussaoui, line 381: the guard is not dropped here, are you sure this solves the blocking issue? you should probably test key rack with this branch of oo7

client/src/portal/mod.rs
```diff
@@ -375,10 +375,25 @@ impl Keyring {
...
+            let key =
+                tokio::task::spawn_blocking(move || keyring.blocking_read().derive_key(&secret))
+                    .await
+                    .unwrap();
+
+            *key_lock = Some(Arc::new(key));
         }
+        drop(key_lock);
 
         Arc::clone(self.key.lock().await.as_ref().unwrap())
```
bilelmoussaoui, line 398: You are using the lock again here...

warusadura: done, thanks :)

client/src/portal/mod.rs
```diff
@@ -375,12 +375,29 @@ impl Keyring {
...
+                tokio::task::spawn_blocking(move || keyring.blocking_read().derive_key(&secret))
+                    .await
+                    .unwrap();
+
+            *key_lock = Some(Arc::new(key));
         }
 
-        Arc::clone(self.key.lock().await.as_ref().unwrap())
+        let key = Arc::clone(key_lock.as_ref().unwrap());
+        drop(key_lock);
```
bilelmoussaoui, line 398: the lock would be automatically dropped at the end of function anywayys

warusadura: understood, thanks!

client/src/portal/mod.rs
```diff
@@ -374,12 +374,76 @@ impl Keyring {
...
+    pub async fn change_secret(&mut self, secret: Secret) -> Result<(), Error> {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("Changing keyring secret and key");
+
+        let keyring = self.keyring.read().await;
+        let key = self.derive_key().await;
+        let items: Vec<Item> = keyring
+            .items
+            .iter()
+            .map(|item| item.clone().decrypt(&key).unwrap())
```
bilelmoussaoui, line 414: Don't unwrap here please and instead propagate the error.

warusadura: Done, thanks!

client/src/portal/mod.rs
```diff
@@ -374,12 +374,76 @@ impl Keyring {
...
+            .unwrap();
+
+        let mut key_lock = self.key.lock().await;
+        *key_lock = Some(Arc::new(key));
+        drop(key_lock);
+
+        let mut keyring = self.keyring.write().await;
+        keyring.items.clear();
+
+        let key = self.derive_key().await;
```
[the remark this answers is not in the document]
warusadura: Replaced `let key = self.derive_key().await;`
with `let key = Arc::clone(key_lock.as_ref().unwrap());`

client/src/portal/mod.rs
```diff
@@ -374,12 +374,76 @@ impl Keyring {
...
+        #[cfg(feature = "tokio")]
+        let key = tokio::task::spawn_blocking(move || keyring.blocking_read().derive_key(&secret))
+            .await
+            .unwrap();
+
+        let mut key_lock = self.key.lock().await;
+        *key_lock = Some(Arc::new(key));
+        let key = Arc::clone(key_lock.as_ref().unwrap());
+        drop(key_lock);
+
```
bilelmoussaoui, line 436: all this code is what derive_key does, not sure why you want to duplicate it here

warusadura: yeah, it is duplicate code :)

`derive_key(&self)` only sets the `key` if it's value is [not set](https://github.com/bilelmoussaoui/oo7/blob/dc77973722536a0f985339d9982fe3f5ab0a03c0/client/src/portal/mod.rs#L382), right. since we're changing the Keyring secret here at `change_secret`, we need to derive a key from the new secret and update `Keyring`'s `secret`, `key` fields. otherwise `change_secret` doesn't  work and we get `MacError`.

bilelmoussaoui: I still don't understand the reasoning here, you can unset the key before calling derive_key.

warusadura: yes! that way is much cleaner. done, thanks!

client/src/portal/mod.rs
```diff
@@ -374,12 +374,61 @@ impl Keyring {
...
+        let key = self.derive_key().await;
+
+        let mut items = Vec::new();
+        for item in &keyring.items {
+            items.push(item.clone().decrypt(&key)?);
+        }
+        drop(keyring);
+
+        self.secret = Arc::new(secret);
+        self.key = Default::default();
```
bilelmoussaoui, line 419:
```suggestion
        // Unset the old key
        *self.key.get_mut().unwrap() = None;
```

Default::default works but is not clear what it does in this case

client/src/portal/mod.rs
```diff
@@ -374,12 +374,61 @@ impl Keyring {
...
+    ///
+    /// * `secret` - The new secret to store.
+    pub async fn change_secret(&mut self, secret: Secret) -> Result<(), Error> {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("Changing keyring secret and key");
+
+        let keyring = self.keyring.read().await;
+        let key = self.derive_key().await;
+
+        let mut items = Vec::new();
```
bilelmoussaoui, line 412:
```suggestion
        let mut items = Vec::with_capacity(keyring.items.len());
```

client/src/portal/mod.rs
```diff
@@ -374,12 +374,61 @@ impl Keyring {
...
+            items.push(item.clone().decrypt(&key)?);
+        }
+        drop(keyring);
+
+        self.secret = Arc::new(secret);
+        self.key = Default::default();
+        let key = self.derive_key().await;
+
+        let mut keyring = self.keyring.write().await;
+        keyring.items.clear();
```
bilelmoussaoui, line 423: Is clearing the items enough here? what about the salt/iteration_count & so on? cc @ueno

client/src/portal/mod.rs
```diff
@@ -609,4 +658,29 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn change_secret() -> Result<(), Error> {
+        let path = PathBuf::from("../../tests/test_rekeying.keyring");
+
+        let mut keyring = Keyring::load(&path, strong_key()).await?;
+        let attributes: HashMap<&str, &str> = HashMap::default();
```
bilelmoussaoui, line 667: Please use some attributes here

client/src/portal/mod.rs
```diff
@@ -609,4 +658,29 @@ mod tests {
...
+        keyring
+            .create_item("test", &attributes, "password", false)
+            .await?;
+        let item_before = keyring.lookup_item(&attributes).await?;
+
+        let new_secret = Secret::from(String::from("password").as_bytes().to_vec());
```
bilelmoussaoui, line 673: `b"password"` instead

client/src/portal/mod.rs
```diff
@@ -609,4 +658,29 @@ mod tests {
...
+        keyring.change_secret(new_secret).await?;
+
+        let new_secret = Secret::from(String::from("password").as_bytes().to_vec());
+        let keyring = Keyring::load(&path, new_secret).await?;
+        let item_now = keyring.lookup_item(&attributes).await?;
+
+        assert_eq!(item_before.unwrap().label(), item_now.unwrap().label());
+
+        std::fs::remove_file(path)?; // to allow multiple executions of this test
```
bilelmoussaoui, line 682: Use tokio's fs api to remove the file

client/src/portal/mod.rs
warusadura: yeah, make sense. done.

client/src/portal/mod.rs
warusadura: Done, thanks!

client/src/portal/mod.rs
warusadura: Done.

client/src/portal/mod.rs
warusadura: Done, thanks!

client/src/portal/mod.rs
```diff
@@ -609,4 +659,29 @@ mod tests {
...
+        let item_before = keyring.lookup_item(&attributes).await?;
+
+        let new_secret = Secret::from(b"password".to_vec());
+        keyring.change_secret(new_secret).await?;
+
+        let new_secret = Secret::from(b"password".to_vec());
+        let keyring = Keyring::load(&path, new_secret).await?;
+        let item_now = keyring.lookup_item(&attributes).await?;
+
+        assert_eq!(item_before.unwrap().label(), item_now.unwrap().label());
```
bilelmoussaoui, line 681: The labels are not encrypted so this check alone is not enough. you should also check if the secret is still readable and the same for attributes.

warusadura: yeah! done, thanks!

client/src/portal/mod.rs
```diff
@@ -374,12 +374,62 @@ impl Keyring {
...
+
+        Arc::clone(key_lock.as_ref().unwrap())
+    }
+
+    /// Change keyring secret
+    ///
+    /// # Arguments
+    ///
+    /// * `secret` - The new secret to store.
+    pub async fn change_secret(&mut self, secret: Secret) -> Result<(), Error> {
```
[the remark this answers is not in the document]
warusadura: we need `&mut` :)

client/src/portal/mod.rs
warusadura: I really don't know. What do you think Daiki @ueno?

client/src/portal/mod.rs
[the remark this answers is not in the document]
bilelmoussaoui: for what? we have inner mutability iirc

client/src/portal/mod.rs
```diff
@@ -609,4 +659,40 @@ mod tests {
...
+
+        let new_secret = Secret::from(b"password".to_vec());
+        keyring.change_secret(new_secret).await?;
+
+        let new_secret = Secret::from(b"password".to_vec());
+        let keyring = Keyring::load(&path, new_secret).await?;
+        let item_now = keyring.lookup_item(&attributes).await?;
+
+        assert_eq!(
+            item_before.clone().unwrap().label(),
```
bilelmoussaoui, line 682: All the clones and unwrap here are wrong, unwrap when retrieving the item. But doesn't create_item return the item anyways?

warusadura: yes. done, thanks!

client/src/portal/mod.rs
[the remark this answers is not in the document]
warusadura:
```
error[E0594]: cannot assign to `self.secret`, which is behind a `&` reference
   --> client/src/portal/mod.rs:418:9
    |
418 |         self.secret = Arc::new(secret);
    |         ^^^^^^^^^^^ `self` is a `&` reference, so the data it refers to cannot be written
    |
help: consider changing this to be a mutable reference
    |
405 |     pub async fn change_secret(&mut self, secret: Secret) -> Result<(), Error> {
    |                                ~~~~~~~~~

error[E0596]: cannot borrow `self.key` as mutable, as it is behind a `&` reference
   --> client/src/portal/mod.rs:420:10
    |
420 |         *self.key.get_mut() = None;
    |          ^^^^^^^^ `self` is a `&` reference, so the data it refers to cannot be borrowed as mutable
    |
help: consider changing this to be a mutable reference
    |
405 |     pub async fn change_secret(&mut self, secret: Secret) -> Result<(), Error> {
    |      
```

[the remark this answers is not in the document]
bilelmoussaoui: `self.secret = Arc::new(secret);`

Sounds like the secret field should be behind a Mutex as well.

client/src/portal/mod.rs
```diff
@@ -609,4 +659,30 @@ mod tests {
...
+        let new_secret = Secret::from(b"password".to_vec());
+        keyring.change_secret(new_secret).await?;
+
+        let new_secret = Secret::from(b"password".to_vec());
+        let keyring = Keyring::load(&path, new_secret).await?;
+        let item_now = keyring.lookup_item(&attributes).await?.unwrap();
+
+        assert_eq!(item_before.label(), item_now.label());
+        assert_eq!(item_before.secret(), item_now.secret());
+        assert_eq!(item_before.attributes(), item_now.attributes());
```
bilelmoussaoui, line 682:
```suggestion
        assert_eq!(item_before.attributes(), item_now.attributes());
        assert_eq!(item_now.attributes(), attributes);
```

client/src/portal/mod.rs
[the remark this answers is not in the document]
warusadura: Done, thanks!

client/src/portal/mod.rs
warusadura:
```
assert_eq!(item_now.attributes(), attributes);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `&std::collections::HashMap<std::string::String, AttributeValue> == std::collections::HashMap<&str, &str>`
```

client/src/portal/mod.rs
ueno: That's a good point. I wasn't able to find any recommendation (in NIST SP800-132) that suggests creating a new salt for a new password, though that would be a good practice indeed.

warusadura: Understood, thanks Daiki!

warusadura: Done through https://github.com/bilelmoussaoui/oo7/pull/77/commits/77185a0869588beeea9b62872126552ca83e1946

client/src/portal/api/mod.rs
```diff
@@ -252,6 +252,17 @@ impl Keyring {
             self.iteration_count.try_into().unwrap(),
         )
     }
+
+    pub(crate) fn reset_salt(&mut self) {
+        self.salt = rand::thread_rng().gen::<[u8; DEFAULT_SALT_SIZE]>().to_vec();
+    }
+
+    pub(crate) fn set_modified_time(&mut self) {
```
bilelmoussaoui, line 260: Please merge both methods.

bilelmoussaoui: And also reset iteration_count/usage_count, the salt_len also needs to be updated.

client/src/portal/mod.rs
```diff
@@ -374,12 +374,78 @@ impl Keyring {
...
+        let mut keyring = self.keyring.write().await;
+        keyring.items.clear();
+
+        for item in items {
+            let encrypted_item = item.encrypt(&key)?;
+            keyring.items.push(encrypted_item);
         }
-        self.key.get().unwrap()
+
+        keyring.set_modified_time();
```
bilelmoussaoui, line 445: self.write would end up calling dump which sets the time anyways...

client/src/portal/api/mod.rs
warusadura: Introduced `reset()` to update/reset all the Keyring fields: https://github.com/bilelmoussaoui/oo7/pull/77/commits/4a97e2f129d304a96cbe2a0dfb3372f40a0b6a05

bilelmoussaoui (approved):
Looks good to me, but would appreciate a review / an ack from @ueno

ueno (approved):
LGTM, thanks for all the work here!

PR #79, opened 2024-04-08
Add symlinks to top-level LICENSE file

ueno:
After migration to workspace, each binary subcrates didn't include the LICENSE file, which rust2rpm complains:

```console
  ERROR: No license files were detected. In almost all cases, this is an issue
         with the upstream project that should be reported.
```

This adds symlinks to the top-level LICENSE file so they are properly included.


bilelmoussaoui:
Thanks!

PR #85, opened 2024-05-05
client: Fix compile issue with `tracing` feature

zecakeh:
`PathBuf` has two `AsRef` implementations, so the compiler doesn't know which to choose.

We don't use `.as_ref()` instead and log the debug representation of the path, like in other places.


bilelmoussaoui:
Annoying, thanks for the fix. Will do a bugfix release later

A6GibKm:
Ideally, one ci build tests this

bilelmoussaoui:
> Ideally, one ci build tests this

Was added in a separate PR

PR #86, opened 2024-05-05
ci: Replace actions-rs actions and add clippy checks for more client features

zecakeh:
Replace actions-rs actions because the [whole organization](https://github.com/actions-rs) was archived.

Add clippy checks for more client features to avoid issues like the one in #85 from making it to a release.

Needs #85 to pass CI.


bilelmoussaoui:
A rebase would make the CI happy

zecakeh:
Rebased

PR #87, opened 2024-05-05
client: Do not create features for optional deps already behind a feature

zecakeh:
Makes the [list of features](https://docs.rs/crate/oo7/latest/features) quite long and those features would not have an actual effect in the code.


bilelmoussaoui:
Thanks a lot! do you have more stuff in the pipeline or should I do a 0.3.2 release with the fixes you sent already?

zecakeh:
I have a few clippy nightly lints, so you can do a release now or wait for the PR in the next 5 minutes.

PR #90, opened 2024-05-05
client: expose SecretInner

warusadura:
This change exposes `SecretInner` under unstable feature.
We need this change for the server side implementation.


bilelmoussaoui:
Not truly sure about this, but for now it is good enough

PR #98, opened 2024-07-04
client/item: Replace existing item in `Item::set_attributes`

### LINKED ISSUE #97: `oo7::Item::set_attributes()` with the file backend creates a new item
The reason to call that method should be to update the attributes of an existing item.

However, for the file backend, that method calls `Keyring::create_item()` with `replace: true`, that only replaces items with the same attributes, so if the attributes changed (which is usually the case) a new item is created.

Maybe the implementation should be changed to use `Keyring::replace_item_index()` instead?

zecakeh:
Instead of creating a new one because the attributes have changed.

Fixes #97.

### DIFF
--- a/client/src/keyring.rs
+++ b/client/src/keyring.rs
@@ -238,11 +238,20 @@ impl Item {
     pub async fn set_attributes(&self, attributes: &impl AsAttributes) -> Result<()> {
         match self {
             Self::File(item, backend) => {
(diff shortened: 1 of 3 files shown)


bilelmoussaoui (approved):

bilelmoussaoui:
Thank you!

PR #99, opened 2024-07-22
docs: rework READMEs

xi:
I had a little trouble understanding what this library does and how it fits into the ecosystem. @A6GibKm explained it to me at GUADEC, so I tried to improve the documentation. I mostly tried to clarify that libsecret provides the same backends. Apart from that, these are mostly small tweaks.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -4,13 +4,13 @@
 
 This library allows to store secrets using two different backends:
 
-- The backend defined in the `dbus` module uses the [`org.freedesktop.Secret` service](https://specifications.freedesktop.org/secret-service/latest/) directly.
+- [`dbus`](./src/dbus) implements the [`org.freedesktop.Secret`](https://specifications.freedesktop.org/secret-service/latest/) specification.
 
-- The backend defined in the `portal` module stores secrets in a file that is encrypted using a master secret retrieved from the [`org.freedesktop.portal.Secrets` portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Secret.html). The file format is compatible with libsecret.
+- [`portal`](./src/portal) stores secrets in a file that is encrypted using a secret retrieved from the [`org.freedesktop.portal.Secrets` portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Secret.html). The file format is compatible with libsecret.
 
 Sandboxed applications should prefer using the file backend as it doesn't expose the application secrets to other applications that can talk to the `org.freedesktop.Secrets` service.
 
-The library provides helper methods to automatically pick a backend based on whether the application is sandboxed or not.
+The library provides types that automatically pick a backend based on whether the application is sandboxed or not. Applications developers should use those APIs.
 
 ## Goals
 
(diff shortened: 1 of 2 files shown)


sophie-h:
There is still a part of the readme

> James Bond went on a new mission and this time as a [Secret Service provider](https://specifications.freedesktop.org/secret-service/latest/).

which I think is misleading. A secret service provider would be a daemon, right? That's not part of the repo yet afaik.

client/README.md
```diff
@@ -4,21 +4,21 @@
 
 James Bond went on a new mission and this time as a [Secret Service provider](https://specifications.freedesktop.org/secret-service/latest/).
 
-The library consists of two modules:
+Similar to [libsecret](https://gitlab.gnome.org/GNOME/libsecret), this library allows to store secrets using two different backends:
```
bilelmoussaoui, line 7: Explaining what the library does by comparing to another one that is specific to gnome doesn't make it clearer...

bilelmoussaoui:
> There is still a part of the readme
> 
> > James Bond went on a new mission and this time as a [Secret Service provider](https://specifications.freedesktop.org/secret-service/latest/).
> 
> which I think is misleading. A secret service provider would be a daemon, right? That's not part of the repo yet afaik.

yep, this should stay only in the main readme i guess

client/README.md
xi: How would you describe the relation to libsecret? This is one part that confused me, so I think it is important to explain this clearly. For example, it wasn't clear to me that both libsecret and oo7 provide both a secret service and a file backend.

A6GibKm: I agree here, defining the new thing in terms of the old thing is not ideal. For example:

oo7-client is a library intended to be used by apps (or anything that could need to store/retrieve secret really (clients)) that allows you to
 
- Interact with a encrypted local storage for secrets (sandbox-friendly way) using the Secrets portal (https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Secret.html). 
- Interact with the secret service API (insert link to https://specifications.freedesktop.org/secret-service/latest/)

Of course the difficult part is writing this succinctly.

> How would you describe the relation to libsecret

oo7-client and libsecret are libraries for clients of both secret APIs
oo7-portal and gnome-keyring are servers of the portal API
oo7-(yet-to-be-named https://github.com/bilelmoussaoui/oo7/pull/73) and gnome-keyring are servers of the fdo secret API

> provide both a secret service backend and a file backend.

The FDO api is a *system* api, it is not intended to be used by apps, anyone interacting with it can see every secret. For apps you want the file backend and for system components you want the FDO api.

xi: I removed the reference to libsecret from the top and instead extended the "How does it compare to other libraries?" section. I wonder whether the that section should be moved up because it is easy to miss after the extensive examples. But examples are important, too. What do you think?

portal/README.md
```diff
@@ -0,0 +1,9 @@
+# oo7-portal
+
+[![crates.io](https://img.shields.io/crates/v/oo7-portal)](https://crates.io/crates/oo7-portal) ![CI](https://github.com/bilelmoussaoui/oo7/workflows/CI/badge.svg)
+
+A implementation of [`org.freedesktop.impl.portal.Secret`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.Secret.html)
```
bilelmoussaoui, line 5:
```suggestion
An implementation of [`org.freedesktop.impl.portal.Secret`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.Secret.html)
```

bilelmoussaoui (commented):
otherwise looks good to me, thanks

xi:
Thanks for the suggestions. I liked them and just applied all of them without change.

bilelmoussaoui (approved):

PR #101, opened 2024-08-10
portal: Add meson support

A6GibKm:
# Test on Silverblue

Temporarily disable selinux

    setenforce 0

Create /run/extensions

    mkdir /run/extensions

Have a preferences file `~/.config/xdg-desktop-portal/gnome-portals.conf` with contents

```
[preferred]
default=gnome;gtk;
org.freedesktop.impl.portal.Secret=oo7-portal;
```

and restart xdg-desktop-portal

```
systemctl --user restart xdg-desktop-portal
```

From oo7/portal run in a toolbox:

    meson --prefix=/usr _build -Dprofile=development
    ninja -C _build && DESTDIR=oo7-extension meson install -C _build

From outside the sandbox:

```
cp -r _build/oo7-extension /run/extensions/ && systemd-sysext refresh --force
```

One can get more debug output by replacing services for xdp and oo7-portals:
```
/usr/libexec/xdg-desktop-portal -r -v
```

```
RUST_LOG=oo7_portal=debug RUST_BACKTRACE=full /usr/libexec/oo7-portal
```

### DIFF
--- a/portal/Cargo.toml
+++ b/portal/Cargo.toml
@@ -21,5 +21,5 @@ serde.workspace = true
 tokio = { workspace = true, features = ["io-util", "net", "macros", "rt-multi-thread"] }
 tracing = "0.1"
 tracing-subscriber.workspace = true
-zbus.workspace = true
+zbus = { workspace = true, features = ["tokio"] }
 zeroize.workspace = true
--- a/portal/Cargo.toml
+++ b/portal/Cargo.toml
@@ -12,6 +12,7 @@ rust-version.workspace = true
 version.workspace = true
 
 [dependencies]
+clap.workspace = true
 futures-channel.workspace = true
 futures-util.workspace = true
 oo7.workspace = true
--- a/portal/src/main.rs
+++ b/portal/src/main.rs
@@ -129,6 +129,12 @@ async fn send_secret_to_app(app_id: &str, fd: zvariant::OwnedFd) -> Result<(), E
 async fn main() -> Result<(), zbus::Error> {
     tracing_subscriber::fmt::init();
 
+    tracing::info!(
+        "Initializing {} {}",
+        env!("CARGO_BIN_NAME"),
+        env!("CARGO_PKG_VERSION")
+    );
+
     let backend = Secret;
     let cnx = zbus::ConnectionBuilder::session()?
         .serve_at(oo7::portal::SecretProxy::PATH.unwrap(), backend)?
--- a/portal/src/main.rs
+++ b/portal/src/main.rs
@@ -3,6 +3,7 @@ mod request;
 
 use std::{collections::HashMap, future::pending, os::unix::net::UnixStream};
 
+use clap::Parser;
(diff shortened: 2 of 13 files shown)


portal/Cargo.toml
```diff
@@ -14,12 +14,12 @@ version.workspace = true
 [dependencies]
 futures-channel.workspace = true
 futures-util.workspace = true
-oo7 = { workspace = true, features = ["unstable"] }
+oo7 = { workspace = true, default-features = false, features = ["tokio", "native_crypto", "unstable"] }
```
bilelmoussaoui, line 17:
```suggestion
oo7 = { workspace = true, features = ["tokio", "native_crypto", "unstable"] }
```
you set that in the workspace already, also it is weird that the cli uses native crypto & async-std and the portal uses native / tokio, which are the default-features anyway

portal/src/main.rs
```diff
@@ -139,6 +141,8 @@ async fn main() -> Result<(), zbus::Error> {
         | zbus::fdo::RequestNameFlags::AllowReplacement;
     cnx.request_name_with_flags(PORTAL_NAME, flags).await?;
 
+    tracing::info!("Got dbus name {PORTAL_NAME}");
```
bilelmoussaoui, line 144: what is this for ?

portal/src/meson.build
```diff
@@ -0,0 +1,30 @@
+cargo = find_program('cargo', required: true)
```
bilelmoussaoui, line 1:
```suggestion
cargo = find_program('cargo')
```

portal/src/meson.build
```diff
@@ -0,0 +1,30 @@
...
+
+is_devel = true
```
bilelmoussaoui, line 3: should this stay true?

portal/src/main.rs
A6GibKm: Just to know that the binary reached the "main" loop without erring out, but it might be unnecessary.

A6GibKm: I removed it, if the process fails it will just close.

portal/Cargo.toml
A6GibKm: I removed the redundant default-features.

portal/src/main.rs
```diff
@@ -125,18 +126,33 @@ async fn send_secret_to_app(app_id: &str, fd: zvariant::OwnedFd) -> Result<(), E
...
+    replace: bool,
+}
+
 #[tokio::main]
 async fn main() -> Result<(), zbus::Error> {
+    let args = Args::parse();
+
     tracing_subscriber::fmt::init();
 
+    tracing::info!("Initializing oo7-portal {}", env!("CARGO_PKG_VERSION"));
```
bilelmoussaoui, line 144:
```suggestion
    tracing::info!("Initializing {} {}", env!("CARGO_BIN_NAME"), env!("CARGO_PKG_VERSION"));
```

PR #103, opened 2024-08-10
daemon: Add meson

A6GibKm:


server/meson_options.txt
```diff
@@ -0,0 +1,26 @@
...
+
+option (
+  'profile',
+  type: 'combo',
+  choices: [
+    'default',
+    'development'
+  ],
+  value: 'default',
+  description: 'The build profile for Obfuscate. One of "default" or "development".'
```
bilelmoussaoui, line 25: is it for obfuscate?

A6GibKm: Is it possible that i used the same copy paste inthe portal too? Lol

bilelmoussaoui: seems to be!

PR #109, opened 2024-08-13
 portal: Switch to using ashpd 

bilelmoussaoui:


A6GibKm:
Ok, this is more code than I expected. I wonder what `cargo update` when you are generating the updated lock file.

bilelmoussaoui:
> Ok, this is removes more code than I expected. 

Yes, and the API will only get better with time once there are actual usages of it + it would allow us to stay up to date here in case the portal ends up gaining new features / changes.

> I wonder what cargo update when you are generating the updated lock file.

Not sure what you mean

bilelmoussaoui:
what, this wasn't merged??

PR #111, opened 2024-08-18
client/dbus: Allow setting window id

bilelmoussaoui:
A primary step to make the library / portal & in the future, the backend as well cross-desktop


bilelmoussaoui:
This looks like the correct approach going forward, especially if we want to untangle the server side implementation from being GNOME specific. So I am going to land this

PR #114, opened 2024-08-18
cli: Replace chrono with time

bilelmoussaoui:
as it is much smaller and fits better the usage in the cli

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -9,6 +9,7 @@ use oo7::{
     dbus::{Collection, Service},
     AsAttributes,
 };
+use time::{OffsetDateTime, UtcOffset};
 
 const BINARY_NAME: &str = env!("CARGO_BIN_NAME");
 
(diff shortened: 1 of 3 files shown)


A6GibKm:
As per the docs of [parse](https://docs.rs/time/latest/time/format_description/fn.parse.html):

> This function exists for backward compatibility reasons. It is equivalent to calling parse_borrowed::<1>(s). In the future, this function will be deprecated in favor of parse_borrowed.

It might be a good idea to switch to `parse_borrowed::<2>`.

cli/src/main.rs
```diff
@@ -116,7 +117,7 @@ struct Cli {
     command: Commands,
 }
 
-#[tokio::main]
+#[tokio::main(flavor = "current_thread")]
```
A6GibKm, line 136: Is there a specific reason for this change?

bilelmoussaoui: Because this simple CLI app doesn't need a multi threaded runtime to do basic zbus calls

PR #115, opened 2024-08-18
portal: Switch to using ashpd

bilelmoussaoui:
re-opening https://github.com/bilelmoussaoui/oo7/pull/109


bilelmoussaoui:
The API in ashpd is still not perfect, but it is good enough for a start

PR #118, opened 2024-08-24
portal: meson: Set fallback dbus_service_dir

A6GibKm:
Needed to install the dbus activation file without specifiying the directory.

### DIFF
--- a/portal/meson.build
+++ b/portal/meson.build
@@ -6,6 +6,7 @@ project(
 )
 
 dbus_name = 'org.freedesktop.impl.portal.desktop.oo7'
+service_name = meson.project_name()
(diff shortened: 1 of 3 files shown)


portal/meson.build
```diff
@@ -30,6 +34,7 @@ portal_dir = datadir / 'xdg-desktop-portal' / 'portals'
 conf = configuration_data()
 conf.set('libexecdir', prefix / libexecdir)
 conf.set('dbus_name', dbus_name)
+conf.set('service_name', service_name)
```
bilelmoussaoui, line 37: why not use meson.project_name?

A6GibKm: I set `service_name = meson.project_name()`.

PR #122, opened 2024-09-02
 portal: Keyring::load() verify secret before returning a Keyring

warusadura:
Adds `verify_secret()`.

With this change, calls to `oo7::portal::Keyring::load()`, and
`oo7::portal::Keyring::open()` will fail when used with a
wrong secret/password.

Also adds `oo7::portal::Keyring::verify_secret()` API and introduces `InvalidSecret` error.

### DIFF
--- a/client/src/portal/api/mod.rs
+++ b/client/src/portal/api/mod.rs
@@ -259,6 +259,31 @@ impl Keyring {
         )
     }
 
+    pub(crate) fn verify_secret(&self, secret: &Secret) -> bool {
+        let key = self.derive_key(secret);
+
+        // if this is a new keyring or a keyring with zero items
+        if self.items.is_empty() {
+            return true;
+        }
+
+        // decrypting the first item may be enough to verify the secret.
+        // but to avoid a scenario where the first item is corrupted,
+        // decrypting all the items here.
+        let mut decrypt_counter = 0;
+        for item in self.items.clone() {
+            if item.decrypt(&key).is_ok() {
+                decrypt_counter += 1;
+            }
+        }
+
+        if decrypt_counter == self.items.len() {
+            return true;
+        }
+
+        false
+    }
+
     // Reset Keyring content
     pub(crate) fn reset(&mut self) {
         let salt = rand::thread_rng().gen::<[u8; DEFAULT_SALT_SIZE]>().to_vec();
--- a/client/src/portal/mod.rs
+++ b/client/src/portal/mod.rs
@@ -454,6 +454,33 @@ impl Keyring {
 
         self.write().await
     }
+
+    /// Verify the secret associated with a keyring.
+    ///
+    /// This function supports all keyring formats (legacy and latest).
+    ///
+    /// # Arguments
+    /// *`name` - The name of the keyring.
+    /// * `secret` - The service key, usually retrieved from the Secrets portal.
+    pub async fn verify_secret(name: &str, secret: Secret) -> bool {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("Trying to verify {} keyring secret", name);
+
+        let mut ret = false;
+
+        match Self::open(name, secret).await {
+            Err(err) => {
+                if matches!(err, Error::InvalidSecret) {
+                    ret = false;
+                }
+            }
+            Ok(_) => {
+                ret = true;
+            }
+        }
+
+        ret
+    }
 }
 
 #[cfg(test)]
@@ -741,6 +768,34 @@ mod tests {
         assert_eq!(item_before.attributes(), item_now.attributes());
 
         fs::remove_file(path).await?;
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn verify_secret() -> Result<(), Error> {
+        let temp_dir = tempdir()?;
+        let keyring_dir = temp_dir.path().join("keyrings");
+
+        fs::create_dir_all(&keyring_dir).await?;
+
+        let fixture_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("fixtures")
+            .join("legacy.keyring");
+
+        fs::copy(&fixture_path, &keyring_dir.join("legacy.keyring")).await?;
+
+        std::env::set_var("XDG_DATA_HOME", &temp_dir.path());
+
+        // test with a wrong secret
+        let password = b"wrong";
+        let secret = Secret::from(password.to_vec());
+        assert_eq!(Keyring::verify_secret("legacy", secret).await, false);
+
+        // test with the correct secret
+        let password = b"test";
+        let secret = Secret::from(password.to_vec());
+        assert!(Keyring::verify_secret("legacy", secret).await);
 
         Ok(())
     }
--- a/client/src/portal/mod.rs
+++ b/client/src/portal/mod.rs
@@ -120,6 +120,10 @@ impl Keyring {
 
                 let keyring = api::Keyring::try_from(content.as_slice())?;
 
+                if !keyring.verify_secret(&secret) {
+                    return Err(Error::InvalidSecret);
+                }
+
                 (mtime, keyring)
             }
         };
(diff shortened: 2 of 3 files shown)


client/src/portal/mod.rs
```diff
@@ -210,6 +210,49 @@ impl Keyring {
...
+        let mut ret = false;
+
+        let items = keyring.items().await;
+        for item in items {
+            match item {
+                Ok(_) => {
+                    #[cfg(feature = "tracing")]
+                    tracing::debug!("{} keyring secret matches", name);
+                    ret = true;
+                    break;
```
A6GibKm, line 243: change this to `return true`. And remove the `let mut ret`, you can just return `false` at the end.

client/src/portal/mod.rs
```diff
@@ -210,6 +210,49 @@ impl Keyring {
...
+                tracing::debug!("Wrong secret for {}: {}", name, _err);
+                return false;
+            }
+        };
+
+        let mut ret = false;
+
+        let items = keyring.items().await;
+        for item in items {
+            match item {
```
A6GibKm, line 238: There is probably one method in the `Iterator` trait to verify all of the items are `Ok`.

client/src/portal/mod.rs
```diff
@@ -210,6 +210,49 @@ impl Keyring {
...
 
+    /// Verify the secret associated with a keyring.
+    ///
+    /// This function supports all keyring formats (legacy and latest).
+    ///
+    /// # Arguments
+    ///
+    /// *`name` - The name of the keyring.
+    /// * `secret` - The service key, usually retrieved from the Secrets portal.
+    pub async fn verify_keyring_secret(name: &str, secret: Secret) -> bool {
```
bilelmoussaoui, line 221: I am not sure this approach is right at all. Why would you just verify if it is the right secret without returning the corresponding keyring if the secret is ok? why this can't be done by default? there are various questions that needs answers before going on  with an implementation

client/src/portal/mod.rs
warusadura: got it, thanks @A6GibKm for the prompt review!

client/src/portal/mod.rs
warusadura: yeah, I agree. I got the same feeling afterwards and that's why I have marked this as a draft :)

client/src/portal/mod.rs
```diff
@@ -679,6 +710,82 @@ mod tests {
...
+                assert!(matches!(err, Error::ChecksumMismatch));
+            }
+            Ok(_) => assert!(true),
+        };
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn repeated_open() -> Result<(), Error> {
```
bilelmoussaoui, line 752: where this test came from?

bilelmoussaoui: I am not sure in general what all those tests are intended to verify. You should add one test here in general, which tries to open a fixture file that exists already with the wrong password which would be passing when it shouldn't. The next commit should do the changes for the test's behaviour to be fixed.

client/src/portal/mod.rs
```diff
@@ -450,6 +454,33 @@ impl Keyring {
...
+    /// # Arguments
+    /// *`name` - The name of the keyring.
+    /// * `secret` - The service key, usually retrieved from the Secrets portal.
+    pub async fn verify_secret(name: &str, secret: Secret) -> bool {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("Trying to verify {} keyring secret", name);
+
+        let mut ret = false;
+
+        match Self::open(name, secret).await {
```
bilelmoussaoui, line 471: Not sure why I would open the keyring file a second time, when all of this verification the moment you call open.

client/src/portal/mod.rs
```diff
@@ -120,6 +120,10 @@ impl Keyring {
 
                 let keyring = api::Keyring::try_from(content.as_slice())?;
 
+                if !keyring.verify_secret(&secret) {
+                    return Err(Error::ChecksumMismatch);
```
bilelmoussaoui, line 124: Return a new error type please.

client/src/portal/api/mod.rs
```diff
@@ -259,6 +259,22 @@ impl Keyring {
...
+    pub(crate) fn verify_secret(&self, secret: &Secret) -> bool {
+        let key = self.derive_key(secret);
+
+        // if this is a new keyring or a keyring with zero items
+        if self.items.is_empty() {
+            return true;
+        }
+
+        // attempting to decrypt the first item
+        if self.items[0].clone().decrypt(&key).is_ok() {
```
bilelmoussaoui, line 271: Should this really call decrypt? Is trying to decrypt one item enough?

A6GibKm, line 271: Regarding style, 

`if let Some(value) = self.items.get(0)` is simpler than asking if the vec is empty and then calling `[0]`.

client/src/portal/mod.rs
warusadura: this test comes from a separate commit, after introducing `verify_secret()`. apologies if this isn't clearer.

warusadura: > why we need it

repeated calls to `oo7::portal::Keyring::open()` failing to identify/validate incorrect keyring secret prompted this entire pr.

so, oo7_daemon uses `Keyring::open()` during `oo7_daemon -l` command (-l option reads keyring secret from stdin).

currently during the very first call to `Keyring::open()` everything works as expected (keyring migration validates the keyring secret). after that, calls to `Keyring::open()` succeed (so `oo7_daemon -l` as well) even with an incorrect keyring secret. this is what I'm trying to fix and this test make sure that introducing `verify_secret()` fixes that issue.

client/src/portal/mod.rs
warusadura: what second time? :) 

this is a new function: `verify_secret`. we need this for the server side implementation. specifically to verify the secret during Prompt based unlock. this function just validates the secret and it doesn't return a keyring.

client/src/portal/mod.rs
warusadura: what should be the new error?

client/src/portal/api/mod.rs
warusadura: > Should this really call decrypt?

yes, I think so. I don't see any other way to verify the keyring secret.

> Is trying to decrypt one item enough?

yes, I think so. all secrets (items) are encrypted using the same key.

cc @ueno

ueno: I stumbled upon this as well; while it might be out of scope of our threat model, it is possible that the first item is corrupted. Maybe good to decrypt all items or at least have a comment explaining why decrypting the first item is sufficient.

warusadura: understood, thanks Daiki!

client/src/portal/mod.rs
bilelmoussaoui: load -> verify_secret -> open -> you end up opening reading the keyring twice. I am not sure the changes you made here even correct because on the other hand open would call load as well.... Not good.

bilelmoussaoui: The reason of your confusion, is there is a method load, that is used only when interacting with a secret that comes from the secret portal. So it is assumed to be always correct where open is what you would use in the server side. 

The secret verification should be done to both cases just in case the API is misused.

client/src/portal/mod.rs
warusadura: done, thanks!

client/src/portal/mod.rs
warusadura: > load -> verify_secret -> open -> you end up opening reading the keyring twice. I am not sure the changes you made here even correct because on the other hand open would call load as well.... Not good.

Thanks. sorry for the late reply.

apologies if what I'm saying doesn't make any sense :)

so, when I call `oo7::portal::Keyring::verify_secret()` what happens is,
* calling `Keyring::open()` right,
	- if the keyring is v0, `Keyring::open()` will call `Keyring::migrate()`. in this case the keyring file is opened once and the secret verification is taken care [here](https://github.com/bilelmoussaoui/oo7/blob/e9259e72ccac6e2317fa9db4c6a3b146c927ec60/client/src/portal/mod.rs#L162). there isn't another call to `Keyring::load()`.
	- if the keyring is v1 `Keyring::open()` will call `Keyring::load()`. in this case the keyring file is opened once and secret verification is done through [verify_secret()](https://github.com/bilelmoussaoui/oo7/blob/a16c422936572a1cf9dd1ad93ca4fec2b4ce0837/client/src/portal/api/mod.rs#L262).
	
so, for both cases: `Keyring::load()`, `Keyring::open()` secret verification are covered now and the keyring file is opened once during the vertification. am I still missing something here?

client/src/portal/api/mod.rs
warusadura: Now, I'm decrypting all the items. Does it still make sense to follow this approach?

bilelmoussaoui:
I pushed a simpler implementation at https://github.com/bilelmoussaoui/oo7/commit/3ba3e8d9c66c9cf797c57c645d09f3e042e69693

client/src/portal/mod.rs
bilelmoussaoui: What I meant is:

- If you use `Keyring::open`, 
  * We verify if v1 exists -> call Keyring::load on it
  * if v0 exists, call Keyring::migrate which by itself verifies the secret as it decrypts the v0 items to migrate them to v1

- So basically, by handling the verification in Keyring::load code path, you cover all the use cases.

PR #124, opened 2024-09-07
Make the cli easier to pipe and add examples

A6GibKm:


A6GibKm:
At the moment when using lookup from a non interactive terminal it prints the formated table, but without newlines at the end, I think `--secret-only` should be the default behavior when not in a tty.

cli/src/main.rs
```diff
@@ -174,11 +174,18 @@ where
 async fn store(label: &str, attributes: &impl AsAttributes) -> Result<(), Error> {
     let collection = collection().await?;
 
-    print!("Type a secret: ");
-    std::io::stdout()
-        .flush()
-        .map_err(|_| Error::new("Could not flush stdout"))?;
-    let secret = rpassword::read_password().map_err(|_| Error::new("Can't read password"))?;
+    let mut stdin = std::io::stdin().lock();
```
bilelmoussaoui, line 177: The commit mentions lookup but then you are changing the store method. Please update the commit message & also include a brief justification of why this change is necessary

A6GibKm: Done

PR #125, opened 2024-09-07
cli: Add collection arg

A6GibKm:


portal/src/main.rs
```diff
@@ -49,7 +49,10 @@ async fn send_secret_to_app(app_id: &AppID, fd: std::os::fd::OwnedFd) -> Result<
     let collection = match service.default_collection().await {
         Err(oo7::dbus::Error::NotFound(_)) => {
             service
-                .create_collection("Default", Some(oo7::dbus::DEFAULT_COLLECTION))
+                .create_collection(
+                    oo7::dbus::DEFAULT_COLLECTION_LABEL,
+                    Some(oo7::dbus::DEFAULT_COLLECTION),
+                )
```
bilelmoussaoui, line 55: instead, maybe we should have a `create_default` helper?

bilelmoussaoui: I think that is fine to expose under unstable, make it easier / simpler to fix things.

client/src/dbus/mod.rs
```diff
@@ -37,6 +37,9 @@
 /// In general, you are supposed to use [`Service::default_collection`].
 pub const DEFAULT_COLLECTION: &str = "default";
 
+/// The default collection label.
+pub const DEFAULT_COLLECTION_LABEL: &str = "Default";
```
bilelmoussaoui, line 41: Another open question I had, is there some sort of "default" label for the session collection?

A6GibKm: I don't think so, but I can dig later the code of gnome-keyring. This machine has a clean install and uses "Login" so maybe that is better.

A6GibKm: libsecret/libsecret/secret-methods.c:1058 uses 

```
		properties = _secret_collection_properties_new (_("Default keyring"));
		secret_service_create_collection_dbus_path (service, properties, "default",
```

whereas gnome-keyring/daemon/login/gdk-login.c uses
```
static GckObject*
create_login_keyring (GckSession *session, GckObject *cred, GError **error)
{
	GckBuilder builder = GCK_BUILDER_INIT;

	g_return_val_if_fail (GCK_IS_SESSION (session), NULL);
	g_return_val_if_fail (GCK_IS_OBJECT (cred), NULL);

	gck_builder_add_ulong (&builder, CKA_CLASS, CKO_G_COLLECTION);
	gck_builder_add_string (&builder, CKA_ID, "login");
	gck_builder_add_ulong (&builder, CKA_G_CREDENTIAL, gck_object_get_handle (cred));
	gck_builder_add_boolean (&builder, CKA_TOKEN, TRUE);

	/* TRANSLATORS: This is the display label for the login keyring */
	gck_builder_add_string (&builder, CKA_LABEL, _("Login"));

	return gck_session_create_object (session, gck_builder_end (&builder), NULL, error);
}
```

bilelmoussaoui:
Can you rebase this one please?

A6GibKm:
> Can you rebase this one please?

Rebased, but I think we should use `Login`. Note too that this string is translatable in other projects.

bilelmoussaoui:
> > Can you rebase this one please?
> 
> Rebased, but I think we should use `Login`. Note too that this string is translatable in other projects.

Hmm, that makes providing both the helper methods & constants weird. As the CLI tool could become translatable, not the oo7 library itself...

A6GibKm:
> > > Can you rebase this one please?
> > 
> > 
> > Rebased, but I think we should use `Login`. Note too that this string is translatable in other projects.
> 
> Hmm, that makes providing both the helper methods & constants weird. As the CLI tool could become translatable, not the oo7 library itself...

Sounds like a problem for tomorrow, but it makes sense to have it translatable everywhere.

bilelmoussaoui:
> > > > Can you rebase this one please?
> > > 
> > > 
> > > Rebased, but I think we should use `Login`. Note too that this string is translatable in other projects.
> > 
> > 
> > Hmm, that makes providing both the helper methods & constants weird. As the CLI tool could become translatable, not the oo7 library itself...
> 
> Sounds like a problem for tomorrow, but it makes sense to have it translatable everywhere.

I went ahead and add various helper APIs here that makes the whole thing nicer while allowing people to translate things if they want to.

cli/src/main.rs
```diff
@@ -131,27 +131,35 @@ enum Commands {
...
+        long,
+        help = "Specify a collection. The default collection will be used if not specified"
+    )]
+    collection: Option<String>,
 }
 
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> Result<(), Error> {
     let cli = Cli::parse();
+    let collection = collection(cli.collection.as_deref()).await?;
```
bilelmoussaoui, line 146: The last commit should make the amount of changes here minimal.

A6GibKm:
rebased, but please dont merge, the arg is not properly working , e.g.

```
$ cargo run -- search --collection c a=b
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.40s
     Running `/var/home/msandova/Projects/oo7/target/debug/oo7-cli search --collection c a=b`
error: unexpected argument '--collection' found

  tip: to pass '--collection' as a value, use '-- --collection'

Usage: oo7-cli search [OPTIONS] [ATTRIBUTES]...

For more information, try '--help'.
```

cli/src/main.rs
```diff
@@ -124,9 +124,19 @@ enum Commands {
...
         let service = Service::new().await?;
-        let collection = service.default_collection().await?;
+        let collection = if let Some(alias) = &args.collection {
+            match service.with_alias(alias).await {
+                Ok(Some(collection)) => Ok(collection),
+                Ok(None) => Err(oo7::dbus::Error::NotFound(format!(
+                    "Collection {alias} not found"
+                ))),
+                Err(err) => Err(err),
+            }
```
bilelmoussaoui, line 136:
```suggestion
            match service.with_alias(alias).await?.unwrap_or_else(|| ...)`
```
is the same, no?

bilelmoussaoui (commented):
otherwise lgtm, once you solved the cli arg parser issue :)

cli/src/main.rs
A6GibKm: Had to use ok_or_else though, but yes.

A6GibKm:
> otherwise lgtm, once you solved the cli arg parser issue :)

Done, it was a matter of using global=true.

bilelmoussaoui (approved):

PR #129, opened 2024-09-22
portal: Set nonblocking on unix stream

A6GibKm:
As per tokio's docs:

The caller is responsible for ensuring that the stream is in non-blocking mode. Otherwise all I/O operations on the stream will block the thread, which will cause unexpected behavior. Non-blocking mode can be set using set_nonblocking.


bilelmoussaoui:
Lgtm, thanks

PR #130, opened 2024-09-22
portal: Set generic xdg:schema

A6GibKm:


client/src/dbus/mod.rs
```diff
@@ -59,3 +59,6 @@ pub use error::{Error, ServiceError};
 pub use service::Service;
 mod collection;
 pub use collection::Collection;
+
+pub const XDG_SCHEMA: &str = "xdg:schema";
```
bilelmoussaoui, line 63: Both should probably be inside the AsAttributes trait with a comment explaining what it is useful for

bilelmoussaoui:
I took the liberty to make some changes by moving the schema const outside of the dbus module as it is not specific to it.

PR #137, opened 2024-10-15
server: adds minimal Secret Service server implementation

warusadura:
This change adds the *minimal server side interface implementation of the freedesktop Secret Service API. The implementation is NOT complete. This change provides a foundation for adding further, more accurate implementations on top of it.

*minimal: Included all the Secret Service API method signatures without any implementation (todo!). Also, added a simple main().

Secret Service API spec: https://specifications.freedesktop.org/secret-service-spec/latest/

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -814,6 +814,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "78b3ae25bc7c8c38cec158d1f2757ee79e9b3740fbc7ccf0e59e4b08d793fa89"
 
 [[package]]
+name = "lock_api"
+version = "0.4.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07af8b9cdd281b7915f413fa73f29ebd5d55d0d3f0155584dade1ff18cea1b17"
+dependencies = [
+ "autocfg",
+ "scopeguard",
+]
+
+[[package]]
 name = "log"
 version = "0.4.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1057,6 +1067,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "oo7-daemon"
+version = "0.3.0"
+dependencies = [
+ "oo7",
+ "serde",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
+ "zbus",
+]
+
+[[package]]
 name = "oo7-portal"
 version = "0.3.0"
 dependencies = [
@@ -1129,6 +1151,29 @@ name = "parking"
 version = "2.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f38d5652c16fde515bb1ecef450ab0f6a219d619a7274976324d5e377f7dceba"
+
+[[package]]
+name = "parking_lot"
+version = "0.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f1bf18183cf54e8d6059647fc3063646a1801cf30896933ec2311622cc4b9a27"
+dependencies = [
+ "lock_api",
+ "parking_lot_core",
+]
+
+[[package]]
+name = "parking_lot_core"
+version = "0.9.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e401f977ab385c9e4e3ab30627d6f26d00e2c73eef317493c4ec6d468726cf8"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "redox_syscall",
+ "smallvec",
+ "windows-targets 0.52.6",
+]
 
 [[package]]
 name = "pbkdf2"
@@ -1263,6 +1308,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "redox_syscall"
+version = "0.5.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b6dfecf2c74bce2466cabf93f6664d6998a69eb21e39f4207930065b27b771f"
+dependencies = [
+ "bitflags",
+]
+
+[[package]]
 name = "rpassword"
 version = "7.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1307,6 +1361,12 @@ name = "ryu"
 version = "1.0.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f3cb5ba0dc43242ce17de99c180e96db90b235b8a9fdc9543c96d2209116bd9f"
+
+[[package]]
+name = "scopeguard"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 
 [[package]]
 name = "serde"
@@ -1558,6 +1618,7 @@ dependencies = [
  "bytes",
  "libc",
  "mio",
+ "parking_lot",
  "pin-project-lite",
  "signal-hook-registry",
  "socket2",
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "oo7-daemon"
+edition.workspace = true
+exclude.workspace = true
+homepage.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
+oo7 = { workspace = true, features = ["unstable"] }
+serde.workspace = true
+tokio = { workspace = true, features = ["full"] }
+tracing = "0.1"
+tracing-subscriber.workspace = true
+zbus.workspace = true
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -1,0 +1,21 @@
+mod collection;
+mod item;
+mod prompt;
+mod service;
+mod session;
+
+use service::{Result, Service};
+
+const BINARY_NAME: &str = env!("CARGO_BIN_NAME");
+
+#[tokio::main]
+async fn main() -> Result<()> {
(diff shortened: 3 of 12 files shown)


warusadura:
I think this pr is still too large :) I'll try to make this smaller.

bilelmoussaoui:
I would expect a first PR that just adds service.rs with the bare minimum required methods. Or all the files but without any implementations and just todo! calls

warusadura:
before: `733` lines, now: `324`. I have removed all the implementations. I think this provides a good base now :)

server/src/daemon/error.rs
```diff
@@ -0,0 +1,21 @@
+// Errors
+
+#[derive(zbus::DBusError, Debug)]
+#[zbus(prefix = "org.freedesktop.Secret.Error")]
+pub enum ServiceError {
```
bilelmoussaoui, line 5: The exact same error is defined https://github.com/bilelmoussaoui/oo7/blob/main/client/src/dbus/error.rs#L6 already, make use of that instead

server/src/daemon/error.rs
```diff
@@ -0,0 +1,21 @@
...
+    NoSession,
+    // Collection/Item does not exist.
+    NoSuchObject,
+}
+
+impl From<oo7::portal::Error> for ServiceError {
+    fn from(value: oo7::portal::Error) -> Self {
+        Self::ZBus(zbus::Error::Failure(value.to_string()))
+    }
+}
```
bilelmoussaoui, line 21: This is not used now

server/src/daemon/item.rs
```diff
@@ -0,0 +1,27 @@
+// org.freedesktop.Secret.Item
+
+use oo7::dbus::api::SecretInner;
+use zbus::zvariant::{ObjectPath, OwnedObjectPath};
+
+use super::Result;
+
+#[derive(Debug)]
+pub struct Item {
+    _path: OwnedObjectPath,
```
bilelmoussaoui, line 10: Drop the unused fields

server/src/daemon/item.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+#[zbus::interface(name = "org.freedesktop.Secret.Item")]
+impl Item {
+    pub async fn delete(&self) -> Result<ObjectPath> {
+        todo!()
+    }
+
+    #[zbus(name = "GetSecret")]
```
bilelmoussaoui, line 19: The methods should name the in & out arguments so they are correctly displayed on apps like d-spy

server/src/daemon/service.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    pub async fn set_alias(&self, _name: &str, _collection: ObjectPath<'_>) -> Result<()> {
+        todo!()
+    }
+}
+
+impl Service {
+    pub async fn new() -> Self {
+        Self {
+            cnx: Default::default(),
```
bilelmoussaoui, line 84: Why the cnx should be a mutex ? Just remove the new method and have run create the connection itself

server/src/daemon/service.rs
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    pub async fn run(self) -> Result<()> {
+        let cnx = zbus::connection::Builder::session()?
+            .name("org.freedesktop.secrets.Devel")?
+            .build()
+            .await?;
+        *self.cnx.lock().unwrap() = Some(cnx.clone());
+        cnx.object_server()
```
A6GibKm, line 94: Can this be done before the previous line and then not have to do the `clone()`?

server/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "oo7-daemon"
```
bilelmoussaoui, line 2: This needs to be added to the CI as well so that it always passes before merging any further things

Cargo.lock
```diff
@@ -813,6 +813,16 @@ version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "78b3ae25bc7c8c38cec158d1f2757ee79e9b3740fbc7ccf0e59e4b08d793fa89"
 
+[[package]]
```
bilelmoussaoui, line 816: The updates to Cargo.lock are irrelevant, please split into a separate commit.

bilelmoussaoui (commented):
much better!

server/src/main.rs
```diff
@@ -0,0 +1,17 @@
+mod daemon;
```
bilelmoussaoui, line 1: I don't understand the reason for having this in a separate mod; it is not like we are going to expose it as a separate library anyways. So drop the folder. initially, i was suggesting to name "server" as "daemon" instead of adding an extra mod.

server/src/daemon/service.rs
A6GibKm: Anotherway of saying this,  the `cnx` should only be set once. There is no need to set a `default` dummy value.

server/src/daemon/error.rs
warusadura: done, thanks for the prompt review.

server/src/daemon/item.rs
warusadura: done.

server/src/daemon/error.rs
warusadura: removed.

server/src/daemon/service.rs
warusadura: understood, thanks! @A6GibKm

server/src/main.rs
warusadura: sorry for the mix up. this is fixed now.

server/src/daemon/item.rs
warusadura: I couldn't find a way to name `IN` arguments. but, I have named all the `OUT` args.

server/Cargo.toml
warusadura: looking into this now.

warusadura: done.

server/src/daemon/item.rs
bilelmoussaoui: The in argument are the name you give to the function parameters

Cargo.lock
bilelmoussaoui: Please update the commit to just say "update dependencies"

bilelmoussaoui:
Please update the commits like the following:
- Add a minimal Secret Service server implementation
and include the CI bits in it
- Rename the update dependencies one to "Update dependencies" and put it before the server impl one
otherwise good to be merged!

server/src/daemon/item.rs
warusadura: got it, thanks!

Cargo.lock
warusadura: done, updated commit names and order.

A6GibKm:
This is not necessarily something that should be addressed in this PR, but I am not particularly happy with putting the project in a `daemon` folder, the reason being that there are actually two daemons, the portal and the fdo secrets interface provider.

warusadura:
> This is not necessarily something that should be addressed in this PR, but I am not particularly happy with putting the project in a `daemon` folder, the reason being that there are actually two daemons, the portal and the fdo secrets interface provider.

I'm happy to rename the project dir. wdyt @bilelmoussaoui 

> fdo secrets interface provider.

this is not done yet right (as in not in oo7)? is there a place I can read more about this? I don't know anything about fdo :) @A6GibKm

A6GibKm:
https://www.freedesktop.org/wiki/Specifications/secret-storage-spec/secrets-api-0.1.html freedesktop being a reference to it being a freedesktop spec. 

Well `server` has the same problem. Again, I don't think this should be addressed here (or if it does need addressing), but it is something to consider for the long run.

warusadura:
my bad I was thinking about FIDO :D

bilelmoussaoui:
> Well server has the same problem. Again, I don't think this should be addressed here (or if it does need addressing), but it is something to consider for the long run.

Tbh, the portal is already clearly named. Server matches the client side, so i would go with that. I don't think it is really important at the end, no one will ever deal with it

daemon/src/service.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub async fn run(self) -> Result<()> {
+        let connection = zbus::connection::Builder::session()?
+            .name("org.freedesktop.secrets.Devel")?
+            .build()
+            .await?;
+        connection
+            .object_server()
+            .at(oo7::dbus::api::Service::PATH.unwrap(), self)
+            .await?;
+        Ok(())
```
bilelmoussaoui, line 92:
```suggestion
    pub async fn run() -> Result<()> {
        let connection = zbus::connection::Builder::session()?
            .name("org.freedesktop.secrets.Devel")?
            .build()
            .await?;
        connection
            .object_server()
            .at(oo7::dbus::api::Service::PATH.unwrap(), Self::default())
            .await?;
        Ok(())
```

and then just call Service::run().await?; instead.

daemon/src/service.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    pub async fn set_alias(&self, _name: &str, _collection: ObjectPath<'_>) -> Result<()> {
+        todo!()
+    }
+}
+
+impl Service {
+    pub async fn run(self) -> Result<()> {
+        let connection = zbus::connection::Builder::session()?
+            .name("org.freedesktop.secrets.Devel")?
```
bilelmoussaoui, line 85: why is this name hardcoded? it is accessible from oo7::dbus::api like the path. If it is for making debugging easier, we would need a proper infrastructure for it, which can be added in a separate commit.

daemon/src/service.rs
warusadura: done, thanks!

daemon/src/service.rs
warusadura: yeah, done.

warusadura:
included a README file. I'll add license information later. I'm suppose to follow [this](https://wiki.gnome.org/Foundation/OpenSourceGuidelines).

bilelmoussaoui:
nice, thanks!

PR #138, opened 2024-10-15
client: expose crypto::decrypt() and Key::new()

warusadura:
Needed for the server side implementation.


client/src/crypto/native.rs
```diff
@@ -38,11 +38,7 @@ pub fn encrypt(data: impl AsRef<[u8]>, key: &Key, iv: impl AsRef<[u8]>) -> Vec<u
     blob
 }
 
-pub(crate) fn decrypt(
```
bilelmoussaoui, line 41: Needs the same change for openssl

warusadura: done, thanks!

PR #139, opened 2024-10-17
server: Implement service method OpenSession

warusadura:
This change implements the service method OpenSession. Also, completes the Secret Session interface implementation. And adds ServiceManager to keep track of Sessions.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -2,6 +2,7 @@ mod collection;
 mod item;
 mod prompt;
 mod service;
+mod service_manager;
 mod session;
 
 use service::{Result, Service};
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -1,30 +1,76 @@
 // org.freedesktop.Secret.Service
 
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
-use oo7::dbus::{
-    api::{Properties, SecretInner},
-    Algorithm, ServiceError,
+use oo7::{
+    dbus::{
+        api::{Properties, SecretInner},
+        Algorithm, ServiceError,
+    },
+    Key,
 };
+use tokio::sync::{Mutex, RwLock};
 use zbus::{
     proxy::ProxyDefault,
     zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Value},
 };
 
+use crate::{service_manager::ServiceManager, session::Session};
+
 pub type Result<T> = std::result::Result<T, ServiceError>;
 
 #[derive(Debug, Default)]
-pub struct Service {}
+pub struct Service {
+    manager: Arc<Mutex<ServiceManager>>,
+    session_index: RwLock<i32>,
+}
 
 #[zbus::interface(name = "org.freedesktop.Secret.Service")]
 impl Service {
     #[zbus(out_args("output", "result"))]
     pub async fn open_session(
         &self,
-        _algorithm: Algorithm,
-        _input: Value<'_>,
+        algorithm: Algorithm,
+        input: Value<'_>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(OwnedValue, OwnedObjectPath)> {
-        todo!()
+        let (public_key, aes_key) = match algorithm {
+            Algorithm::Plain => (None, None),
+            Algorithm::Encrypted => {
+                let client_public_key = Key::from(input);
+                let private_key = Key::generate_private_key();
+                (
+                    Some(Key::generate_public_key(&private_key)),
+                    Some(Key::generate_aes_key(&private_key, &client_public_key)),
+                )
+            }
+        };
+
+        let mut session_index = *self.session_index.read().await;
+        session_index += 1;
+        let session = Session::new(
+            aes_key.map(Arc::new),
+            Arc::clone(&self.manager),
+            session_index,
+        );
+        *self.session_index.write().await = session_index;
+        let path = session.path().clone();
+
+        {
+            let session = Arc::new(session.clone());
+            self.manager
+                .lock()
+                .await
+                .insert_session(path.clone(), Arc::clone(&session));
+        }
+
+        object_server.at(&path, session).await?;
+
+        let service_key = public_key
+            .map(|k| OwnedValue::from(&k))
+            .unwrap_or_else(|| Value::new::<Vec<u8>>(vec![]).try_to_owned().unwrap());
+
+        Ok((service_key, path))
     }
 
     #[zbus(out_args("collection", "prompt"))]
--- a/server/src/service_manager.rs
+++ b/server/src/service_manager.rs
@@ -1,0 +1,23 @@
+// Secret Service Manager
+
+use std::{collections::HashMap, sync::Arc};
+
+use zbus::zvariant::OwnedObjectPath;
+
+use crate::session::Session;
+
+#[derive(Debug, Default)]
+pub struct ServiceManager {
+    // sessions mapped to their corresponding object path on the bus
+    sessions: HashMap<OwnedObjectPath, Arc<Session>>,
+}
+
+impl ServiceManager {
+    pub fn insert_session(&mut self, path: OwnedObjectPath, session: Arc<Session>) {
+        self.sessions.insert(path, session);
+    }
+
+    pub fn remove_session(&mut self, path: &OwnedObjectPath) {
+        self.sessions.remove(path);
+    }
+}
--- a/server/src/session.rs
+++ b/server/src/session.rs
@@ -1,15 +1,45 @@
 // org.freedesktop.Secret.Session
 
-use zbus::interface;
+use std::sync::Arc;
+
+use oo7::Key;
+use tokio::sync::Mutex;
+use zbus::{interface, zvariant::OwnedObjectPath};
 
 use super::Result;
+use crate::service_manager::ServiceManager;
 
-#[derive(Debug)]
-pub struct Session {}
+#[derive(Debug, Clone)]
+pub struct Session {
+    _aes_key: Option<Arc<Key>>,
+    manager: Arc<Mutex<ServiceManager>>,
+    path: OwnedObjectPath,
+}
 
 #[interface(name = "org.freedesktop.Secret.Session")]
 impl Session {
-    pub async fn close(&self) -> Result<()> {
-        todo!()
+    pub async fn close(
+        &self,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<()> {
+        self.manager.lock().await.remove_session(&self.path);
+        object_server.remove::<Self, _>(&self.path).await?;
+
+        Ok(())
+    }
+}
+
+impl Session {
+    pub fn new(aes_key: Option<Arc<Key>>, manager: Arc<Mutex<ServiceManager>>, index: i32) -> Self {
+        Self {
+            path: OwnedObjectPath::try_from(format!("/org/freedesktop/secrets/session/s{index}"))
+                .unwrap(),
+            _aes_key: aes_key,
+            manager,
+        }
+    }
+
+    pub fn path(&self) -> &OwnedObjectPath {
+        &self.path
     }
 }


server/src/main.rs
```diff
@@ -1,7 +1,9 @@
 mod collection;
+pub mod consts;
```
bilelmoussaoui, line 2: doesn't have to be pub?

server/src/service_manager.rs
```diff
@@ -0,0 +1,26 @@
...
+};
+
+use zbus::zvariant::OwnedObjectPath;
+
+use crate::session::Session;
+
+#[derive(Debug, Default)]
+pub struct ServiceManager {
+    // store Secret Sessions mapped to ObjectPaths
+    sessions: Mutex<HashMap<OwnedObjectPath, Arc<Session>>>,
```
bilelmoussaoui, line 15: The service manager itself is already stored behind a Mutex, you don't need an extra one here.

server/src/service_manager.rs
```diff
@@ -0,0 +1,26 @@
...
+}
+
+impl ServiceManager {
+    pub fn insert_session(&self, path: OwnedObjectPath, session: Arc<Session>) {
+        self.sessions.lock().unwrap().insert(path, session);
+    }
+
+    pub fn remove_session(&self, path: OwnedObjectPath) {
```
bilelmoussaoui, line 23: You need a ref, take a ref.

server/src/session.rs
```diff
@@ -1,15 +1,51 @@
...
+
+        Ok(())
+    }
+}
+
+impl Session {
+    pub fn new(
+        aes_key: Option<Arc<Key>>,
+        manager: Arc<Mutex<ServiceManager>>,
+        session_counter: i32,
```
bilelmoussaoui, line 35: You already take the manager, so why require also the counter? increment it here instead.

server/src/session.rs
```diff
@@ -1,15 +1,51 @@
...
+                "{}s{}",
+                SECRET_SESSION_PREFIX, session_counter
+            ))
+            .unwrap(),
+            _aes_key: aes_key,
+            manager,
+        }
+    }
+
+    pub fn path(&self) -> OwnedObjectPath {
```
bilelmoussaoui, line 48: Don't clone here, just return a ref, clone where needed instead.

server/src/service.rs
```diff
@@ -1,30 +1,77 @@
...
+                (
+                    Some(Key::generate_public_key(&private_key)),
+                    Some(Key::generate_aes_key(&private_key, &client_public_key)),
+                )
+            }
+        };
+
+        *self.session_counter.write().await += 1;
+        let session = Session::new(
+            Some(Arc::new(aes_key.unwrap())),
```
bilelmoussaoui, line 54: A plain connection will crash here. You should instead do aes_key.map(Arc::new)

server/src/service.rs
```diff
@@ -1,30 +1,77 @@
...
+            Arc::clone(&self.manager),
+            *self.session_counter.read().await,
```
bilelmoussaoui, line 56: You pass the manager already

server/src/service.rs
```diff
@@ -1,30 +1,77 @@
...
+                .lock()
+                .unwrap()
+                .insert_session(path.clone(), Arc::clone(&session));
+        }
+
+        object_server.at(&path, session).await?;
+
+        let service_key = public_key
+            .map(|k| OwnedValue::from(&k))
+            .unwrap_or_else(|| Value::new::<Vec<u8>>(vec![]).try_to_owned().unwrap());
```
bilelmoussaoui, line 72: This looks weird.

server/src/consts.rs
```diff
@@ -0,0 +1,3 @@
+// constants
+
+pub const SECRET_SESSION_PREFIX: &str = "/org/freedesktop/secrets/session/";
```
bilelmoussaoui, line 3: No need for this file at all. Just hardcode the value where it is used for now

server/src/service_manager.rs
```diff
@@ -0,0 +1,26 @@
...
+    sync::{Arc, Mutex},
+};
+
+use zbus::zvariant::OwnedObjectPath;
+
+use crate::session::Session;
+
+#[derive(Debug, Default)]
+pub struct ServiceManager {
+    // store Secret Sessions mapped to ObjectPaths
```
bilelmoussaoui, line 14:
```suggestion
    // sessions mapped to their corresponding object path on the bus
```

bilelmoussaoui:
The commit should be "server/service: Implement OpenSession"

server/src/service_manager.rs
warusadura: understood, done.

server/src/session.rs
warusadura: done, thanks!

server/src/service.rs
warusadura: good catch! fixed this now.

server/src/consts.rs
warusadura: removed the file.

server/src/service_manager.rs
warusadura: done.

server/src/service_manager.rs
warusadura: done, thanks!

server/src/session.rs
warusadura: I don't get it. The `session_counter` attribute is in the `Service` struct and not in the `ServiceManager`. Do you mean I should get the session count from `sessions` HashMap?

server/src/service.rs
warusadura: I had to try many times to get this right :) and I'm not sure how to make this better.

server/src/session.rs
bilelmoussaoui: My bad, I read too fast. Then rename session_counter to session_index.

server/src/service.rs
bilelmoussaoui: I would suggest reading from the mutex first, storing that in a variable, then increment it, and store it afterwards.

server/src/service.rs
bilelmoussaoui: That is okay, I will look at it after we merge this

server/src/session.rs
```diff
@@ -1,15 +1,53 @@
...
+
+impl Session {
+    pub fn new(
+        aes_key: Option<Arc<Key>>,
+        manager: Arc<Mutex<ServiceManager>>,
+        session_counter: i32,
+    ) -> Self {
+        Self {
+            path: OwnedObjectPath::try_from(format!(
+                "{}s{}",
```
bilelmoussaoui, line 41: Just hardcode the value here

warusadura: Hard coded the session path.

server/src/service.rs
warusadura: done, thanks!

server/src/session.rs
warusadura: no worries, done :)

server/src/service.rs
```diff
@@ -1,30 +1,76 @@
...
+                let client_public_key = Key::from(input);
+                let private_key = Key::generate_private_key();
+                (
+                    Some(Key::generate_public_key(&private_key)),
+                    Some(Key::generate_aes_key(&private_key, &client_public_key)),
+                )
+            }
+        };
+
+        let mut sessions = *self.session_index.read().await;
```
bilelmoussaoui, line 52: Rename the variable to session_index as well

server/src/service.rs
```diff
@@ -1,30 +1,76 @@
...
+        sessions += 1;
+        *self.session_index.write().await = sessions;
```
bilelmoussaoui, line 54: Write the new value after usage. Also I am curious why you use RwLock from Tokio and Mutex from std.

server/src/service.rs
```diff
@@ -1,30 +1,76 @@
...
+
+        let session = Session::new(aes_key.map(Arc::new), Arc::clone(&self.manager), sessions);
+        let path = session.path().clone();
```
bilelmoussaoui, line 57: Why clone here?

server/src/service.rs
```diff
@@ -1,30 +1,76 @@
...
+                .insert_session(path.clone(), Arc::clone(&session));
+        }
+
+        object_server.at(&path, session).await?;
+
+        let service_key = public_key
+            .map(|k| OwnedValue::from(&k))
+            .unwrap_or_else(|| Value::new::<Vec<u8>>(vec![]).try_to_owned().unwrap());
+
+        Ok((service_key, path))
```
bilelmoussaoui, line 73: Instead of here

server/src/service.rs
warusadura: >  Also I am curious why you use RwLock from Tokio and Mutex from std.

I read a bunch of reddit posts explaining what's better and decided to go with their findings. I thought to myself you'll catch this if this is wrong and I'm glad that you did :)

What should I go with? std or tokio?

server/src/service.rs
warusadura: Yeah I agree. It doesn't make sense to clone here (but below). But when removed, I'm getting a lengthy error,

```
error[E0277]: the trait bound `zbus::zvariant::ObjectPath<'_>: From<&&OwnedObjectPath>` is not satisfied
  --> server/src/service.rs:70:23
   |
70 |         object_server.at(&path, session).await?;
   |                       ^^ the trait `From<&&OwnedObjectPath>` is not implemented for `zbus::zvariant::ObjectPath<'_>`, which is required by `zbus::zvariant::ObjectPath<'_>: TryFrom<&&OwnedObjectPath>`
   |
   = help: the following other types implement trait `From<T>`:
             `zbus::zvariant::ObjectPath<'o>` implements `From<&zbus::zvariant::ObjectPath<'o>>`
             `zbus::zvariant::ObjectPath<'static>` implements `From<OwnedObjectPath>`
             `zbus::zvariant::ObjectPath<'unowned>` implements `From<&'owned OwnedObjectPath>`
   = note: required for `&&OwnedObjectPath` to implement `Into<zbus::zvariant::ObjectPath<'_>>`
   = note: required for `zbus::zvariant::ObjectPath<'_>` to implement `TryFrom<&&OwnedObjectPath>`

error[E0277]: the trait bound `zbus::zvariant::ObjectPath<'_>: From<&&OwnedObjectPath>` is not satisfied
  --> server/src/service.rs:70:9
   |
70 |         object_server.at(&path, session).await?;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<&&OwnedObjectPath>` is not implemented for `zbus::zvariant::ObjectPath<'_>`, which is required by `zbus::zvariant::ObjectPath<'_>: TryFrom<&&OwnedObjectPath>`
   |
   = help: the following other types implement trait `From<T>`:
             `zbus::zvariant::ObjectPath<'o>` implements `From<&zbus::zvariant::ObjectPath<'o>>`
             `zbus::zvariant::ObjectPath<'static>` implements `From<OwnedObjectPath>`
             `zbus::zvariant::ObjectPath<'unowned>` implements `From<&'owned OwnedObjectPath>`
   = note: required for `&&OwnedObjectPath` to implement `Into<zbus::zvariant::ObjectPath<'_>>`
   = note: required for `zbus::zvariant::ObjectPath<'_>` to implement `TryFrom<&&OwnedObjectPath>`

error[E0277]: the trait bound `zbus::zvariant::ObjectPath<'_>: From<&&OwnedObjectPath>` is not satisfied
  --> server/src/service.rs:70:42
   |
70 |         object_server.at(&path, session).await?;
   |                                          ^^^^^ the trait `From<&&OwnedObjectPath>` is not implemented for `zbus::zvariant::ObjectPath<'_>`, which is required by `zbus::zvariant::ObjectPath<'_>: TryFrom<&&OwnedObjectPath>`
   |
   = help: the following other types implement trait `From<T>`:
             `zbus::zvariant::ObjectPath<'o>` implements `From<&zbus::zvariant::ObjectPath<'o>>`
             `zbus::zvariant::ObjectPath<'static>` implements `From<OwnedObjectPath>`
             `zbus::zvariant::ObjectPath<'unowned>` implements `From<&'owned OwnedObjectPath>`
   = note: required for `&&OwnedObjectPath` to implement `Into<zbus::zvariant::ObjectPath<'_>>`
   = note: required for `zbus::zvariant::ObjectPath<'_>` to implement `TryFrom<&&OwnedObjectPath>`

For more information about this error, try `rustc --explain E0277`.

```

warusadura: ah! figured out the error.

warusadura: so, `object_server.at(path.clone(), session).await?;` call moves `session`, that's why I have cloned the path here.

server/src/service.rs
warusadura: done.

server/src/service.rs
bilelmoussaoui: I would have to read about it to give you a concrete answer, but just use whatever we use in the client so it is consistent

server/src/session.rs
```diff
@@ -1,15 +1,44 @@
...
+    pub fn new(aes_key: Option<Arc<Key>>, manager: Arc<Mutex<ServiceManager>>, index: i32) -> Self {
+        Self {
+            path: OwnedObjectPath::try_from(format!("/org/freedesktop/secrets/session/s{index}"))
+                .unwrap(),
+            _aes_key: aes_key,
+            manager,
+        }
+    }
+
+    pub fn path(&self) -> &OwnedObjectPath {
```
A6GibKm, line 41: Does this need to be public?

A6GibKm: Can it be a zvariant::ObjectPath instead?

bilelmoussaoui: it is a binary, so it doesn't matter for the public part. And it can actually return an ObjectPath i think

server/src/service.rs
warusadura: Now both Mutex, RwLock are from tokio. I think this is ok for now. I'll continue to read on this and will update it if necessary.

PR #140, opened 2024-10-18
server: Store connection to the session bus in Service

warusadura:
This is to avoid the connection to the session bus being dropped.

A bit of background: I've been trying to get a response to `busctl --user tree org.freedesktop.secrets` command from oo7-daemon (after killing gnome-keyring-daemon). As soon as I issue the command, `gnome-keyring-daemon` starts running again and becomes the secret service provider. I'm not sure why this happens. But this change fixes this issue and the object tree is now nicely generated.

```
└─ /org
  └─ /org/freedesktop
    └─ /org/freedesktop/secrets
```

### DIFF
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -13,15 +13,17 @@ use tokio::sync::{Mutex, RwLock};
 use zbus::{
     proxy::Defaults,
     zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Value},
+    Connection,
 };
 
 use crate::{service_manager::ServiceManager, session::Session};
 
 pub type Result<T> = std::result::Result<T, ServiceError>;
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct Service {
     manager: Arc<Mutex<ServiceManager>>,
+    connection: Connection,
     session_index: RwLock<i32>,
 }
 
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
This doesn't look like a fix but rather a change that makes it work somehow. I think the problem is the connection being dropped, so you just have to keep a clone of it in the Service type. Also connection is using Arc internally, so you don't need to wrap it in Arc as well

bilelmoussaoui:
Needs a rebase.

warusadura:
> Needs a rebase.

Done.

server/src/service.rs
```diff
@@ -126,18 +128,29 @@ impl Service {
...
+        let object_server = connection.object_server();
+
+        object_server
+            .at(oo7::dbus::api::Service::PATH.as_deref().unwrap(), self)
             .await?;
+
         Ok(())
     }
+
+    pub fn connection(&self) -> Connection {
```
A6GibKm, line 153: Please either make it so it returns a reference to `Connection` or remove the method and use `let connection = self.conection` in the `run` method. This clone here is a bit unnecesary.

warusadura: Removed the getter method, thanks!

bilelmoussaoui:
It still doesn't fix what I mentioned above though :) I pushed some commits to main that fix it & a quick cleanup.

warusadura:
thanks :)

PR #141, opened 2024-10-20
server: Dispatch temporary session collection to the object tree

warusadura:
This change creates and dispatches the temporary collection called session.
This collection is never stored to disk, but forgotten when the session ends.

`org.freedesktop.secrets` object tree:

```
└─ /org
  └─ /org/freedesktop
    └─ /org/freedesktop/secrets
      └─ /org/freedesktop/secrets/collection
        └─ /org/freedesktop/secrets/collection/session
```

### DIFF
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -1,15 +1,34 @@
 // org.freedesktop.Secret.Collection
 
-use std::collections::HashMap;
+use std::{
+    collections::HashMap,
+    sync::{atomic::AtomicBool, Arc},
+    time::{Duration, SystemTime},
+};
 
 use oo7::dbus::api::{Properties, SecretInner};
+use tokio::sync::{Mutex, RwLock};
 use zbus::{interface, zvariant};
 use zvariant::{ObjectPath, OwnedObjectPath};
 
 use super::Result;
+use crate::{item, service_manager::ServiceManager};
 
 #[derive(Debug)]
-pub struct Collection {}
+#[allow(unused)]
+pub struct Collection {
+    // Properties
+    items: Mutex<Vec<item::Item>>,
+    label: Mutex<String>,
+    locked: AtomicBool,
+    created: Duration,
+    modified: Mutex<Duration>,
+    // Other attributes
+    alias: Mutex<String>,
+    manager: Arc<Mutex<ServiceManager>>,
+    n_items: RwLock<i32>,
+    path: OwnedObjectPath,
+}
 
 #[interface(name = "org.freedesktop.Secret.Collection")]
 impl Collection {
@@ -31,5 +50,33 @@ impl Collection {
         _replace: bool,
     ) -> Result<(OwnedObjectPath, ObjectPath)> {
         todo!()
+    }
+}
+
+impl Collection {
+    pub fn new(label: &str, alias: &str, manager: Arc<Mutex<ServiceManager>>) -> Self {
+        let created = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap();
+
+        Self {
+            items: Default::default(),
+            label: Mutex::new(label.to_owned()),
+            locked: AtomicBool::new(true),
+            modified: Mutex::new(created),
+            alias: Mutex::new(alias.to_owned()),
+            n_items: RwLock::new(0),
+            path: OwnedObjectPath::try_from(format!(
+                "/org/freedesktop/secrets/collection/{}",
+                label
+            ))
+            .unwrap(),
+            created,
+            manager,
+        }
+    }
+
+    pub fn path(&self) -> &OwnedObjectPath {
+        &self.path
     }
 }
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -15,12 +15,15 @@ use zbus::{
     zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Value},
 };
 
-use crate::{service_manager::ServiceManager, session::Session};
+use crate::{collection::Collection, service_manager::ServiceManager, session::Session};
 
 pub type Result<T> = std::result::Result<T, ServiceError>;
 
 #[derive(Debug)]
 pub struct Service {
+    // Properties
+    collections: Mutex<Vec<OwnedObjectPath>>,
+    // Other attributes
     manager: Arc<Mutex<ServiceManager>>,
     #[allow(unused)]
     connection: zbus::Connection,
@@ -125,15 +128,41 @@ impl Service {
             .name(oo7::dbus::api::Service::DESTINATION.as_deref().unwrap())?
             .build()
             .await?;
+        let object_server = connection.object_server();
         let service = Self {
+            collections: Default::default(),
             manager: Default::default(),
             connection: connection.clone(),
         };
+        let collections = service.fetch_collections().await;
 
-        connection
-            .object_server()
+        object_server
             .at(oo7::dbus::api::Service::PATH.as_deref().unwrap(), service)
             .await?;
+
+        for collection in collections {
+            object_server
+                .at(collection.path().clone(), collection)
+                .await?;
+        }
+
         Ok(())
+    }
+
+    async fn fetch_collections(&self) -> Vec<Collection> {
+        let mut collections = Vec::new();
+        // todo: create default collection
+
+        // create temporary session collection
+        let session_collection = Collection::new("session", "session", Arc::clone(&self.manager));
+        collections.push(session_collection);
+
+        let mut lock = self.collections.lock().await;
+        for collection in &collections {
+            lock.push(collection.path().clone());
+        }
+        drop(lock);
+
+        collections
     }
 }


server/src/collection.rs
```diff
@@ -33,3 +52,31 @@ impl Collection {
         todo!()
     }
 }
+
+impl Collection {
+    pub fn new(label: String, alias: String, manager: Arc<Mutex<ServiceManager>>) -> Self {
```
bilelmoussaoui, line 57:
```suggestion
    pub fn new(label: &str, alias: &str, manager: Arc<Mutex<ServiceManager>>) -> Self {
```
always take a ref, and clone only where needed.

server/src/collection.rs
```diff
@@ -33,3 +52,31 @@ impl Collection {
...
+        let created = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap();
+
+        Self {
+            items: Default::default(),
+            label: Mutex::new(label.clone()),
+            locked: AtomicBool::new(true),
+            modified: created,
```
bilelmoussaoui, line 66: I suppose this is something that is mutable?

server/src/collection.rs
```diff
@@ -33,3 +52,31 @@ impl Collection {
...
+            alias: Mutex::new(alias),
+            n_items: RwLock::new(0),
```
bilelmoussaoui, line 68: Like n_sesssions, you have the list of the items. You can get the n_items...

server/src/collection.rs
```diff
@@ -1,15 +1,34 @@
...
 use zvariant::{ObjectPath, OwnedObjectPath};
 
 use super::Result;
+use crate::{item, service_manager::ServiceManager};
 
 #[derive(Debug)]
-pub struct Collection {}
+#[allow(unused)] // todo: remove this after completing the ifce impl
+pub struct Collection {
+    // Secret Collection properties
```
bilelmoussaoui, line 20:
```suggestion
    // Properties
```

server/src/collection.rs
```diff
@@ -1,15 +1,34 @@
...
+    items: Mutex<Vec<item::Item>>,
+    label: Mutex<String>,
+    locked: AtomicBool,
+    created: Duration,
+    modified: Duration,
+    // other attributes
```
bilelmoussaoui, line 26:
```suggestion
    // Other attributes
```

server/src/collection.rs
```diff
@@ -1,15 +1,34 @@
...
+use tokio::sync::{Mutex, RwLock};
 use zbus::{interface, zvariant};
 use zvariant::{ObjectPath, OwnedObjectPath};
 
 use super::Result;
+use crate::{item, service_manager::ServiceManager};
 
 #[derive(Debug)]
-pub struct Collection {}
+#[allow(unused)] // todo: remove this after completing the ifce impl
```
bilelmoussaoui, line 18:
```suggestion
#[allow(unused)]
```

No need to keep this comment, we will go through things before we are fully done.

server/src/service.rs
```diff
@@ -15,12 +15,15 @@ use zbus::{
...
 
-use crate::{service_manager::ServiceManager, session::Session};
+use crate::{collection::Collection, service_manager::ServiceManager, session::Session};
 
 pub type Result<T> = std::result::Result<T, ServiceError>;
 
 #[derive(Debug)]
 pub struct Service {
+    // Secret Service properties
+    collections: Mutex<Vec<OwnedObjectPath>>,
```
bilelmoussaoui, line 25: Is this correct? How are you going to get the collection again from it path?

server/src/service.rs
```diff
@@ -125,15 +128,46 @@ impl Service {
             .name(oo7::dbus::api::Service::DESTINATION.as_deref().unwrap())?
             .build()
             .await?;
+        let object_server = connection.object_server();
         let service = Self {
+            collections: Mutex::new(Vec::new()),
```
bilelmoussaoui, line 133:
```suggestion
            collections: Default::default(),
```

server/src/service.rs
```diff
@@ -125,15 +128,46 @@ impl Service {
...
             manager: Default::default(),
             connection: connection.clone(),
         };
+        let collections = service.init_collections().await;
```
bilelmoussaoui, line 137:
```suggestion
        let collections = service.fetch_collections().await;
```

server/src/service.rs
```diff
@@ -125,15 +128,46 @@ impl Service {
...
+            object_server
+                .at(collection.path().clone(), collection)
+                .await?;
+        }
+
         Ok(())
     }
+
+    async fn init_collections(&self) -> Vec<Collection> {
+        let mut collections: Vec<Collection> = Vec::new();
```
bilelmoussaoui, line 153:
```suggestion
        let mut collections = Vec::new();
```

server/src/service.rs
```diff
@@ -125,15 +128,46 @@ impl Service {
...
+        // todo: create default collection
+
+        // create temporary session collection
+        let session = Collection::new(
```
bilelmoussaoui, line 157: this is not a session

server/src/service.rs
```diff
@@ -125,15 +128,46 @@ impl Service {
...
+            "session".to_string(),
+            "session".to_string(),
+            Arc::clone(&self.manager),
+        );
+        collections.push(session);
+
+        for collection in &collections {
+            self.collections
```
bilelmoussaoui, line 165: Lock once before for instead of doing that for every iteration...

server/src/service.rs
bilelmoussaoui, line 157: How do you identify this as a collection that never gets stored on disk? nothing tells so.

server/src/service.rs
```diff
@@ -125,15 +128,46 @@ impl Service {
...
 
-        connection
-            .object_server()
+        object_server
             .at(oo7::dbus::api::Service::PATH.as_deref().unwrap(), service)
             .await?;
+
+        for collection in collections {
+            object_server
+                .at(collection.path().clone(), collection)
```
bilelmoussaoui, line 145:
```suggestion
                .at(collection.path(), collection)
```
doesn't work??

server/src/service.rs
```diff
@@ -15,12 +15,15 @@ use zbus::{
...
 };
 
-use crate::{service_manager::ServiceManager, session::Session};
+use crate::{collection::Collection, service_manager::ServiceManager, session::Session};
 
 pub type Result<T> = std::result::Result<T, ServiceError>;
 
 #[derive(Debug)]
 pub struct Service {
+    // Secret Service properties
```
bilelmoussaoui, line 24:
```suggestion
    // Properties
```

server/src/service.rs
```diff
@@ -15,12 +15,15 @@ use zbus::{
...
+    collections: Mutex<Vec<OwnedObjectPath>>,
+    // other attributes
```
bilelmoussaoui, line 26:
```suggestion
    // Other attributes
```

server/src/collection.rs
bilelmoussaoui: None of those are exposed as properties though... I suppose would be done later?

server/src/collection.rs
warusadura: got it, thanks!

server/src/collection.rs
warusadura: yes :)

server/src/collection.rs
warusadura: yeah, removed.

server/src/service.rs
warusadura: renamed to `session_collection`

server/src/service.rs
warusadura: understood.

server/src/service.rs
warusadura: yes, it doesn't work.

server/src/service.rs
warusadura: The followings are the only two places mentioned about this that I could find.
[1] https://github.com/GNOME/gnome-keyring/blob/4132075144c7bb21b897570dd53b005ac38250aa/README#L6
[2] https://github.com/GNOME/libsecret/blob/e39b831776a932bd195571c1fa7d555702fd9284/libsecret/secret-collection.c#L81

server/src/collection.rs
warusadura: `/org/freedesktop/secrets/collection/login/iiii`

Yeah. The only thing that I don't like about this approach (counting the items to set the last part of an objectpath) is `iiii` should be unique/specific. With this approach there is always a chance (when an item is deleted and then another created, the object tree will be the same) that two items has the same `iiii` and so the the same objectpath. 

Reference: https://specifications.freedesktop.org/secret-service-spec/latest/object-paths.html

server/src/service.rs
bilelmoussaoui: What I meant is that the code you wrote doesn't do anything for it to potentiallly work.  It needs work on the client::portal first as it doesn't have a way to create a new instance that never writes to disk.

server/src/service.rs
warusadura: > Is this correct?

Yes. I'm certain for single collection/keyring. But, we may need to modify this when adding multiple keyring support. I did this to avoid many instances of cloning.

> How are you going to get the collection again from it path?

I can access all the methods and attributes following [this](https://docs.rs/zbus/latest/zbus/object_server/struct.ObjectServer.html#examples) approach.

server/src/collection.rs
bilelmoussaoui: Then this should at least use Default::default then.

bilelmoussaoui: Well, the spec says auto-generated not necessarily the counter. So maybe we should use some uuid or random string.

server/src/service.rs
warusadura: Ah yes I get it, thanks! But we still need this. Because, seahorse app won't work without this (not sure about Keyrack). During the launch of seahorse, it tries to retrieve the objectpath of session alias.

server/src/collection.rs
warusadura: Yeah, but at least for now, I'd like to follow the gnome-keyring approach. It's simple and we can have a clearer object tree with this approach :)

server/src/service.rs
bilelmoussaoui: I have pushed a commit that adds support for tmp keyring through  `oo7::portal::api::Keyring::temporary()`

warusadura: Very nice \o/, thanks!

PR #146, opened 2024-10-24
server: Add ReadAlias and SetAlias implementation

warusadura:
Implemented according to the gnome-keyring-daemon. This means, only the 'default' alias is supported.

### DIFF
--- a/client/src/dbus/error.rs
+++ b/client/src/dbus/error.rs
@@ -13,7 +13,7 @@ pub enum ServiceError {
     /// Session does not exist.
     NoSession,
     /// Collection/Item does not exist.
-    NoSuchObject,
+    NoSuchObject(String),
 }
 
 /// DBus backend specific errors.
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -95,4 +95,16 @@ impl Collection {
     pub fn path(&self) -> &OwnedObjectPath {
         &self.path
     }
+
+    pub async fn label(&self) -> String {
+        self.label.lock().await.clone()
+    }
+
+    pub async fn set_alias(&self, alias: String) {
+        *self.alias.lock().await = alias;
+    }
+
+    pub async fn alias(&self) -> String {
+        self.alias.lock().await.clone()
+    }
 }
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -114,16 +114,61 @@ impl Service {
     }
 
     #[zbus(out_args("collection"))]
-    pub async fn read_alias(&self, _name: &str) -> Result<ObjectPath, ServiceError> {
-        todo!()
+    pub async fn read_alias(
+        &self,
+        name: &str,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<ObjectPath, ServiceError> {
+        let mut objectpath = ObjectPath::default();
+        for collection in self.collections.lock().await.iter() {
+            let collection_ifce_ref = object_server.interface::<_, Collection>(collection).await?;
+            let collection = collection_ifce_ref.get_mut().await;
+
+            if collection.alias().await == name || collection.label().await == name {
+                objectpath = ObjectPath::from(collection.path().clone());
+            }
+        }
+
+        Ok(objectpath)
     }
 
     pub async fn set_alias(
         &self,
-        _name: &str,
-        _collection: ObjectPath<'_>,
+        name: &str,
+        collection: ObjectPath<'_>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(), ServiceError> {
-        todo!()
+        if name != "default" {
+            return Err(ServiceError::ZBus(zbus::Error::Failure(String::from(
+                "Only the 'default' alias is supported",
+            ))));
+        }
+
+        let collections = self.collections.lock().await;
+
+        if !collections
+            .to_vec()
+            .contains(&OwnedObjectPath::from(collection.clone()))
+        {
+            return Err(ServiceError::NoSuchObject(String::from(
+                "The collection does not exist",
+            )));
+        }
+
+        for path in collections.iter() {
+            if **path == collection {
+                let collection_ifce_ref = object_server.interface::<_, Collection>(path).await?;
+                let collection = collection_ifce_ref.get_mut().await;
+
+                collection.set_alias(String::from(name)).await;
+            }
+        }
+
+        tracing::info!("Collection: {} alias updated to {}.", collection, name);
+
+        drop(collections);
+
+        Ok(())
     }
 }
 


server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
+        name: &str,
+        collection: ObjectPath<'_>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(), ServiceError> {
-        todo!()
+        if name != "default" {
+            return Err(ServiceError::ZBus(zbus::Error::Failure(String::from(
+                "Only the 'default' alias is supported",
+            ))));
+        }
```
bilelmoussaoui, line 145: I don't think it makes sense to re-add the same restrictions as gnome-keyring, the goal is to be better than gnome-keyring.

server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
+
+        let collections = self.collections.lock().await;
+
+        if !collections
+            .to_vec()
+            .contains(&OwnedObjectPath::from(collection.clone()))
```
bilelmoussaoui, line 151: Instead, change the method to take a OwnedObjectPath, let zbus handle the rest

server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
+        {
+            return Err(ServiceError::NoSuchObject(String::from(
+                "The collection does not exist",
+            )));
```
bilelmoussaoui, line 155: Does this error take a string?

server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
+        }
+
+        for path in collections.iter() {
+            if **path == collection {
+                let collection_ifce_ref = object_server.interface::<_, Collection>(path).await?;
+                let collection = collection_ifce_ref.get_mut().await;
+
+                collection.set_alias(String::from(name)).await;
```
bilelmoussaoui, line 163: Make the function take a ref, and clone where needed. Also add a break statement, no need to keep iterating

server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
+        &self,
+        name: &str,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<ObjectPath, ServiceError> {
+        let mut objectpath = ObjectPath::default();
+        for collection in self.collections.lock().await.iter() {
+            let collection_ifce_ref = object_server.interface::<_, Collection>(collection).await?;
+            let collection = collection_ifce_ref.get_mut().await;
+
+            if collection.alias().await == name || collection.label().await == name {
```
bilelmoussaoui, line 127: Why does it check the label as well?

server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
+                objectpath = ObjectPath::from(collection.path().clone());
```
bilelmoussaoui, line 128:
```suggestion
                return Ok(collection.path());
```

and `Ok(ObjectPath::default())` at the end while also logging which collection wasn't found

server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
     #[zbus(out_args("collection"))]
-    pub async fn read_alias(&self, _name: &str) -> Result<ObjectPath, ServiceError> {
-        todo!()
+    pub async fn read_alias(
+        &self,
+        name: &str,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<ObjectPath, ServiceError> {
+        let mut objectpath = ObjectPath::default();
+        for collection in self.collections.lock().await.iter() {
```
bilelmoussaoui, line 123: Store the collections in a separate variable

server/src/collection.rs
```diff
@@ -95,4 +95,16 @@ impl Collection {
     pub fn path(&self) -> &OwnedObjectPath {
         &self.path
     }
+
+    pub async fn label(&self) -> String {
+        self.label.lock().await.clone()
+    }
```
bilelmoussaoui, line 101: Not useful for this part, so remove it

server/src/collection.rs
```diff
@@ -95,4 +95,16 @@ impl Collection {
...
+
+    pub async fn set_alias(&self, alias: String) {
+        *self.alias.lock().await = alias;
```
bilelmoussaoui, line 104: Should emit a CollectionChanged as well

bilelmoussaoui: Doesn't changing the alias means renaming the keyring file as well?

server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
-        todo!()
+        if name != "default" {
+            return Err(ServiceError::ZBus(zbus::Error::Failure(String::from(
+                "Only the 'default' alias is supported",
+            ))));
+        }
+
+        let collections = self.collections.lock().await;
+
+        if !collections
```
bilelmoussaoui, line 149: This check here is useless, instead, if you iterate the collections, make it return Ok(() the moment you found the corresponding collection, you can return the error at the end of the function as you won't reach it unless the collection doesn't exists

server/src/service.rs
```diff
@@ -114,16 +114,61 @@ impl Service {
...
+        for path in collections.iter() {
+            if **path == collection {
+                let collection_ifce_ref = object_server.interface::<_, Collection>(path).await?;
+                let collection = collection_ifce_ref.get_mut().await;
+
+                collection.set_alias(String::from(name)).await;
+            }
+        }
+
+        tracing::info!("Collection: {} alias updated to {}.", collection, name);
```
bilelmoussaoui, line 167: This should be moved inside the if condition

client/src/dbus/error.rs
```diff
@@ -13,7 +13,7 @@ pub enum ServiceError {
     /// Session does not exist.
     NoSession,
     /// Collection/Item does not exist.
-    NoSuchObject,
+    NoSuchObject(String),
```
bilelmoussaoui, line 16: This change is wrong I believe. Can you point me to where in gnome-keyring, it passes extra details?

server/src/service.rs
warusadura: Yup, done. Thanks!

server/src/service.rs
warusadura: Done, thanks!

server/src/service.rs
warusadura: Done, thanks!

server/src/service.rs
warusadura: Yeah, I agree. Thanks!

server/src/service.rs
warusadura: Done.

client/src/dbus/error.rs
warusadura: https://gitlab.gnome.org/GNOME/gnome-keyring/-/blob/master/daemon/dbus/gkd-secret-service.c#L844

server/src/service.rs
warusadura: Done.

server/src/service.rs
warusadura: Yes.

server/src/service.rs
warusadura: Without this check I have no way of responding to the following call,
`busctl --user call org.freedesktop.secrets /org/freedesktop/secrets org.freedesktop.Secret.Service ReadAlias "s" login`

server/src/collection.rs
warusadura: > Should emit a CollectionChanged as well

No, it doesn't emit a CollectionChanged. At least gnome-keyring does not :)

warusadura: > Doesn't changing the alias means renaming the keyring file as well?

No, I don't think so. So, currently gnome-keyring names their keyring files based on the "label" property. And it doesn't rename them once we change the alias.

server/src/service.rs
warusadura: But then again this is only for the default/login collection. I think I need to put another filter here.

server/src/service.rs
warusadura: Yeah I agree. I think we should get to gnome-keyring level first, then we can add more better/cool stuff later :)

server/src/service.rs
warusadura: So, if the name is "login" we're comparing the label property. Otherwise, alias.

client/src/dbus/error.rs
bilelmoussaoui: Can you check the other errors and update them accordingly as well?

server/src/service.rs
```diff
@@ -114,16 +114,70 @@ impl Service {
...
+    ) -> Result<OwnedObjectPath, ServiceError> {
+        let collections = self.collections.lock().await;
+
+        for collection in collections.iter() {
+            let collection_ifce_ref = object_server.interface::<_, Collection>(collection).await?;
+            let collection = collection_ifce_ref.get_mut().await;
+
+            if name == "login" {
+                if collection.label().await == name {
+                    drop(collections);
```
bilelmoussaoui, line 130: No need to drop the mutex manually, it will be dropped at the end of the scope, this is not C.

server/src/service.rs
```diff
@@ -114,16 +114,70 @@ impl Service {
...
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<OwnedObjectPath, ServiceError> {
+        let collections = self.collections.lock().await;
+
+        for collection in collections.iter() {
+            let collection_ifce_ref = object_server.interface::<_, Collection>(collection).await?;
+            let collection = collection_ifce_ref.get_mut().await;
+
+            if name == "login" {
+                if collection.label().await == name {
```
bilelmoussaoui, line 129:
```suggestion
            if name == "login" && collection.label().await == name {
```

bilelmoussaoui: Also comment why is this needed, sounds very weird

server/src/service.rs
bilelmoussaoui: This condition doesn't affect anything anyways, so please drop it. As my goal is not to just be compatible with gnome-keyring but with spec and the spec says nothing about this restriction.

server/src/service.rs
```diff
@@ -114,16 +114,70 @@ impl Service {
...
+                let collection = collection_ifce_ref.get_mut().await;
+
+                collection.set_alias(name).await;
+
+                tracing::info!(
+                    "Collection: {} alias updated to {}.",
+                    collection.path(),
+                    name
+                );
+                drop(collections);
```
bilelmoussaoui, line 170: Same here, no need for the extra drops

server/src/service.rs
```diff
@@ -114,16 +114,70 @@ impl Service {
...
+
+                return Ok(());
+            }
+        }
+
+        drop(collections);
+
+        Err(ServiceError::NoSuchObject(String::from(
+            "The collection does not exist",
```
bilelmoussaoui, line 179: Which collection? add the alias to the error message

server/src/service.rs
```diff
@@ -114,16 +114,70 @@ impl Service {
...
+                if collection.label().await == name {
+                    drop(collections);
+                    return Ok(collection.path().clone());
+                }
+            } else if collection.alias().await == name {
+                drop(collections);
+                return Ok(collection.path().clone());
+            }
+        }
+
```
bilelmoussaoui, line 138: Log which collection wasn't found

client/src/dbus/error.rs
warusadura: Will do.

server/src/service.rs

server/src/service.rs

server/src/service.rs

server/src/service.rs
warusadura: Done, thanks!

server/src/service.rs

Note: I think KWallet does the same thing: https://invent.kde.org/frameworks/kwallet/-/blob/master/src/runtime/kwalletd/kwalletfreedesktopservice.cpp#L776

bilelmoussaoui: No it doesn't error out, it just sets the information for the default collection, nothing more.

server/src/service.rs
```diff
@@ -114,16 +114,64 @@ impl Service {
...
+                tracing::info!(
+                    "Collection: {} alias updated to {}.",
+                    collection.path(),
+                    name
+                );
+                return Ok(());
+            }
+        }
+
+        drop(collections);
```
bilelmoussaoui, line 167:
```suggestion
```

server/src/service.rs
```diff
@@ -114,16 +114,64 @@ impl Service {
...
+                // To support login alias lookup
+                if collection.label().await == name {
+                    return Ok(collection.path().clone());
+                }
+            } else if collection.alias().await == name {
+                return Ok(collection.path().clone());
+            }
+        }
+
+        drop(collections);
```
bilelmoussaoui, line 138:
```suggestion
```

server/src/service.rs
```diff
@@ -114,16 +114,64 @@ impl Service {
...
+    ) -> Result<OwnedObjectPath, ServiceError> {
+        let collections = self.collections.lock().await;
+
+        for collection in collections.iter() {
+            let collection_ifce_ref = object_server.interface::<_, Collection>(collection).await?;
+            let collection = collection_ifce_ref.get_mut().await;
+
+            if name == "login" {
+                // To support login alias lookup
+                if collection.label().await == name {
```
bilelmoussaoui, line 130:
```suggestion
            // To support login alias lookup
            if name == "login" && collection.label().await == name {
```

server/src/service.rs
```diff
@@ -114,16 +114,64 @@ impl Service {
...
-        _collection: ObjectPath<'_>,
+        name: &str,
+        collection: OwnedObjectPath,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(), ServiceError> {
-        todo!()
+        let collections = self.collections.lock().await;
+
+        for path in collections.iter() {
+            if *path == collection {
```
bilelmoussaoui, line 152:
```suggestion
            if path == &collection {
```

client/src/dbus/error.rs
bilelmoussaoui: Could you do so as part of the same commit please? To avoid having one commit per enum member.

server/src/service.rs

server/src/service.rs
warusadura: I'm getting the following clippy warning with this approach.
`warning: this `if` has identical blocks`

server/src/service.rs

server/src/service.rs
warusadura: Ah ok, thanks!

client/src/dbus/error.rs
warusadura: Yeah, look into this now.

warusadura: Yes, IsLocked also takes error message: https://gitlab.gnome.org/GNOME/gnome-keyring/-/blob/master/daemon/dbus/gkd-secret-objects.c#L519. And same for the NoSession: https://gitlab.gnome.org/GNOME/gnome-keyring/-/blob/master/daemon/dbus/gkd-secret-objects.c#L557

server/src/service.rs
bilelmoussaoui: yes that makes sense because the else if is returning the exact same error. So you should write

`if (name == "login" && collection.label().await == name ) || collection.alias().await == name`

warusadura: and also, the comment doesn't make sense anymore :)

bilelmoussaoui: why do you have to check the label for a login collection instead of it alias? because the alias is set to default? then check if the alias is default?

warusadura: Yeah, I'll take another look.

server/src/collection.rs
bilelmoussaoui, line 101: This would be gone after your rebase, as it is a property ;)

warusadura: Oh! I had already implemented these changes (signals and properties) locally. Since you said smaller prs. I just didn't wanted to trouble you with many prs at once :)

warusadura: Rebased.

server/src/service.rs
warusadura: I have double checked this. Seahorse still works without this ("login" alias lookup) change. So, I have removed this.

server/src/service.rs
bilelmoussaoui: This was removed?

bilelmoussaoui (commented):
almost there

server/src/service.rs
warusadura: > Log which collection wasn't found

I think this comment was related to service.rs:161

warusadura: Do we really need logging in ReadAlias call?

bilelmoussaoui: Yes.

warusadura: ok, I'm on it.

warusadura: Done.

server/src/service.rs
```diff
@@ -115,16 +115,62 @@ impl Service {
...
+                tracing::info!(
+                    "Collection: {} found for alias: {}",
+                    collection.path().clone(),
+                    name
+                );
+                return Ok(collection.path().clone());
+            }
+        }
+
+        tracing::info!("Collection: does not exist for alias: {}", name);
```
bilelmoussaoui, line 139:
```suggestion
        tracing::info!("Collection with alias {} does not exist.", name);
```

PR #149, opened 2024-10-29
server: Dispatch default collection items

warusadura:
And, Collection::new() now takes an additional parameter: locked. Based on this the locked status of a collection is determined. When the daemon is executed with the -l option, default collection and its items will be in unlocked status.

### DIFF
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -17,13 +17,13 @@ use tokio::sync::{Mutex, RwLock};
 use zbus::{interface, object_server::SignalEmitter, zvariant};
 use zvariant::{ObjectPath, OwnedObjectPath};
 
-use crate::{item, service_manager::ServiceManager};
+use crate::{error::Error, item, service_manager::ServiceManager};
 
 #[derive(Debug)]
 #[allow(unused)]
 pub struct Collection {
     // Properties
-    items: Mutex<Vec<item::Item>>,
+    items: Mutex<Vec<OwnedObjectPath>>,
     label: Mutex<String>,
     locked: AtomicBool,
     created: Duration,
@@ -33,7 +33,7 @@ pub struct Collection {
     #[allow(unused)]
     keyring: Arc<Keyring>,
     manager: Arc<Mutex<ServiceManager>>,
-    n_items: RwLock<i32>,
+    item_index: RwLock<u32>,
     path: OwnedObjectPath,
 }
 
@@ -64,12 +64,7 @@ impl Collection {
 
     #[zbus(property, name = "Items")]
     pub async fn items(&self) -> Vec<OwnedObjectPath> {
-        self.items
-            .lock()
-            .await
-            .iter()
-            .map(|item| OwnedObjectPath::from(item.path()))
-            .collect()
+        self.items.lock().await.clone()
     }
 
     #[zbus(property, name = "Label")]
@@ -120,6 +115,7 @@ impl Collection {
     pub fn new(
         label: &str,
         alias: &str,
+        locked: bool,
         manager: Arc<Mutex<ServiceManager>>,
         keyring: Arc<Keyring>,
     ) -> Self {
@@ -130,10 +126,10 @@ impl Collection {
         Self {
             items: Default::default(),
             label: Mutex::new(label.to_owned()),
-            locked: AtomicBool::new(true),
+            locked: AtomicBool::new(locked),
             modified: Mutex::new(created),
             alias: Mutex::new(alias.to_owned()),
-            n_items: RwLock::new(0),
+            item_index: RwLock::new(0),
             path: OwnedObjectPath::try_from(format!(
                 "/org/freedesktop/secrets/collection/{}",
                 label
@@ -155,5 +151,31 @@ impl Collection {
 
     pub async fn alias(&self) -> String {
         self.alias.lock().await.clone()
+    }
+
+    pub async fn dispatch_items(&self) -> Result<(), Error> {
+        let keyring_items = self.keyring.items().await;
+        let mut items = self.items.lock().await;
+        let service_manager = self.manager.lock().await;
+        let object_server = service_manager.object_server();
+        let mut n_items = 1;
+
+        for keyring_item in keyring_items {
+            let item = item::Item::new(
+                keyring_item.map_err(Error::InvalidItem)?,
+                self.is_locked().await,
+                Arc::clone(&self.manager),
+                self.path.clone(),
+                n_items,
+            );
+            n_items += 1;
+
+            items.push(item.path().clone());
+            object_server.at(item.path().clone(), item).await?;
+        }
+
+        *self.item_index.write().await = n_items;
+
+        Ok(())
     }
 }
--- a/server/src/item.rs
+++ b/server/src/item.rs
@@ -1,15 +1,27 @@
 // org.freedesktop.Secret.Item
 
-use std::{collections::HashMap, sync::atomic::AtomicBool};
+use std::{
+    collections::HashMap,
+    sync::{atomic::AtomicBool, Arc},
+};
 
-use oo7::dbus::{api::SecretInner, ServiceError};
+use oo7::{
+    dbus::{api::SecretInner, ServiceError},
+    portal,
+};
 use tokio::sync::Mutex;
 use zbus::zvariant::{ObjectPath, OwnedObjectPath};
 
+use crate::service_manager::ServiceManager;
+
 #[derive(Debug)]
+#[allow(unused)]
 pub struct Item {
+    // Properties
     locked: AtomicBool,
     inner: Mutex<oo7::portal::Item>,
+    // Other attributes
+    manager: Arc<Mutex<ServiceManager>>,
     path: OwnedObjectPath,
 }
 
@@ -72,7 +84,22 @@ impl Item {
 }
 
 impl Item {
-    pub fn path(&self) -> ObjectPath<'_> {
-        self.path.as_ref()
+    pub fn new(
+        item: portal::Item,
+        locked: bool,
+        manager: Arc<Mutex<ServiceManager>>,
+        collection_path: OwnedObjectPath,
+        item_index: u32,
+    ) -> Self {
+        Self {
+            locked: AtomicBool::new(locked),
+            inner: Mutex::new(item),
+            path: OwnedObjectPath::try_from(format!("{}/{}", collection_path, item_index)).unwrap(),
+            manager,
+        }
+    }
+
+    pub fn path(&self) -> &OwnedObjectPath {
+        &self.path
     }
 }
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -220,10 +220,12 @@ impl Service {
             let collection = Collection::new(
                 "login",
                 "default",
+                false,
                 Arc::clone(&service.manager),
                 Arc::new(Keyring::open("login", secret).await?),
             );
             collections.push(collection.path().clone());
+            collection.dispatch_items().await?;
             object_server
                 .at(collection.path().clone(), collection)
                 .await?;
@@ -232,6 +234,7 @@ impl Service {
         let collection = Collection::new(
             "session",
             "session",
+            false,
             Arc::clone(&service.manager),
             Arc::new(Keyring::temporary(Secret::random()).await?),
         );
@@ -239,8 +242,6 @@ impl Service {
         object_server
             .at(collection.path().clone(), collection)
             .await?;
-
-        drop(collections);
 
         Ok(())
     }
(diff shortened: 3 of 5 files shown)


server/src/collection.rs
```diff
@@ -130,7 +126,7 @@ impl Collection {
         Self {
             items: Default::default(),
             label: Mutex::new(label.to_owned()),
-            locked: AtomicBool::new(true),
+            locked: AtomicBool::new(false),
```
bilelmoussaoui, line 129: What is the cause of this change? collections should be locked by default

server/src/collection.rs
```diff
@@ -156,4 +152,24 @@ impl Collection {
     pub async fn alias(&self) -> String {
         self.alias.lock().await.clone()
     }
+
+    pub fn keyring(&self) -> Arc<Keyring> {
```
bilelmoussaoui, line 156: what is this useful for?

server/src/collection.rs
```diff
@@ -11,19 +11,20 @@ use oo7::{
...
 
-use crate::{item, service_manager::ServiceManager};
+use crate::{error::Error, item, service_manager::ServiceManager};
 
 #[derive(Debug)]
 #[allow(unused)]
 pub struct Collection {
     // Properties
-    items: Mutex<Vec<item::Item>>,
+    items: Mutex<Vec<OwnedObjectPath>>,
```
bilelmoussaoui, line 27: Why is this storing paths instead of items?

server/src/service.rs
```diff
@@ -217,34 +218,75 @@ impl Service {
...
-            let collection = Collection::new(
-                "login",
-                "default",
-                Arc::clone(&service.manager),
-                Arc::new(Keyring::open("login", secret).await?),
-            );
-            collections.push(collection.path().clone());
-            object_server
-                .at(collection.path().clone(), collection)
+        if secret.is_some() {
```
bilelmoussaoui, line 221: Why this change...

server/src/service.rs
```diff
@@ -217,34 +218,75 @@ impl Service {
...
+        );
+        self.collections
+            .lock()
+            .await
+            .push(collection.path().clone());
+
+        let keyring = collection.keyring();
+        let n_items = keyring.n_items().await;
+
+        if n_items > 0 {
```
bilelmoussaoui, line 250: This condition is useless

server/src/service.rs
```diff
@@ -217,34 +218,75 @@ impl Service {
...
+            for item in keyring.items().await {
+                items.push(match item {
+                    Ok(item) => item,
+                    Err(err) => return Err(Error::InvalidItem(err)),
+                })
+            }
+
+            for item in items {
+                collection.dispatch_items(item, object_server).await?;
+            }
```
bilelmoussaoui, line 261: why call dispatch for each item separately instead of all of them at once? why not call dispatch directly when you query it from the keyring instead of storing it in a list just for this?

server/src/service.rs
```diff
@@ -7,6 +7,7 @@ use oo7::{
         api::{Properties, SecretInner},
         Algorithm, ServiceError,
     },
+    portal,
```
bilelmoussaoui, line 10: It is only used for item, so just import it as PortalItem or use `oo7::portal::Item`

server/src/item.rs
```diff
@@ -1,15 +1,27 @@
...
-use std::{collections::HashMap, sync::atomic::AtomicBool};
+use std::{
+    collections::HashMap,
+    sync::{atomic::AtomicBool, Arc},
+};
 
-use oo7::dbus::{api::SecretInner, ServiceError};
+use oo7::{
+    dbus::{api::SecretInner, ServiceError},
+    portal,
```
bilelmoussaoui, line 10: Same as for service

server/src/service.rs
```diff
@@ -217,34 +218,75 @@ impl Service {
...
+            .lock()
+            .await
+            .push(collection.path().clone());
+
+        let keyring = collection.keyring();
+        let n_items = keyring.n_items().await;
+
+        if n_items > 0 {
+            let mut items: Vec<portal::Item> = Vec::with_capacity(n_items);
+            for item in keyring.items().await {
```
bilelmoussaoui, line 252: All of this should be in Collection, otherwise you will have to duplicate all of it for every separate collection.

server/src/collection.rs
warusadura: Removed.

server/src/service.rs

server/src/service.rs
warusadura: Yeah. removed, thanks!

server/src/service.rs
warusadura: Yeah, it was just more readable. Now removed.

server/src/service.rs

server/src/service.rs
warusadura: Yeah, I agree. Done, thanks!

server/src/collection.rs
```diff
@@ -156,4 +152,28 @@ impl Collection {
...
+                        self.path.clone(),
+                        n_items,
+                    );
+                    *self.n_items.write().await = n_items;
+
+                    self.items.lock().await.push(item.path().clone());
+                    object_server.at(item.path().clone(), item).await?;
+                }
+                Err(err) => return Err(Error::InvalidItem(err)),
+            }
```
bilelmoussaoui, line 174:
```suggestion
let keyring_items = self.keyring.items().await;
let item_index = 0; // should it start at 1?
for keyring_item in keyring_items {
    let item = item::Item::new(
        keyring_item.map_err(Error::InvalidItem)?,
        self.is_locked().await,
        Arc::clone(&self.manager),
        self.path.clone(),
        item_index,
    );
    item_index += 1

    self.items.lock().await.push(item.path().clone());
    object_server.at(item.path().clone(), item).await?;
}
*self.n_items.write().await = item_index;

```
much simpler.

server/src/collection.rs
```diff
@@ -120,6 +115,7 @@ impl Collection {
     pub fn new(
         label: &str,
         alias: &str,
+        locked: Option<bool>,
```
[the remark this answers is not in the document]
bilelmoussaoui: I don't think it should be.

server/src/collection.rs
warusadura: We really don't need to store the items itself. Just the path is enough for every usecase (I have tested this). Otherwise we have have to introduce cloning everywhere.

bilelmoussaoui: Not sure why you are "afraid" of Clone trait. All of involved types are Clone, no?

server/src/collection.rs
warusadura: Yup, done. Thanks!

server/src/collection.rs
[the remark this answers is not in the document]
warusadura: Yeah, removed Optional.

server/src/collection.rs
```diff
@@ -156,4 +152,27 @@ impl Collection {
...
+                self.path.clone(),
+                n_items,
+            );
+            n_items += 1;
+
+            self.items.lock().await.push(item.path().clone());
+            object_server.at(item.path().clone(), item).await?;
+        }
+
+        *self.n_items.write().await = n_items;
```
bilelmoussaoui, line 174: no, this should be done at the end of the iteration.

server/src/item.rs
```diff
@@ -72,7 +84,22 @@ impl Item {
...
 
 impl Item {
-    pub fn path(&self) -> ObjectPath<'_> {
-        self.path.as_ref()
+    pub fn new(
+        item: portal::Item,
+        locked: bool,
+        manager: Arc<Mutex<ServiceManager>>,
+        collection_path: OwnedObjectPath,
+        n_items: i32,
```
bilelmoussaoui, line 92: this is the index, not n_items, also should be usize, not i32.

server/src/collection.rs
warusadura: Yes, right.

server/src/item.rs
warusadura: Yup, done. Thanks!

server/src/collection.rs
```diff
@@ -156,4 +152,26 @@ impl Collection {
...
+
+        for keyring_item in keyring_items {
+            let item = item::Item::new(
+                keyring_item.map_err(Error::InvalidItem)?,
+                self.is_locked().await,
+                Arc::clone(&self.manager),
+                self.path.clone(),
+                n_items,
+            );
+            *self.item_index.write().await = n_items;
```
bilelmoussaoui, line 168: I already said two times before, this should be done at the end of the for iteration. There is 0 reason to update the value every time, as this method happens at the start of the service, so no one would be creating new items until it us up and ready.

server/src/collection.rs
```diff
@@ -156,4 +152,26 @@ impl Collection {
...
+            n_items += 1;
+
+            self.items.lock().await.push(item.path().clone());
```
bilelmoussaoui, line 171: Get the lock once, before the for iterations.

server/src/collection.rs
warusadura: I feel "not cloning the secret service interface types" approach more natural. Because, we create an object and then we dispatch it. When zbus takes care of every scenario nicely, why keep a clone just to access its attributes.

bilelmoussaoui: There is probably a mis-understanding from your side what Clone means here. It just increments the ref count, no copy of the data or anything like that, as all the inner types are either Copy (bool/numerals) or are wrapped in Arc.

bilelmoussaoui: You can keep it like this if you prefer.

server/src/collection.rs
warusadura: Ah! my bad. Actually last [time](https://github.com/bilelmoussaoui/oo7/compare/5b69092873f820f3018ccb8f9f84a3fd5cde8faf..35458ea6ba4c781797652fc33be86f0110492971) it was correctly positioned.

server/src/service_manager.rs
```diff
@@ -2,17 +2,29 @@
 
 use std::{collections::HashMap, sync::Arc};
 
-use zbus::zvariant::OwnedObjectPath;
+use zbus::{zvariant::OwnedObjectPath, Connection};
```
bilelmoussaoui, line 5: In the whole codebase, we only import items if they are used more than two times in the code. Otherwise it just makes it difficult to see where that type comes from.

server/src/collection.rs
```diff
@@ -156,4 +152,29 @@ impl Collection {
     pub async fn alias(&self) -> String {
         self.alias.lock().await.clone()
     }
+
+    pub async fn dispatch_items(&self) -> Result<(), Error> {
+        let keyring_items = self.keyring.items().await;
+        let mut items = self.items.lock().await;
+        let object_server = self.manager.lock().await.object_server().clone();
```
bilelmoussaoui, line 159:
```suggestion
        let object_server = self.manager.lock().await.object_server();
```
what is the clone needed for?

server/src/service_manager.rs
warusadura: Understood, thanks!

server/src/collection.rs
warusadura: Done.

server/src/service.rs
```diff
@@ -206,8 +204,7 @@ impl Service {
         let object_server = connection.object_server();
         let service = Self {
             collections: Default::default(),
-            manager: Default::default(),
-            connection: connection.clone(),
+            manager: Arc::new(Mutex::new(ServiceManager::new(connection.clone()))),
```
bilelmoussaoui, line 207: Why do you need to clone the connection here? nothing uses it afterwards

server/src/collection.rs
bilelmoussaoui: That doesn't look like the correct fix. Instead
```
let service_manager = self.manager.lock().await;
let object_server = service_manager.object_server();
```
should get around the error, no?

warusadura: Yes, very nicely. Thanks!

server/src/service.rs
warusadura: If I'm to loose the clone here, I'd have to clone the connection above,
```let object_server = connection.clone().object_server();```

Otherwise, `cannot move out of `connection` because it is borrowed`.

Should I apply the recommendation made by the compiler?

warusadura: Or keep this as it is?

warusadura: It doesn't work either (compiler recommendation). To remove the clone here I'll have to do: `let object_server = connection.object_server().clone();`

bilelmoussaoui (approved):

server/src/collection.rs
bilelmoussaoui: I switched it back to using property types, as get_ref/get_ref_mut gives us a global lock, so it will make debugging async code harder in this "complex" project

PR #150, opened 2024-10-31
server: Add service method Lock and Unlock implementation

warusadura:
Please see the commits.

### DIFF
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -14,10 +14,10 @@ use oo7::{
     portal::Keyring,
 };
 use tokio::sync::{Mutex, RwLock};
-use zbus::{interface, object_server::SignalEmitter, zvariant};
+use zbus::{interface, object_server::SignalEmitter, proxy::Defaults, zvariant};
 use zvariant::{ObjectPath, OwnedObjectPath};
 
-use crate::{error::Error, item, service_manager::ServiceManager};
+use crate::{error::Error, item, service_manager::ServiceManager, Service};
 
 #[derive(Debug, Clone)]
 pub struct Collection {
@@ -130,9 +130,9 @@ impl Collection {
     ) -> zbus::Result<()>;
 
     #[zbus(signal, name = "ItemChanged")]
-    async fn item_changed(
+    pub async fn item_changed(
         signal_emitter: &SignalEmitter<'_>,
-        item: OwnedObjectPath,
+        item: &OwnedObjectPath,
     ) -> zbus::Result<()>;
 }
 
@@ -203,6 +203,33 @@ impl Collection {
         }
 
         None
+    }
+
+    pub async fn set_locked(&self, locked: bool) -> Result<(), ServiceError> {
+        let items = self.items.lock().await;
+
+        for item in items.iter() {
+            item.set_locked(locked).await?;
+        }
+
+        let manager = self.manager.lock().await;
+
+        self.locked
+            .store(locked, std::sync::atomic::Ordering::Relaxed);
+        let signal_emitter = manager.signal_emitter(&self.path)?;
+        self.locked_changed(&signal_emitter).await?;
+
+        let service_path = oo7::dbus::api::Service::PATH.as_ref().unwrap();
+        let signal_emitter = manager.signal_emitter(service_path)?;
+        Service::collection_changed(&signal_emitter, &self.path).await?;
+
+        tracing::debug!(
+            "Collection: {} is {}.",
+            self.path,
+            if locked { "locked" } else { "unlocked" }
+        );
+
+        Ok(())
     }
 
     pub async fn dispatch_items(&self) -> Result<(), Error> {
--- a/server/src/item.rs
+++ b/server/src/item.rs
@@ -12,7 +12,7 @@ use oo7::{
 use tokio::sync::Mutex;
 use zbus::zvariant::{ObjectPath, OwnedObjectPath};
 
-use crate::service_manager::ServiceManager;
+use crate::{collection::Collection, service_manager::ServiceManager};
 
 #[derive(Debug, Clone)]
 pub struct Item {
@@ -21,6 +21,7 @@ pub struct Item {
     inner: Arc<Mutex<oo7::portal::Item>>,
     // Other attributes
     manager: Arc<Mutex<ServiceManager>>,
+    collection_path: OwnedObjectPath,
     path: OwnedObjectPath,
 }
 
@@ -160,6 +161,7 @@ impl Item {
         Self {
             locked: Arc::new(AtomicBool::new(locked)),
             inner: Arc::new(Mutex::new(item)),
+            collection_path: collection_path.clone(),
             path: OwnedObjectPath::try_from(format!("{}/{}", collection_path, item_index)).unwrap(),
             manager,
         }
@@ -167,5 +169,25 @@ impl Item {
 
     pub fn path(&self) -> &OwnedObjectPath {
         &self.path
+    }
+
+    pub async fn set_locked(&self, locked: bool) -> Result<(), ServiceError> {
+        let manager = self.manager.lock().await;
+
+        self.locked
+            .store(locked, std::sync::atomic::Ordering::Relaxed);
+        let signal_emitter = manager.signal_emitter(&self.path)?;
+        self.locked_changed(&signal_emitter).await?;
+
+        let signal_emitter = manager.signal_emitter(&self.collection_path)?;
+        Collection::item_changed(&signal_emitter, &self.path).await?;
+
+        tracing::debug!(
+            "Item: {} is {}.",
+            self.path,
+            if locked { "locked" } else { "unlocked" }
+        );
+
+        Ok(())
     }
 }
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -115,17 +115,21 @@ impl Service {
     #[zbus(out_args("unlocked", "prompt"))]
     pub async fn unlock(
         &mut self,
-        _objects: Vec<OwnedObjectPath>,
-    ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        objects: Vec<OwnedObjectPath>,
+    ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
+        let (unlocked, _not_unlocked) = self.set_locked(false, &objects).await?;
+
+        Ok((unlocked, OwnedObjectPath::default()))
     }
 
     #[zbus(out_args("locked", "prompt"))]
     pub async fn lock(
         &mut self,
-        _objects: Vec<OwnedObjectPath>,
-    ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        objects: Vec<OwnedObjectPath>,
+    ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
+        let (locked, _not_locked) = self.set_locked(true, &objects).await?;
+
+        Ok((locked, OwnedObjectPath::default()))
     }
 
     #[zbus(out_args("secrets"))]
@@ -231,9 +235,9 @@ impl Service {
     ) -> zbus::Result<()>;
 
     #[zbus(signal, name = "CollectionChanged")]
-    async fn collection_changed(
+    pub async fn collection_changed(
         signal_emitter: &SignalEmitter<'_>,
-        collection: OwnedObjectPath,
+        collection: &OwnedObjectPath,
     ) -> zbus::Result<()>;
 }
 
@@ -292,5 +296,58 @@ impl Service {
             .await?;
 
         Ok(())
+    }
+
+    pub async fn set_locked(
+        &self,
+        locked: bool,
+        objects: &[OwnedObjectPath],
+    ) -> Result<(Vec<OwnedObjectPath>, Vec<OwnedObjectPath>), ServiceError> {
+        let mut without_prompt = Vec::new();
+        let mut with_prompt = Vec::new();
+        let collections = self.collections.lock().await;
+
+        for object in objects {
+            for collection in collections.iter() {
+                let collection_locked = collection.is_locked().await;
+                if object == collection.path() {
+                    if collection_locked == locked {
+                        tracing::debug!(
+                            "Collection: {} is already {}.",
+                            object,
+                            if locked { "locked" } else { "unlocked" }
+                        );
+                        without_prompt.push(object.clone());
+                    } else {
+                        // TODO: remove this once the prompt implementation is complete.
+                        collection.set_locked(locked).await?;
+                        with_prompt.push(object.clone());
+                    }
+                    break;
+                } else if let Some(item) = collection.item_from_path(object).await {
+                    if locked == item.is_locked().await {
+                        tracing::debug!(
+                            "Item: {} is already {}.",
+                            object,
+                            if locked { "locked" } else { "unlocked" }
+                        );
+                        without_prompt.push(object.clone());
+                    // If the collection is in a similar state, update the Item
+                    // without a prompt.
+                    } else if collection_locked == locked {
+                        item.set_locked(locked).await?;
+                        without_prompt.push(object.clone());
+                    } else {
+                        // TODO: remove this once the prompt implementation is complete.
+                        item.set_locked(locked).await?;
+                        with_prompt.push(object.clone());
+                    }
+                    break;
+                }
+                tracing::warn!("Object: {} does not exist.", object);
+            }
+        }
+
+        Ok((without_prompt, with_prompt))
     }
 }
--- a/server/src/service_manager.rs
+++ b/server/src/service_manager.rs
@@ -2,6 +2,7 @@
 
 use std::{collections::HashMap, sync::Arc};
 
+use tokio::sync::RwLock;
 use zbus::zvariant::OwnedObjectPath;
 
 use crate::session::Session;
@@ -11,22 +12,20 @@ pub struct ServiceManager {
     connection: zbus::Connection,
     // sessions mapped to their corresponding object path on the bus
     sessions: HashMap<OwnedObjectPath, Arc<Session>>,
+    session_index: Arc<RwLock<u32>>,
 }
 
 impl ServiceManager {
     pub fn new(connection: zbus::Connection) -> Self {
         Self {
             sessions: Default::default(),
+            session_index: Default::default(),
             connection,
         }
     }
 
     pub fn object_server(&self) -> &zbus::ObjectServer {
         self.connection.object_server()
-    }
-
-    pub fn n_sessions(&self) -> usize {
-        self.sessions.len()
     }
 
     pub fn session(&self, path: &OwnedObjectPath) -> Option<Arc<Session>> {
@@ -39,5 +38,12 @@ impl ServiceManager {
 
     pub fn remove_session(&mut self, path: &OwnedObjectPath) {
         self.sessions.remove(path);
+    }
+
+    pub async fn session_index(&self) -> u32 {
+        let n_sessions = *self.session_index.read().await + 1;
+        *self.session_index.write().await = n_sessions;
+
+        n_sessions
     }
 }
--- a/server/src/service_manager.rs
+++ b/server/src/service_manager.rs
@@ -3,7 +3,7 @@
 use std::{collections::HashMap, sync::Arc};
 
 use tokio::sync::RwLock;
-use zbus::zvariant::OwnedObjectPath;
+use zbus::zvariant::{ObjectPath, OwnedObjectPath};
 
 use crate::session::Session;
 
@@ -38,6 +38,19 @@ impl ServiceManager {
 
     pub fn remove_session(&mut self, path: &OwnedObjectPath) {
         self.sessions.remove(path);
+    }
+
+    pub fn signal_emitter<'a, P>(
+        &self,
+        path: P,
+    ) -> Result<zbus::object_server::SignalEmitter<'a>, oo7::dbus::ServiceError>
+    where
+        P: TryInto<ObjectPath<'a>>,
+        P::Error: Into<zbus::Error>,
+    {
+        let signal_emitter = zbus::object_server::SignalEmitter::new(&self.connection, path)?;
+
+        Ok(signal_emitter)
     }
 
     pub async fn session_index(&self) -> u32 {
--- a/server/src/session.rs
+++ b/server/src/session.rs
@@ -30,7 +30,7 @@ impl Session {
 
 impl Session {
     pub async fn new(aes_key: Option<Arc<Key>>, manager: Arc<Mutex<ServiceManager>>) -> Self {
-        let index = manager.lock().await.n_sessions() + 1;
+        let index = manager.lock().await.session_index().await;
         Self {
             path: OwnedObjectPath::try_from(format!("/org/freedesktop/secrets/session/s{index}"))
                 .unwrap(),


warusadura:
Service method Lock should emit `PropertiesChanged` signal for each Locked item and collection. Currently oo7_daemon can't do this with this change. Can you help me figure out how to do this. So far I have tried moving set methods (set_locked) inside the Service impl (inside interface attribute) and also marking them with zbus properties attribute. None of these approaches worked so far. @bilelmoussaoui

bilelmoussaoui:
> Service method Lock should emit `PropertiesChanged` signal for each Locked item and collection

That is normal, because you have to make use of the auto-generated method called `$property_changed`. Which you should call after you updated the value of locked in the set_locked method...

warusadura:
> > Service method Lock should emit `PropertiesChanged` signal for each Locked item and collection
> 
> That is normal, because you have to make use of the auto-generated method called `$property_changed`. Which you should call after you updated the value of locked in the set_locked method...

Understood, thanks!

server/src/collection.rs
```diff
@@ -157,6 +157,15 @@ impl Collection {
         self.alias.lock().await.clone()
     }
 
+    pub async fn get_items(&self) -> Vec<item::Item> {
+        self.items.lock().await.to_vec()
+    }
```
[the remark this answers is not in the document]
bilelmoussaoui: Also cloning all the items for every object in lock, for every collection... is a bit too much. You just need to get one item that correspond to the path. So I would add a 
```rust

    pub async fn find_item( self, item: ObjectPath<'_>) -> Option<item::Item> {
         self.items.lock().await.iter().find(|i| i.path().as_ref() == item).cloned()
    }
``` 
and something similar for collections, then you iterate over the paths you receive. 

So you can write the code with 

```rust

let collections = self.collections.lock().await;
let mut locked_objects = Vec::new();
for object in objects {
    if let Some(collection) = collections.iter().find(|c| c.path() == object) {
        if !collection.is_locked().await {
            collection.set_is_locked(true);
            locked_objects.push(object);
        }
    } else {
        for collection in collections {
            if let Some(item) = collection.find_item(&object) {
                // update stuff
                if !item.is_locked().await {
                    item.set_is_locked(true);
                    locked_objects.push(object);
                }
                break;
            }
        }
    }
}
```

the problem you will have though, is duplicating the exact same function for unlocking I guess. But let us deal with that later ;)

server/src/service.rs
```diff
@@ -99,9 +99,42 @@ impl Service {
...
     pub async fn lock(
         &mut self,
-        _objects: Vec<OwnedObjectPath>,
+        objects: Vec<OwnedObjectPath>,
+        #[zbus(connection)] connection: &zbus::Connection,
+        #[zbus(signal_emitter)] signal_emitter: SignalEmitter<'_>,
     ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        let mut locked: Vec<OwnedObjectPath> = Vec::new();
+        let prompt = ObjectPath::default();
```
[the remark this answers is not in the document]
warusadura: Yeah I agree. But, a prompt isn't required here, like at all. So, returning an empty objectpath: '/' is enough. And we don't have to worry about cleaning up later :)

server/src/collection.rs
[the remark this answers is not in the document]
bilelmoussaoui: Then you should modify the set_locked method of a collection to also change the state of all the items inside of it.

And no, the spec clearly says the items passed can either be collections or items.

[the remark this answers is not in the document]
warusadura: Sorry for late reply. Yes, you're right.

server/src/collection.rs
```diff
@@ -157,6 +157,15 @@ impl Collection {
...
+
+    pub async fn set_locked(&self, locked: bool) {
```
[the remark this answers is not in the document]
warusadura: Yeah, it's already outside.

server/src/service.rs
```diff
@@ -99,9 +99,42 @@ impl Service {
...
+                            item.set_locked(true).await;
+                        }
+                        Collection::item_changed(&signal_emitter_for_item, item.path().clone())
+                            .await?;
+                    }
+                    if !collection.is_locked().await {
+                        collection.set_locked(true).await;
+                    }
+                    locked.push(collection.path().clone());
+                    Self::collection_changed(&signal_emitter, collection.path().clone()).await?;
```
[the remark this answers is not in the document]

server/src/service.rs
```diff
@@ -99,9 +99,42 @@ impl Service {
...
+                    tracing::info!("Collection: {} is locked", object);
+                    break;
+                }
+            }
+        }
+
+        if locked.is_empty() {
+            tracing::info!("Collections: {:?} not found.", objects);
+        }
```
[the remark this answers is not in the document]
warusadura: Yeah, modified.

server/src/service.rs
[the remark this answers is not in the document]
warusadura: Based on three things:
* There is nothing on the spec about keeping track of the objects that are locked with a prompt and a way to return them. This can be consider as a hint there is nothing related to actually prompting a system promt here.
* gnome-keyring also returns an empty object path as the prompt: https://gitlab.gnome.org/GNOME/gnome-keyring/-/blob/master/daemon/dbus/gkd-secret-service.c#L609
* I have tested this call many time. And never found any use case involving a prompt.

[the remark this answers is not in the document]
bilelmoussaoui: gnome-keyring-daemon is the least respectful for the secrets spec, i wouldn't follow it "blindly"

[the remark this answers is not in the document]
bilelmoussaoui: > There is nothing on the spec about keeping track of the objects that are locked with a prompt and a way to return them. This can be consider as a hint there is nothing related to actually prompting a system prompt here.

There is, the prompt has a details return variable that is a zvariant::Value, it can contain anything, even a list of object paths ;)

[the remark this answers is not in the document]
bilelmoussaoui: yes, there is no way to pass a list of items to unlock into prompt & then get the result of what got locked and what wasn't. But there should definitely be a prompt used here otherwise the spec is wrong which I don't think it is...

server/src/service.rs
```diff
@@ -99,9 +99,41 @@ impl Service {
...
+            for collection in collections.iter() {
+                if object == collection.path() || collection.items().await.contains(object) {
+                    for item in collection.inner_items().await {
+                        if !item.is_locked().await {
+                            item.set_locked(true).await;
+                            let signal_emitter = manager.signal_emitter(item.path().clone())?;
+                            item.locked_changed(&signal_emitter).await?;
+                        }
+                        let signal_emitter = manager.signal_emitter(collection.path().clone())?;
+                        Collection::item_changed(&signal_emitter, item.path()).await?;
```
bilelmoussaoui, line 120: This is not really what I meant. Instead, make a new method in impl Item, called `changed` that would get the signal emitter from the service manager, and call `Collection::item_changed` and pass the correct params to it. Same for 
`Service::collection_changed`.
You can then call that method inside set_locked so that locking/unlocking only calls that without having to deal with all the details

warusadura: Good idea, Thanks! I have named the new method as `locked_updated` because we have other properties. If you still think `changed` is more appropriate here, I'm happy to change this.

warusadura: Oh! found another issue in my impl. Fixing this now.

bilelmoussaoui: No, the changed is for the item_changed, it has nothing to do with locked. Anything specific to locked state, should be called in set_locked, no extra methods.

Also the impl is not exactly what I meant, you want to have Service::collection_changed, in a `async fn changed(&self)` for Collection type as well as the same thing for Item. No passing state of a Collection to Item.

server/src/service.rs
```diff
@@ -99,9 +99,42 @@ impl Service {
...
+                        if !item.is_locked().await {
+                            item.set_locked(true).await;
+                        }
+                        Collection::item_changed(&signal_emitter_for_item, item.path().clone())
+                            .await?;
+                    }
+                    if !collection.is_locked().await {
+                        collection.set_locked(true).await;
+                    }
+                    locked.push(collection.path().clone());
```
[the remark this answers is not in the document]
warusadura: No. Even if the collection/object/item is already locked (true) we still have to return this path. If the object is already locked (that's why the if block) we don't update/set the locked property. So, the signal PropertiesChanged is not emitted. But ItemChanged and CollectionChanged should be emitted. All 3 signals emitted if the object is not locked.

server/src/service.rs
warusadura: > Oh! found another issue in my impl. Fixing this now.

Fixed the [issue](https://github.com/bilelmoussaoui/oo7/pull/150#discussion_r1826894735)

server/src/service.rs
```diff
@@ -99,9 +99,38 @@ impl Service {
...
     ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        let mut locked: Vec<OwnedObjectPath> = Vec::new();
+        let prompt = ObjectPath::default();
+        let collections = self.collections.lock().await;
+
+        for object in &objects {
+            for collection in collections.iter() {
+                if object == collection.path() || collection.items().await.contains(object) {
+                    for item in collection.inner_items().await {
```
bilelmoussaoui, line 112: All of this shoule be done inside Collection.set_locked implementation.

server/src/collection.rs
```diff
@@ -157,6 +157,29 @@ impl Collection {
...
     }
 
+    pub async fn set_locked(&self, locked: bool) -> Result<(), ServiceError> {
+        let items = self.items.lock().await;
+
+        for item in items.iter() {
+            if !item.is_locked().await {
+                item.set_locked(true).await?;
+            }
+            let manager = self.manager.lock().await;
```
bilelmoussaoui, line 167: Why you would have to get the manager multiple times ;)

server/src/service_manager.rs
```diff
@@ -25,6 +25,15 @@ impl ServiceManager {
         self.connection.object_server()
     }
 
+    pub fn signal_emitter(
+        &self,
+        path: OwnedObjectPath,
```
bilelmoussaoui, line 30: Take a ref please.

server/src/collection.rs
```diff
@@ -157,6 +157,29 @@ impl Collection {
...
+            let signal_emitter = manager.signal_emitter(self.path.clone())?;
+            Self::item_changed(&signal_emitter, item.path()).await?;
```
bilelmoussaoui, line 169: Why do you call item_changed if you don't change the item state? that sounds useless.

server/src/collection.rs
```diff
@@ -157,6 +157,29 @@ impl Collection {
...
+        }
+
+        if !self.is_locked().await {
```
bilelmoussaoui, line 172: What is the condition for? So you cannot call set_locked if it is not locked?

bilelmoussaoui: in other words, if there is some obvious reasoning, it is not that obvious and should be documented but I think it is just wrong.

server/src/collection.rs
bilelmoussaoui: The item_changed should be called form the Item. As I stated multiple times :)

server/src/collection.rs
```diff
@@ -157,6 +157,29 @@ impl Collection {
...
+            self.locked
+                .store(locked, std::sync::atomic::Ordering::Relaxed);
+            let manager = self.manager.lock().await;
+            let signal_emitter = manager.signal_emitter(self.path.clone())?;
+            self.locked_changed(&signal_emitter).await?;
```
bilelmoussaoui, line 177: Why this doesn't call collection_changed as well?

server/src/service.rs
```diff
@@ -99,9 +99,27 @@ impl Service {
     #[zbus(out_args("locked", "prompt"))]
     pub async fn lock(
         &mut self,
-        _objects: Vec<OwnedObjectPath>,
+        objects: Vec<OwnedObjectPath>,
+        #[zbus(signal_emitter)] signal_emitter: SignalEmitter<'_>,
     ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        let mut locked: Vec<OwnedObjectPath> = Vec::new();
```
bilelmoussaoui, line 105: The type info is not needed.

server/src/service.rs
```diff
@@ -99,9 +99,27 @@ impl Service {
...
+        let prompt = ObjectPath::default();
+        let collections = self.collections.lock().await;
+
+        for object in &objects {
+            for collection in collections.iter() {
+                if object == collection.path() || collection.items().await.contains(object) {
+                    collection.set_locked(true).await?;
+                    locked.push(object.clone());
+                    Self::collection_changed(&signal_emitter, collection.path()).await?;
```
bilelmoussaoui, line 114: This should be moved to collection.set_lock

server/src/service.rs
```diff
@@ -99,9 +99,27 @@ impl Service {
...
+        #[zbus(signal_emitter)] signal_emitter: SignalEmitter<'_>,
     ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        let mut locked: Vec<OwnedObjectPath> = Vec::new();
+        let prompt = ObjectPath::default();
+        let collections = self.collections.lock().await;
+
+        for object in &objects {
+            for collection in collections.iter() {
+                if object == collection.path() || collection.items().await.contains(object) {
```
bilelmoussaoui, line 111: So the first part makes sense, but the second one, not sure about it :) because you would lock a collection just because the user locked a single item. That doesn't seem correct.

server/src/service_manager.rs
warusadura: Done, thanks!

server/src/collection.rs
warusadura: Yeah I know :) Since the Item side locks the manager first, this was the most suitable place to lock/get the manager.

server/src/service.rs
warusadura: Removed, thanks!

server/src/service.rs
warusadura: Yeah :) But, this is what gnome-keyring-daemon does. I also think this is wrong. What should I do here?

server/src/collection.rs
warusadura: My bad :) Since, we define the item_changed and it's part of the Secret Collection interface, doesn't make sense to call it inside Collection? Same for calling collection_changed inside Service?

bilelmoussaoui: Where it is deifned has nothing to do with where it can/should be called.

server/src/collection.rs
```diff
@@ -193,6 +193,29 @@ impl Collection {
         items
     }
 
+    pub async fn set_locked(&self, locked: bool) -> Result<(), ServiceError> {
+        let items = self.items.lock().await;
+
+        for item in items.iter() {
+            if !item.is_locked().await {
```
bilelmoussaoui, line 200: This is still very wrong.

server/src/collection.rs
bilelmoussaoui: hmm, what?

server/src/service.rs
bilelmoussaoui: I said before, gnome-keyring-daemon is not a reference implementation.

There should be a clear difference between case when you lock/unlock the whole collection or when you lock / unlock an item in the collection.

warusadura: Thanks for the reply.

My priority is GNOME :) We don't have this in Seahorse and KeyRack. I'm not sure about KDE side.

bilelmoussaoui: Sure, but the goal is not to re-write the same crap otherwise we can just keep gnome-keyring-daemon.

server/src/service.rs

server/src/collection.rs
warusadura: I have added to avoid emitting PropertiesChanged signal if the item is already locked. This is now moved to the item side.

server/src/service.rs
warusadura: So, we are planning to deprecate gnome-keyring-daemon because its implementation is really complex (to maintain and to add new features: TPM and homed related support). And we want the new daemon to support the latest security requirements. Other than that gnome-keyring-daemon works ok (can keep up with existing client requirements). So, I don't think that we should dismiss/ignore its inner workings, at the least the Secret Service implementation part :)

I have modified this change to support locking an item without locking the whole collection.

server/src/collection.rs
warusadura: https://github.com/bilelmoussaoui/oo7/pull/150#discussion_r1846712451

server/src/collection.rs
```diff
@@ -205,6 +205,29 @@ impl Collection {
         None
     }
 
+    pub async fn set_locked(&self, locked: bool) -> Result<(), ServiceError> {
+        let items = self.items.lock().await;
+
+        for item in items.iter() {
+            item.set_locked(true).await?;
```
bilelmoussaoui, line 212: You should instead use the value passed in param....

server/src/collection.rs
```diff
@@ -205,6 +205,29 @@ impl Collection {
...
+        }
+
+        let manager = self.manager.lock().await;
+
```
bilelmoussaoui, line 216: Same, this is wrong

server/src/collection.rs
```diff
@@ -205,6 +205,29 @@ impl Collection {
...
+        if !self.is_locked().await {
+            self.locked
+                .store(locked, std::sync::atomic::Ordering::Relaxed);
+            let signal_emitter = manager.signal_emitter(&self.path)?;
+            self.locked_changed(&signal_emitter).await?;
+        }
+
+        let service_path = OwnedObjectPath::try_from("/org/freedesktop/secrets").unwrap();
```
bilelmoussaoui, line 224: You can get the path from oo7::dbus

server/src/item.rs
```diff
@@ -168,4 +170,20 @@ impl Item {
     pub fn path(&self) -> &OwnedObjectPath {
         &self.path
     }
+
+    pub async fn set_locked(&self, locked: bool) -> Result<(), ServiceError> {
+        let manager = self.manager.lock().await;
+
+        if !self.is_locked().await {
```
bilelmoussaoui, line 177: This condition implies you cant unlock an item.

server/src/service.rs
```diff
@@ -123,9 +123,30 @@ impl Service {
...
+            for collection in collections.iter() {
+                if object == collection.path() {
+                    collection.set_locked(true).await?;
+                    locked.push(object.clone());
+                    tracing::info!("Object: {} is locked.", object);
+                    break;
+                } else if let Some(item) = collection.item_from_path(object).await {
+                    item.set_locked(true).await?;
+                    locked.push(object.clone());
+                    tracing::info!("Object: {} is locked.", object);
```
bilelmoussaoui, line 142: Rename to item and same for above

server/src/collection.rs
warusadura: `OwnedObjectPath::try_from(oo7::dbus::api::Service::PATH.as_deref().unwrap()).unwrap()` has a few unwrap. Hope this is ok?

server/src/item.rs
warusadura: Yeah, I get it. Made a small modification to avoid this.

server/src/service.rs

server/src/collection.rs
bilelmoussaoui: now you changed the impl, it was about the if !self.is_locked.

server/src/collection.rs
```diff
@@ -205,6 +205,30 @@ impl Collection {
...
+
+        if self.is_locked().await != locked {
+            self.locked
+                .store(locked, std::sync::atomic::Ordering::Relaxed);
+            let signal_emitter = manager.signal_emitter(&self.path)?;
+            self.locked_changed(&signal_emitter).await?;
+        }
+
+        let service_path =
+            OwnedObjectPath::try_from(oo7::dbus::api::Service::PATH.as_deref().unwrap()).unwrap();
```
bilelmoussaoui, line 225: It is already an ObjectPath. Maybe the signal_emitter method should be changed to take anything that can be turned into an ObjectPath instead of having to always pass a &OwnedObjectPath.

server/src/item.rs
```diff
@@ -21,6 +21,7 @@ pub struct Item {
     inner: Arc<Mutex<oo7::portal::Item>>,
     // Other attributes
     manager: Arc<Mutex<ServiceManager>>,
+    collection_path: OwnedObjectPath,
```
bilelmoussaoui, line 24: Hmm, that is something I didn't thought of when commenting. Maybe that is not a great idea after all :)

bilelmoussaoui: Actually, never mind. Given that we need it to construct the Item's path anyways, so let us keep it this way.

server/src/collection.rs

server/src/service.rs
```diff
@@ -123,9 +123,30 @@ impl Service {
...
+                    break;
+                } else if let Some(item) = collection.item_from_path(object).await {
+                    item.set_locked(true).await?;
+                    locked.push(object.clone());
+                    tracing::info!("Item: {} is locked.", object);
+                    break;
+                }
+                tracing::info!("Object: {} does not exist.", object);
+            }
+        }
```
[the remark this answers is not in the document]
warusadura: Yeah but, we still have to do a clone when:  2x `locked.push(object);`

server/src/collection.rs
```diff
@@ -205,6 +205,29 @@ impl Collection {
         None
     }
 
+    pub async fn set_locked(&self, locked: bool) -> Result<(), ServiceError> {
+        let items = self.items.lock().await;
```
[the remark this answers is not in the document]
warusadura: Thanks for the prompt review. I'm happy to do this :) but, If we return early here then we'll miss the chance to emit other Signals excluding PropertiesChanged. So, the thing is, if the collection/item is already locked(true) we won't emit PropertiesChanged. But other signals (ItemChanged, CollectionChanged) need to emit.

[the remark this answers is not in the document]
bilelmoussaoui: Why we need to emit those signals if the state didnt change ?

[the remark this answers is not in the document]
warusadura: Well this what gnome-keyring-daemon does :) Sorry we have to keep coming back to this.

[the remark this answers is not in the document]
bilelmoussaoui: That is just wrong, the client shouldnt receive an event that the state has changed when it didn't.

bilelmoussaoui:
Can you add the unlock implementation as well? Without the prompt bits obviously. Given that it will just be calling the same functions added here

warusadura:
> Can you add the unlock implementation as well? Without the prompt bits obviously. Given that it will just be calling the same functions added here

Yeah, of course. Will do.

server/src/collection.rs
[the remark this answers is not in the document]
warusadura: Removed my later if block.

server/src/service.rs
```diff
@@ -115,17 +115,61 @@ impl Service {
...
     }
 
     #[zbus(out_args("locked", "prompt"))]
     pub async fn lock(
         &mut self,
-        _objects: Vec<OwnedObjectPath>,
+        objects: Vec<OwnedObjectPath>,
     ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        let mut locked = Vec::new();
```
bilelmoussaoui, line 151: It is the exact same implementation, with only difference being the state of locked, put the common code in a function.

server/src/service.rs
```diff
@@ -115,17 +115,61 @@ impl Service {
...
+        let prompt = ObjectPath::default();
+        let collections = self.collections.lock().await;
+
+        for object in &objects {
+            for collection in collections.iter() {
+                if object == collection.path() {
+                    collection.set_locked(true).await?;
```
bilelmoussaoui, line 158: Why is this not doing the same check as unlock? See above for fixing that

bilelmoussaoui: > For unlock we need a way to identify when to issue a Prompt. But, for Lock we https://github.com/bilelmoussaoui/oo7/pull/150#discussion_r1826575597 need a prompt at all. This I'm sure :)

The spec disagree with you. 

image

The current implementations don't use a prompt here, we might not need it if we agree with both the KDE/GNOME implementors, then it can easily be disabled. But for now, let us assume it is required to be fully compatible with the published specification.

But anyways, the lock/unlock DBus functions would call the "inner" one, so the `Unlock` could grow some functionality around it and the `Lock` one won't.

warusadura: Understood, thanks!

server/src/service.rs
warusadura: Should I still follow this and move this to a new function?

bilelmoussaoui: yes

bilelmoussaoui: basically share what is common between both :)

warusadura: will do, thanks!

server/src/service.rs
bilelmoussaoui: Not sure why you are saying the implementation is wrong, it is incomplete because the whole Prompting is yet to be figured out with, once that is done both lock and unlock implementations as well as anything using Prompt. Hence me insisting on using Prompt:: default earlier

warusadura: Thanks for the reply.

Yeah it's incomplete. But,

> The purpose of service methods (Lock & Unlock) are to filter the objects locked/unlocked without prompt and send them out or send out a prompt without the objects.

And currently we're setting the locked status inside service methods. We shouldn't do this here. All we need to do this return objectspaths (locked/unlocked objects list or a prompt).

Since I can't get a feel of Lock with a prompt, I'm not sure when to issue a prompt and when to return locked objectpaths.

warusadura: > Who is supposed to modify the collection/item state then?

org.gnome.keyring.internal.Prompter.Callback [PromptReady](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/org.gnome.keyring.Prompter.xml?ref_type=heads#L73)

bilelmoussaoui: That will execute a task, the task will be created inside Lock and Unlock.

server/src/collection.rs
```diff
@@ -205,6 +205,34 @@ impl Collection {
         None
     }
 
+    #[allow(unused)]
```
bilelmoussaoui, line 208: what is this for?

server/src/service.rs
```diff
@@ -28,6 +29,7 @@ pub struct Service {
     collections: Arc<Mutex<Vec<Collection>>>,
     // Other attributes
     manager: Arc<Mutex<ServiceManager>>,
+    prompt_index: Arc<RwLock<u32>>,
```
bilelmoussaoui, line 32: Would put this in the ServiceManager instead and would pass that to newly created prompts as well

server/src/service.rs
```diff
@@ -293,4 +320,42 @@ impl Service {
 
         Ok(())
     }
+
+    pub async fn set_locked(
+        &self,
+        locked: bool,
+        objects: &Vec<OwnedObjectPath>,
```
bilelmoussaoui, line 327:
```suggestion
        objects: &[OwnedObjectPath],
```

server/src/service.rs
```diff
@@ -293,4 +320,42 @@ impl Service {
...
+    ) -> Result<Vec<OwnedObjectPath>, ServiceError> {
+        let mut locked_changed = Vec::new();
+        let collections = self.collections.lock().await;
+
+        for object in objects {
+            for collection in collections.iter() {
+                if object == collection.path() {
+                    if !collection.is_locked().await {
```
bilelmoussaoui, line 335: shouldn't this compare it to the value of locked passed to it?

server/src/service.rs
```diff
@@ -293,4 +320,42 @@ impl Service {
...
+                        tracing::debug!(
+                            "Collection: {} is already {}.",
+                            object,
+                            if locked { "locked" } else { "unlocked" }
+                        );
+                        locked_changed.push(object.clone());
+                    }
+                    break;
+                } else if let Some(item) = collection.item_from_path(object).await {
+                    if !item.is_locked().await {
```
bilelmoussaoui, line 345: Same here

server/src/service.rs
```diff
@@ -293,4 +320,42 @@ impl Service {
...
+                        tracing::debug!(
+                            "Collection: {} is already {}.",
+                            object,
+                            if locked { "locked" } else { "unlocked" }
+                        );
+                        locked_changed.push(object.clone());
+                    }
+                    break;
+                }
+                tracing::info!("Object: {} does not exist.", object);
```
bilelmoussaoui, line 355: Maybe warn here instead

server/src/service.rs
```diff
@@ -115,17 +117,41 @@ impl Service {
...
     pub async fn unlock(
         &mut self,
-        _objects: Vec<OwnedObjectPath>,
-    ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        objects: Vec<OwnedObjectPath>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
+        let unlocked = self.set_locked(false, &objects).await?;
+        if unlocked.is_empty() {
```
bilelmoussaoui, line 124: The condition here seems wrong, you would use a prompt only if there are items which don't have the expected state.

server/src/service.rs
```diff
@@ -115,17 +117,41 @@ impl Service {
...
-    ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        objects: Vec<OwnedObjectPath>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
+        let locked = self.set_locked(true, &objects).await?;
+        if locked.is_empty() {
+            let n_prompt = *self.prompt_index.read().await + 1;
+            let prompt = Prompt::new(n_prompt);
+            *self.prompt_index.write().await = n_prompt;
```
bilelmoussaoui, line 147: If you pass the service manager when creating the prompt, you could deal with incrementing the prompt index there

server/src/service.rs

server/src/service.rs
warusadura: Yeah, thanks!

server/src/service.rs

server/src/service.rs
warusadura: Done, thanks! I think we should also do the same for session :)

server/src/service.rs
warusadura: Yeah I agree.

warusadura: Quick question: we can issue a prompt if an object doesn't have the expected state. But what should we do if all the objects (more than one) in the input parameter don't have the expected state? I tried this scenario with busctl and d-spy but, nothing happens :)

server/src/service_manager.rs
```diff
@@ -40,4 +56,11 @@ impl ServiceManager {
     pub fn remove_session(&mut self, path: &OwnedObjectPath) {
         self.sessions.remove(path);
     }
+
+    pub async fn prompt_index(&mut self) -> u32 {
```
bilelmoussaoui, line 60:
```suggestion
    pub async fn prompt_index(&self) -> u32 {
```

It is unfortunate you still make the same mistake :)

server/src/service_manager.rs
```diff
@@ -11,12 +12,14 @@ pub struct ServiceManager {
...
     // sessions mapped to their corresponding object path on the bus
     sessions: HashMap<OwnedObjectPath, Arc<Session>>,
+    prompt_index: Arc<RwLock<u32>>,
 }
 
 impl ServiceManager {
     pub fn new(connection: zbus::Connection) -> Self {
         Self {
             sessions: Default::default(),
+            prompt_index: Arc::new(RwLock::new(0)),
```
bilelmoussaoui, line 22:
```suggestion
            prompt_index: Default::default(),
```
u32::default() is 0

server/src/service.rs
```diff
@@ -293,4 +314,42 @@ impl Service {
...
+        locked: bool,
+        objects: &[OwnedObjectPath],
+    ) -> Result<Vec<OwnedObjectPath>, ServiceError> {
+        let mut locked_changed = Vec::new();
+        let collections = self.collections.lock().await;
+
+        for object in objects {
+            for collection in collections.iter() {
+                if object == collection.path() {
+                    if locked == collection.is_locked().await {
```
bilelmoussaoui, line 329: this is wrong.... If you want to lock an already locked collection, then there is nothing to change and it should be skipped. So the comment below is ok, but not the `locked_changed.push()` part. Same for the Item.

server/src/service.rs
```diff
@@ -293,4 +314,42 @@ impl Service {
...
+                            object,
+                            if locked { "locked" } else { "unlocked" }
+                        );
+                        locked_changed.push(object.clone());
+                    }
+                    break;
+                } else if let Some(item) = collection.item_from_path(object).await {
+                    if locked == item.is_locked().await {
+                        tracing::debug!(
+                            "Collection: {} is already {}.",
```
bilelmoussaoui, line 341: This is not a collection.

server/src/service.rs
```diff
@@ -293,4 +314,42 @@ impl Service {
 
         Ok(())
     }
+
+    pub async fn set_locked(
+        &self,
+        locked: bool,
+        objects: &[OwnedObjectPath],
+    ) -> Result<Vec<OwnedObjectPath>, ServiceError> {
```
bilelmoussaoui, line 322: This should return a `(Vec<OwnedObjectPath>, Vec<OwnedObjectPath>)`

The first is for items that have the state to be changed, and the second for items that won't have it state changed.

server/src/service.rs
bilelmoussaoui: I am not sure what you mean, the prompt should only get the items that need change that require a prompt, the remaining should be skipped from prompting.

server/src/service.rs
```diff
@@ -115,17 +116,37 @@ impl Service {
...
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
+        let locked = self.set_locked(true, &objects).await?;
+        if locked.is_empty() {
+            let prompt = Prompt::new(Arc::clone(&self.manager)).await;
+            let path = prompt.path().clone();
+            object_server.at(path.clone(), prompt).await?;
+
+            return Ok((locked, path));
+        }
```
bilelmoussaoui, line 147: Tbh, all the pormpt related bits should be removed from this commit and included in a separate patchset that handles all the prompt related parts. Currently, you stopped locking/unlocking the items and left unused code while adding a non-working functionality

server/src/service.rs

server/src/service.rs
warusadura: No. The `locked_changed` is there to track/filter-out objects locked/unlocked without a prompt.

server/src/service.rs
```diff
@@ -273,6 +273,13 @@ impl Service {
                 Arc::new(Keyring::open("login", secret).await?),
             );
             collections.push(collection.clone());
+            {
```
bilelmoussaoui, line 295: This commit has no use cases so it should be dropped and added with whatever patch that would make use of it

server/src/service.rs
```diff
@@ -300,4 +319,51 @@ impl Service {
...
+                    if locked == item.is_locked().await {
+                        tracing::debug!(
+                            "Item: {} is already {}.",
+                            object,
+                            if locked { "locked" } else { "unlocked" }
+                        );
+                        locked_changed.push(object.clone());
+                    // If the collection is in a similar state, update locked of
+                    // an item without a prompt.
+                    } else if item.set_locked_if_collection_locked(locked).await? {
```
bilelmoussaoui, line 356: You already have the collection here, you already retrieve it state few blocks above. Just store it in a variable and drop the need for storing the collections in ServiceManager, and this method wouldn't be needed at all :)

server/src/service.rs
```diff
@@ -300,4 +319,51 @@ impl Service {
...
+                } else if let Some(item) = collection.item_from_path(object).await {
+                    if locked == item.is_locked().await {
+                        tracing::debug!(
+                            "Item: {} is already {}.",
+                            object,
+                            if locked { "locked" } else { "unlocked" }
+                        );
+                        locked_changed.push(object.clone());
+                    // If the collection is in a similar state, update locked of
+                    // an item without a prompt.
```
bilelmoussaoui, line 355: `// If the collection is in a similar state, update the Item without a prompt.` without extra verbosity

server/src/service.rs
```diff
@@ -19,9 +19,9 @@ use zbus::{
 };
 
 use crate::{
-    collection::Collection, error::Error, service_manager::ServiceManager, session::Session,
+    collection::Collection, error::Error, prompt::Prompt, service_manager::ServiceManager,
+    session::Session,
 };
-
```
bilelmoussaoui, line 24: Unrelated change

server/src/item.rs
```diff
@@ -168,4 +170,41 @@ impl Item {
     pub fn path(&self) -> &OwnedObjectPath {
         &self.path
     }
+
+    pub async fn set_locked(&self, locked: bool) -> Result<(), ServiceError> {
```
bilelmoussaoui, line 174: Why did the method for setting the collection as locked got removed...? The goal of this PR is to just make Lock/Unlock do the locking/unlocking, forget that prompting even exists.

Once that is properly implemented, create a new patch that adds the infrastructure for prompting and update Lock/Unlock to make use of that.

server/src/service.rs
warusadura: Removed, thanks!

server/src/service.rs
warusadura: Yeah! My bad, removed.

server/src/service.rs

server/src/collection.rs
warusadura: This is to avoid the unused clippy warning.

server/src/item.rs
warusadura: Re-added that part with #[allow(unused)]: https://github.com/bilelmoussaoui/oo7/pull/150#discussion_r1858844837

server/src/service.rs
```diff
@@ -112,20 +112,23 @@ impl Service {
...
-    #[zbus(out_args("unlocked", "prompt"))]
     pub async fn unlock(
         &mut self,
-        _objects: Vec<OwnedObjectPath>,
-    ) -> Result<(Vec<OwnedObjectPath>, ObjectPath), ServiceError> {
-        todo!()
+        objects: Vec<OwnedObjectPath>,
+    ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
+        let (unlocked, _not_unlocked) = self.set_locked(false, &objects).await?;
+
```
bilelmoussaoui, line 120: I would have still called the methods from Collection and Item so you don't need the unused escape. Things can be changed later on as I mentioned previously

server/src/service.rs
```diff
@@ -293,4 +296,52 @@ impl Service {
...
+        objects: &[OwnedObjectPath],
+    ) -> Result<(Vec<OwnedObjectPath>, Vec<OwnedObjectPath>), ServiceError> {
+        let mut locked_changed = Vec::new();
+        let mut locked_unchanged = Vec::new();
+        let collections = self.collections.lock().await;
+
+        for object in objects {
+            for collection in collections.iter() {
+                if object == collection.path() {
+                    if locked == collection.is_locked().await {
```
bilelmoussaoui, line 312: Get the collection locked state once, store it in a variable and use it here and below

server/src/service.rs
warusadura: Done.

server/src/service.rs
```diff
@@ -293,4 +296,57 @@ impl Service {
...
+            for collection in collections.iter() {
+                let collection_locked = collection.is_locked().await;
+                if object == collection.path() {
+                    if collection_locked == locked {
+                        tracing::debug!(
+                            "Collection: {} is already {}.",
+                            object,
+                            if locked { "locked" } else { "unlocked" }
+                        );
+                        locked_changed.push(object.clone());
```
bilelmoussaoui, line 319: Shouldn't be unchanged? And the same elsewhere.

bilelmoussaoui (commented):
Otherwise lgtm

server/src/service.rs
warusadura: [No](https://github.com/bilelmoussaoui/oo7/pull/150#discussion_r1858496398). `locked_unchanged` is there to track/filter-out objects that require a prompt to perform lock/unlock. If the object is already locked/unlocked we don't need a prompt, but returning their object paths is required. So, that why I pushed it to `locked_changed`.

server/src/service.rs
```diff
@@ -112,20 +112,23 @@ impl Service {
         Ok((unlocked, locked))
     }
 
-    #[zbus(out_args("unlocked", "prompt"))]
```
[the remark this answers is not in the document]
warusadura: Probably caused by vim :) Re-added this.

server/src/service.rs
bilelmoussaoui: Then the naming is confusing, maybe we should use some different naming like

Expected state(when the state won't change) and changed state for the other case.

warusadura: Yeah, make sense.

warusadura: is `locked` and `locked_changed` any good?

bilelmoussaoui: not really because they depend on the value of the passed parameter

server/src/service.rs
```diff
@@ -293,4 +297,57 @@ impl Service {
 
         Ok(())
     }
+
+    pub async fn set_locked(
+        &self,
+        locked: bool,
+        objects: &[OwnedObjectPath],
+    ) -> Result<(Vec<OwnedObjectPath>, Vec<OwnedObjectPath>), ServiceError> {
+        let mut locked_without_prompt = Vec::new();
```
bilelmoussaoui, line 306: maybe drop the locked from the names, otherwise with_prompt/without_prompt are decent :)

bilelmoussaoui (approved):

PR #151, opened 2024-11-02
client: Add a generic Secret type 

bilelmoussaoui:
See commits.


bilelmoussaoui:
I guess before we merge this, we should investigate if it is ok to store the content-type as an attribute. If that is the case, we can also make use of that in the portal backend implementation as it would also be relevant for the server side implementation.

bilelmoussaoui:
gnome-keyring-daemon never stores or cares about the content-type, on the other hand, kwallet stores it in an attribute https://invent.kde.org/frameworks/kwallet/-/blob/master/src/runtime/kwalletd/kwalletfreedesktopcollection.h#L16 named `$fdo_mime_type`

I guess we should store it as well, maybe named differently with a fallback for Kwallet use case

client/src/dbus/api/secret.rs
```diff
@@ -22,47 +27,42 @@ pub struct Secret<'a> {
...
     pub(crate) fn new_encrypted(
         session: Arc<Session<'a>>,
-        secret: impl AsRef<[u8]>,
-        content_type: &str,
+        secret: impl Into<Secret>,
         aes_key: &Key,
     ) -> Self {
         let iv = crypto::generate_iv();
-        let secret = crypto::encrypt(secret.as_ref(), aes_key, &iv);
+        let secret = secret.into();
```
warusadura, line 52: Isn't using into() bad?

client/src/portal/item.rs
```diff
@@ -27,7 +27,7 @@ impl Item {
     pub(crate) fn new(
         label: impl ToString,
         attributes: &impl AsAttributes,
-        secret: impl AsRef<[u8]>,
+        secret: impl Into<Secret>,
```
warusadura, line 30: Again related to into:

I have seen many calls to into. I don't really understand this yet. Can't we try to use From here and other places instead of Into.

client/src/dbus/api/secret.rs
```diff
@@ -5,15 +5,20 @@ use zbus::zvariant::{OwnedObjectPath, Type};
...
+use crate::{crypto, dbus::Error, Key, Secret};
 
-#[derive(Debug, Serialize, Deserialize, Type)]
+#[derive(Debug, Serialize, Deserialize, Type, Zeroize, ZeroizeOnDrop)]
 #[zvariant(signature = "(oayays)")]
-pub struct SecretInner(pub OwnedObjectPath, pub Vec<u8>, pub Vec<u8>, pub String);
+pub struct SecretInner(
+    #[zeroize(skip)] pub OwnedObjectPath,
+    pub Vec<u8>,
+    pub Vec<u8>,
```
warusadura, line 20: Why do we need to Zeroize here now? These (the vecs) are just encrypted (and an IV) bytes.

I think with this change I'll have to do some kind of conversion when accessing the SecretInner attribute in server side. Specifically when calling a function such as `oo7::crypto::encrypt()`.

client/src/dbus/api/secret.rs
```diff
@@ -22,47 +27,42 @@ pub struct Secret<'a> {
...
-            value: inner.2,
-            content_type: inner.3,
-        };
-        Ok(secret)
+    ) -> Result<Self, Error> {
+        Ok(Self {
+            session: Arc::new(Session::new(cnx, inner.0.clone()).await?),
+            parameters: inner.1.clone(),
+            value: inner.2.clone(),
+            content_type: inner.3.clone(),
```
warusadura, line 69: Why clone these values?

client/src/dbus/api/secret.rs
```diff
@@ -5,15 +5,20 @@ use zbus::zvariant::{OwnedObjectPath, Type};
...
+    #[zeroize(skip)] pub OwnedObjectPath,
+    pub Vec<u8>,
+    pub Vec<u8>,
+    #[zeroize(skip)] pub String,
+);
 
 #[derive(Debug, Type, Zeroize, ZeroizeOnDrop)]
 #[zvariant(signature = "(oayays)")]
-pub struct Secret<'a> {
+pub struct DBusSecret<'a> {
```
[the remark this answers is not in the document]
bilelmoussaoui: Having two secret types is confusing and one is just used for receiving and returning it through dbus

client/src/dbus/api/secret.rs
bilelmoussaoui: No, there is nothing bad about using Into. It is using it when From is the correct trait to use is the bad pattern. Here, only Into makes sense, From cannot be used to express the same "condition".

client/src/portal/item.rs
bilelmoussaoui: Same as above :)

client/src/dbus/api/secret.rs
bilelmoussaoui: because SecretInner now implements Drop so the values have to be "moved" before they get zeroized. Which is more correct.

client/src/dbus/api/secret.rs
bilelmoussaoui: What kind of conversion? all the crypto methods takes AsRef<[u8]> which is implemented by the zeroized types, so nothing to do here.

bilelmoussaoui: And no, Zeroize should always been there, but probably worth moving into a separate commit.

client/src/dbus/api/secret.rs
A6GibKm: It would better to use `take()` or something like `let {parameters, value, content_type } = inner { ...}`.

bilelmoussaoui: I dropped the Zeroize usage from this commit, it can be done as a separate step to not just block this PR further.

client/src/dbus/api/secret.rs
bilelmoussaoui: see my comment on the other review

PR #152, opened 2024-11-03
server: Add SearchItems implementation

warusadura:
This change add Secret Service method SearchItems and Secret Collection method SearchItems implementation.

### DIFF
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -46,9 +46,30 @@ impl Collection {
     #[zbus(out_args("results"))]
     pub async fn search_items(
         &self,
-        _attributes: HashMap<String, String>,
+        attributes: HashMap<String, String>,
     ) -> Result<Vec<OwnedObjectPath>, ServiceError> {
-        todo!()
+        let results = self
+            .search_inner_items(&attributes)
+            .await
+            .iter()
+            .map(|item| item.path().clone())
+            .collect::<Vec<_>>();
+
+        if results.is_empty() {
+            tracing::debug!(
+                "Items with attributes {:?} does not exist in collection: {}.",
+                attributes,
+                self.path
+            );
+        } else {
+            tracing::debug!(
+                "Items with attributes {:?} found in collection: {}.",
+                attributes,
+                self.path
+            );
+        }
+
+        Ok(results)
     }
 
     #[zbus(out_args("item", "prompt"))]
@@ -155,6 +176,21 @@ impl Collection {
 
     pub async fn alias(&self) -> String {
         self.alias.lock().await.clone()
+    }
+
+    pub async fn search_inner_items(
+        &self,
+        attributes: &HashMap<String, String>,
+    ) -> Vec<item::Item> {
+        let mut items = Vec::new();
+
+        for item in self.items.lock().await.iter() {
+            if item.attributes().await == *attributes {
+                items.push(item.clone());
+            }
+        }
+
+        items
     }
 
     pub async fn dispatch_items(&self) -> Result<(), Error> {
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -83,9 +83,33 @@ impl Service {
     #[zbus(out_args("unlocked", "locked"))]
     pub async fn search_items(
         &self,
-        _attributes: HashMap<&str, &str>,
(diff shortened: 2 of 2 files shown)


server/src/service.rs
```diff
@@ -83,9 +83,34 @@ impl Service {
...
     ) -> Result<(Vec<OwnedObjectPath>, Vec<OwnedObjectPath>), ServiceError> {
-        todo!()
+        let mut unlocked: Vec<OwnedObjectPath> = Vec::new();
+        let mut locked: Vec<OwnedObjectPath> = Vec::new();
+        let collections = self.collections.lock().await;
+
+        for collection in collections.iter() {
+            for item in collection.inner_items().await {
+                if item.attributes().await == attributes {
+                    if !item.is_locked().await {
```
bilelmoussaoui, line 95: Flip the condition, no need for negation here.

server/src/service.rs
```diff
@@ -83,9 +83,34 @@ impl Service {
     #[zbus(out_args("unlocked", "locked"))]
     pub async fn search_items(
         &self,
-        _attributes: HashMap<&str, &str>,
+        attributes: HashMap<String, String>,
     ) -> Result<(Vec<OwnedObjectPath>, Vec<OwnedObjectPath>), ServiceError> {
-        todo!()
+        let mut unlocked: Vec<OwnedObjectPath> = Vec::new();
```
bilelmoussaoui, line 88: No need to specify the type here

server/src/service.rs
```diff
@@ -83,9 +83,34 @@ impl Service {
...
+        let mut locked: Vec<OwnedObjectPath> = Vec::new();
+        let collections = self.collections.lock().await;
+
+        for collection in collections.iter() {
+            for item in collection.inner_items().await {
+                if item.attributes().await == attributes {
```
bilelmoussaoui, line 94: Is this the correct way to match an item,? I don't think so.

server/src/service.rs

server/src/service.rs
warusadura: Removed, thanks!

server/src/collection.rs
```diff
@@ -46,9 +46,30 @@ impl Collection {
     #[zbus(out_args("results"))]
     pub async fn search_items(
         &self,
-        _attributes: HashMap<String, String>,
+        attributes: HashMap<String, String>,
     ) -> Result<Vec<OwnedObjectPath>, ServiceError> {
-        todo!()
+        let results: Vec<OwnedObjectPath> = self
```
bilelmoussaoui, line 51:
```suggestion
        let results = self
```

server/src/service.rs
```diff
@@ -83,9 +83,32 @@ impl Service {
...
+                if item.is_locked().await {
+                    locked.push(item.path().clone());
+                } else {
+                    unlocked.push(item.path().clone());
+                }
+            }
+        }
+
+        if unlocked.is_empty() && locked.is_empty() {
+            tracing::info!(
```
bilelmoussaoui, line 103: Both those tracings should be debug imho

bilelmoussaoui: same for the ones you added in collection.rs

server/src/service.rs
```diff
@@ -83,9 +83,32 @@ impl Service {
...
+        let collections = self.collections.lock().await;
+
+        for collection in collections.iter() {
+            for item in collection.search_inner_items(&attributes).await {
+                if item.is_locked().await {
+                    locked.push(item.path().clone());
+                } else {
+                    unlocked.push(item.path().clone());
+                }
+            }
```
bilelmoussaoui, line 99:
```suggestion

let items = collection.search_inner_items(&attributes).await;
            for item in items {
                if item.is_locked().await {
                    locked.push(item.path().clone());
                } else {
                    unlocked.push(item.path().clone());
                }
            }
```

server/src/service.rs

server/src/service.rs

warusadura:
todo: https://github.com/bilelmoussaoui/oo7/pull/153#discussion_r1836598146

bilelmoussaoui:
Looks good enough

PR #153, opened 2024-11-03
server: Add GetSecret, GetSecrets and SetSecret implementation

warusadura:
This change adds Secret Service method GetSecrets and Secret Item methods GetSecret and SetSecret.

### DIFF
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -193,6 +193,18 @@ impl Collection {
         items
     }
 
+    pub async fn item_from_path(&self, path: &OwnedObjectPath) -> Option<item::Item> {
+        let items = self.items.lock().await;
+
+        for item in items.iter() {
+            if item.path() == path {
+                return Some(item.clone());
+            }
+        }
+
+        None
+    }
+
     pub async fn dispatch_items(&self) -> Result<(), Error> {
         let keyring_items = self.keyring.items().await;
         let mut items = self.items.lock().await;
--- a/server/src/item.rs
+++ b/server/src/item.rs
@@ -20,7 +20,7 @@ pub struct Item {
     locked: Arc<AtomicBool>,
     inner: Arc<Mutex<oo7::portal::Item>>,
     // Other attributes
-    _manager: Arc<Mutex<ServiceManager>>,
+    manager: Arc<Mutex<ServiceManager>>,
     path: OwnedObjectPath,
 }
 
@@ -32,12 +32,79 @@ impl Item {
     }
 
     #[zbus(out_args("secret"))]
-    pub async fn get_secret(&self, _session: ObjectPath<'_>) -> Result<SecretInner, ServiceError> {
-        todo!()
+    pub async fn get_secret(
+        &self,
+        session: OwnedObjectPath,
+    ) -> Result<(SecretInner,), ServiceError> {
+        let manager = self.manager.lock().await;
+
+        let Some(session) = manager.session(&session) else {
+            tracing::error!("The session `{}` does not exist.", session);
+            return Err(ServiceError::NoSession(format!(
+                "The session `{}` does not exist.",
+                session
+            )));
+        };
+
+        if self.is_locked().await {
+            tracing::error!("Cannot get secret of a locked object `{}`", self.path);
+            return Err(ServiceError::IsLocked(format!(
+                "Cannot get secret of a locked object `{}`.",
+                self.path
+            )));
+        }
+
+        let inner = self.inner.lock().await;
+        let secret = inner.secret();
+
+        tracing::debug!("Secret retrieved from the item: {}.", self.path);
+
+        match session.aes_key() {
+            Some(key) => {
+                let iv = oo7::crypto::generate_iv();
+                let encrypted = oo7::crypto::encrypt(secret, &key, &iv);
+
+                Ok((SecretInner(
+                    session.path().clone(),
+                    iv,
+                    encrypted,
+                    "text/plain".to_owned(),
+                ),))
+            }
+            None => Ok((SecretInner(
+                session.path().clone(),
+                Vec::new(),
+                secret.to_vec(),
+                "text/plain".to_owned(),
+            ),)),
+        }
     }
 
-    pub async fn set_secret(&self, _secret: SecretInner) -> Result<(), ServiceError> {
-        todo!()
+    pub async fn set_secret(&self, secret: SecretInner) -> Result<(), ServiceError> {
+        let SecretInner(session, iv, secret, _content_type) = secret;
+        let manager = self.manager.lock().await;
+
+        let Some(session) = manager.session(&session) else {
+            tracing::error!("The session `{}` does not exist.", session);
+            return Err(ServiceError::NoSession(format!(
+                "The session `{}` does not exist.",
+                session
+            )));
+        };
+
+        let mut inner = self.inner.lock().await;
+
+        match session.aes_key() {
+            Some(key) => {
+                let decrypted = oo7::crypto::decrypt(secret, &key, &iv);
+                inner.set_secret(decrypted);
+            }
+            None => {
+                inner.set_secret(secret);
+            }
+        }
+
+        Ok(())
     }
 
     #[zbus(property, name = "Locked")]
@@ -94,7 +161,7 @@ impl Item {
             locked: Arc::new(AtomicBool::new(locked)),
             inner: Arc::new(Mutex::new(item)),
             path: OwnedObjectPath::try_from(format!("{}/{}", collection_path, item_index)).unwrap(),
-            _manager: manager,
+            manager,
         }
     }
 
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -131,10 +131,37 @@ impl Service {
     #[zbus(out_args("secrets"))]
     pub async fn get_secrets(
         &self,
-        _items: Vec<OwnedObjectPath>,
-        _session: ObjectPath<'_>,
+        items: Vec<OwnedObjectPath>,
+        session: OwnedObjectPath,
     ) -> Result<HashMap<OwnedObjectPath, SecretInner>, ServiceError> {
-        todo!()
+        let mut secrets = HashMap::new();
+        let collections = self.collections.lock().await;
+
+        'outer: for collection in collections.iter() {
+            for item in &items {
+                if let Some(item) = collection.item_from_path(item).await {
+                    match item.get_secret(session.clone()).await {
+                        Ok((secret,)) => {
+                            secrets.insert(item.path().clone(), secret);
+                            // To avoid iterating through all the remaining collections, if the
+                            // items secrets are already retrieved.
+                            if secrets.len() == items.len() {
+                                break 'outer;
+                            }
+                        }
+                        // Avoid erroring out if an item is locked.
+                        Err(ServiceError::IsLocked(_)) => {
+                            continue;
+                        }
+                        Err(err) => {
+                            return Err(err);
+                        }
+                    };
+                }
+            }
+        }
+
+        Ok(secrets)
     }
 
     #[zbus(out_args("collection"))]
--- a/server/src/service_manager.rs
+++ b/server/src/service_manager.rs
@@ -29,6 +29,10 @@ impl ServiceManager {
         self.sessions.len()
     }
 
+    pub fn session(&self, path: &OwnedObjectPath) -> Option<Arc<Session>> {
+        self.sessions.get(path).map(Arc::clone)
+    }
+
     pub fn insert_session(&mut self, path: OwnedObjectPath, session: Arc<Session>) {
         self.sessions.insert(path, session);
     }
--- a/server/src/session.rs
+++ b/server/src/session.rs
@@ -10,7 +10,7 @@ use crate::service_manager::ServiceManager;
 
 #[derive(Debug, Clone)]
 pub struct Session {
-    _aes_key: Option<Arc<Key>>,
+    aes_key: Option<Arc<Key>>,
     manager: Arc<Mutex<ServiceManager>>,
     path: OwnedObjectPath,
 }
@@ -34,12 +34,16 @@ impl Session {
         Self {
             path: OwnedObjectPath::try_from(format!("/org/freedesktop/secrets/session/s{index}"))
                 .unwrap(),
-            _aes_key: aes_key,
+            aes_key,
             manager,
         }
     }
 
     pub fn path(&self) -> &OwnedObjectPath {
         &self.path
+    }
+
+    pub fn aes_key(&self) -> Option<Arc<Key>> {
+        self.aes_key.as_ref().map(Arc::clone)
     }
 }


server/src/item.rs
```diff
@@ -32,8 +32,32 @@ impl Item {
...
+        session: OwnedObjectPath,
+    ) -> Result<(SecretInner,), ServiceError> {
+        // wip
+        let inner = self.inner.lock().await;
+        let manager = self.manager.lock().await;
+        let secret = inner.secret();
+        let session = manager.session(&session);
+        if session.is_none() {
+            tracing::error!("The session does not exist");
+            return Err(ServiceError::NoSession("NoSession".to_string()));
```
bilelmoussaoui, line 46: It is useless to return NoSession in an error message that is already named NoSession. Instead pass the info about which session or a more expressive error message :)

server/src/item.rs
```diff
@@ -32,8 +32,32 @@ impl Item {
...
+        }
+
+        let session = session.unwrap();
```
bilelmoussaoui, line 49: Use let Some(session) = manager.session(&session) else { // the error bits } and you won't need the unwrap here.

server/src/item.rs
```diff
@@ -32,8 +32,32 @@ impl Item {
...
+        let iv = oo7::crypto::generate_iv();
+        let key = session.aes_key();
+
+        let secret = oo7::crypto::encrypt(secret, key, iv.clone());
```
bilelmoussaoui, line 53:
```suggestion
        let secret = oo7::crypto::encrypt(secret, key, &iv);
```
no?

server/src/service_manager.rs
```diff
@@ -29,6 +29,10 @@ impl ServiceManager {
         self.sessions.len()
     }
 
+    pub fn session(&self, path: &OwnedObjectPath) -> Option<Arc<Session>> {
+        Some(Arc::clone(self.sessions.get(path)?))
```
bilelmoussaoui, line 33: What. Use map please :)

server/src/session.rs
```diff
@@ -34,12 +34,16 @@ impl Session {
...
             manager,
         }
     }
 
     pub fn path(&self) -> &OwnedObjectPath {
         &self.path
     }
+
+    pub fn aes_key(&self) -> &Arc<Key> {
+        self.aes_key.as_ref().unwrap()
```
bilelmoussaoui, line 47: This is wrong. A "plain" session would crash here.

server/src/item.rs
```diff
@@ -32,8 +32,32 @@ impl Item {
...
-    pub async fn get_secret(&self, _session: ObjectPath<'_>) -> Result<SecretInner, ServiceError> {
-        todo!()
+    pub async fn get_secret(
+        &self,
+        session: OwnedObjectPath,
+    ) -> Result<(SecretInner,), ServiceError> {
+        // wip
+        let inner = self.inner.lock().await;
+        let manager = self.manager.lock().await;
+        let secret = inner.secret();
```
bilelmoussaoui, line 42: Move this after getting the session, no need to "read" the secret if you will fail later anyways

server/src/item.rs
bilelmoussaoui, line 53: There is nothing to encrypt if there is no aes_key

server/src/item.rs
warusadura: Thanks for the review. Yeah of course. This was just placed temporarily.

server/src/item.rs
warusadura: Yes.

server/src/session.rs
warusadura: Yeah, I missed that. Thanks!

server/src/item.rs
warusadura: Yeah, I agree.

server/src/service_manager.rs
warusadura: yeah, done :)

server/src/item.rs
warusadura: Done, thanks!

server/src/item.rs
warusadura: Done, thanks!

server/src/session.rs
```diff
@@ -34,12 +34,20 @@ impl Session {
...
         &self.path
     }
+
+    pub fn aes_key(&self) -> Option<&Key> {
+        if let Some(key) = &self.aes_key {
+            Some(key.as_ref())
+        } else {
+            None
+        }
+    }
```
bilelmoussaoui, line 52: Return a Option<Arc<Key>> like you did for session.

server/src/service.rs
```diff
@@ -235,7 +263,6 @@ impl Service {
             Arc::clone(&service.manager),
             Arc::new(Keyring::temporary(Secret::random()).await?),
         );
-        collections.push(collection.clone());
```
bilelmoussaoui, line 238: why was this removed?

server/src/service.rs
```diff
@@ -107,10 +107,38 @@ impl Service {
...
+        let mut secrets: HashMap<OwnedObjectPath, SecretInner> = HashMap::new();
+        let collections = self.collections.lock().await;
+
+        'outer: for collection in collections.iter() {
+            for item in &items {
+                if let Some(item) = collection.item_inner(item).await {
+                    match item.get_secret(session.clone()).await {
+                        Ok(secret) => {
+                            secrets.insert(item.path().clone(), secret.0);
+                            if secrets.len() == items.len() {
```
bilelmoussaoui, line 122: What is this condition for?

server/src/service.rs
```diff
@@ -107,10 +107,38 @@ impl Service {
...
+                                break 'outer;
+                            }
+                        }
+                        Err(ServiceError::NoSession(err)) => {
+                            return Err(ServiceError::NoSession(err));
+                        }
+                        Err(_) => (),
```
bilelmoussaoui, line 129: This sounds just wrong?

server/src/service.rs
```diff
@@ -107,10 +107,38 @@ impl Service {
...
+        items: Vec<OwnedObjectPath>,
+        session: OwnedObjectPath,
     ) -> Result<HashMap<OwnedObjectPath, SecretInner>, ServiceError> {
-        todo!()
+        let mut secrets: HashMap<OwnedObjectPath, SecretInner> = HashMap::new();
+        let collections = self.collections.lock().await;
+
+        'outer: for collection in collections.iter() {
+            for item in &items {
+                if let Some(item) = collection.item_inner(item).await {
```
bilelmoussaoui, line 118: Instead of adding this function, you should make the search_items_inner function from the that other PR takes a search condition function, that way you can use it both for comparing attributes or item paths.

server/src/service.rs
```diff
@@ -107,10 +107,38 @@ impl Service {
     #[zbus(out_args("secrets"))]
     pub async fn get_secrets(
         &self,
-        _items: Vec<OwnedObjectPath>,
-        _session: ObjectPath<'_>,
+        items: Vec<OwnedObjectPath>,
+        session: OwnedObjectPath,
     ) -> Result<HashMap<OwnedObjectPath, SecretInner>, ServiceError> {
-        todo!()
+        let mut secrets: HashMap<OwnedObjectPath, SecretInner> = HashMap::new();
```
bilelmoussaoui, line 113: the type is not needed here.

server/src/item.rs
```diff
@@ -32,8 +32,49 @@ impl Item {
...
+        if self.is_locked().await {
+            tracing::error!("Cannot get secret of a locked object.");
+            return Err(ServiceError::IsLocked(
+                "Cannot get secret of a locked object.".to_string(),
+            ));
+        }
+
+        let inner = self.inner.lock().await;
+        let secret = inner.secret();
+        let iv = oo7::crypto::generate_iv();
```
bilelmoussaoui, line 60: Is there really an iv passed for a not encrypted session?

server/src/item.rs
```diff
@@ -32,8 +32,49 @@ impl Item {
...
+        let session_inner;
+
+        if let Some(session) = manager.session(&session) {
+            session_inner = session;
+        } else {
+            tracing::error!("The session: {} does not exist.", session);
+            return Err(ServiceError::NoSession(
+                "The session does not exist.".to_string(),
+            ));
+        }
```
bilelmoussaoui, line 49:
```suggestion
        let Some(session) = manager.session(&session) else {
            tracing::error!("The session: {} does not exist.", session);
            return Err(ServiceError::NoSession(
                "The session does not exist.".to_string(),
            ));
        }
```

server/src/service.rs
warusadura: Thanks! Sorry for the late reply :)
Yeah it make sense. But, the returns are different. For this case we're returning a single Item based on the objectpath and in SearchItems we're returning multiple Items based on attributes.
So, it is bad to have two methods: `search_inner_items(attributes)` and `inner_item(objectpath)`?

bilelmoussaoui: The name is just so weird. Maybe find_item_with_path or so?

warusadura: yeah, I agree :) will update it.

server/src/session.rs
warusadura: Done, thanks!

server/src/item.rs
warusadura: Yeah, removed.

server/src/service.rs
warusadura: Thanks, removed.

server/src/item.rs
warusadura: Done, thanks!

server/src/service.rs
warusadura: This change is introduced from a separate [commit](https://github.com/bilelmoussaoui/oo7/pull/153/commits/764ddb9f36a4db498caa77cd646cebc346890525).

> Since session collection is a temporary collection without any items
we shouldn't push it to the Service Collections.

With this change we can avoid an extra iteration when looking for items.

server/src/service.rs
warusadura: This is to avoid iterating through all the available collections (Service.Collections). If we have found the matching items to dbus call arguments, there is no need to continue iterating through remaining collections.

server/src/service.rs
warusadura: Modified to return the error.

server/src/service.rs
warusadura: Yeah, thanks! Removed.

server/src/item.rs
```diff
@@ -32,8 +32,47 @@ impl Item {
...
+        };
+
+        tracing::info!("Secret retrieved from the item: {}.", self.path);
+
+        Ok((SecretInner(
+            session.path().clone(),
+            iv,
+            encrypted,
+            "text/plain".to_string(),
+        ),))
```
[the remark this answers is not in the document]
warusadura: This is much cleaner, thanks!

server/src/service.rs
```diff
@@ -107,10 +107,40 @@ impl Service {
...
+                        }
+                        Err(err) => {
+                            return Err(err);
+                        }
+                    };
+                } else {
+                    tracing::info!(
+                        "Item: {:?} does not exist in collection: {}",
+                        item,
+                        collection.path().clone()
```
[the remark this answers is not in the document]

server/src/item.rs
```diff
@@ -32,8 +32,47 @@ impl Item {
...
+
+        let Some(session) = manager.session(&session) else {
+            tracing::error!("The session: {} does not exist.", session);
+            return Err(ServiceError::NoSession(
+                "The session does not exist.".to_string(),
+            ));
+        };
+
+        if self.is_locked().await {
+            tracing::error!("Cannot get secret of a locked object.");
```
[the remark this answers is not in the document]
warusadura: Done.

server/src/item.rs
```diff
@@ -32,8 +32,47 @@ impl Item {
...
+    pub async fn get_secret(
+        &self,
+        session: OwnedObjectPath,
+    ) -> Result<(SecretInner,), ServiceError> {
+        let manager = self.manager.lock().await;
+
+        let Some(session) = manager.session(&session) else {
+            tracing::error!("The session: {} does not exist.", session);
+            return Err(ServiceError::NoSession(
+                "The session does not exist.".to_string(),
```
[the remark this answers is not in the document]
warusadura: Done, thanks!

server/src/collection.rs
```diff
@@ -157,6 +157,18 @@ impl Collection {
         self.alias.lock().await.clone()
     }
 
+    pub async fn get_item_from_path(&self, path: &OwnedObjectPath) -> Option<item::Item> {
```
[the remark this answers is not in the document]

server/src/service.rs
```diff
@@ -107,10 +107,40 @@ impl Service {
...
+        session: OwnedObjectPath,
     ) -> Result<HashMap<OwnedObjectPath, SecretInner>, ServiceError> {
-        todo!()
+        let mut secrets = HashMap::new();
+        let collections = self.collections.lock().await;
+
+        'outer: for collection in collections.iter() {
+            for item in &items {
+                if let Some(item) = collection.get_item_from_path(item).await {
+                    match item.get_secret(session.clone()).await {
```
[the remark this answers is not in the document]
warusadura: Yeah I could have done this. But, actually we need the match. Because, consider a scenario where a correct Session is present/provided. But the item(s) are locked. In this situation we still need to return `secrets: HashMap<OwnedObjectPath, SecretInner>`. We can error out only if there is no session (returns nothing). Of course this is based on gnome-keyring behavior :)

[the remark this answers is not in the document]
bilelmoussaoui: The match you wrote doesn't do what you are mentioning :)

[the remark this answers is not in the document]
warusadura: Ha :) yes, you're right.

The following is what we need:
Based on gnome-keyring behavior,  `Item.get_secret` call error out if there is no session or the item is locked. But `Service.get_secrets` only error out when there is no session.

I had this correctly implemented before [this](https://github.com/bilelmoussaoui/oo7/pull/153#discussion_r1844014981) change,

I added another arm:
```
Err(ServiceError::IsLocked(_)) => {                                                   
      continue;
 }
```

with this we don't error out anymore if the items are locked when calling Service.get_secrets.

If this is still bad, how can I improve this?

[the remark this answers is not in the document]
bilelmoussaoui: Well, first of all this difference in behaviour between Item::get_secret & Service::get_secrets should be documented in Service::get_secrets along with a reference to gnome-keyring-daemon.

Also, you don't need all the other match statements, only the Ok/Err(IsLocked)/Err(e), are enough.

server/src/item.rs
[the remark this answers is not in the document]
bilelmoussaoui: No, you only did for the error, not the logged message. The error is received by the client, the logs is what we might all have in some scenarios.

server/src/service.rs
bilelmoussaoui: Then add a small comment explaining why.

server/src/service.rs
```diff
@@ -131,10 +131,43 @@ impl Service {
...
+                            return Err(err);
+                        }
+                    };
+                } else {
+                    tracing::debug!(
+                        "Item: {:?} does not exist in collection: {}",
+                        item,
+                        collection.path()
+                    );
+                }
```
bilelmoussaoui, line 166: This is useless.

bilelmoussaoui: I meant the debugging message, it will just print too many debug logs for nothing

server/src/service.rs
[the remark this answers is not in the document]
warusadura: Got it, thanks!

server/src/service.rs
warusadura: Done.

server/src/service.rs
```diff
@@ -131,10 +131,43 @@ impl Service {
...
-        _items: Vec<OwnedObjectPath>,
-        _session: ObjectPath<'_>,
+        items: Vec<OwnedObjectPath>,
+        session: OwnedObjectPath,
     ) -> Result<HashMap<OwnedObjectPath, SecretInner>, ServiceError> {
-        todo!()
+        let mut secrets = HashMap::new();
+        let collections = self.collections.lock().await;
+
+        'outer: for collection in collections.iter() {
```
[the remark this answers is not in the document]
warusadura: Why though :) is it bad practice to use a labeled loop block?

server/src/item.rs
```diff
@@ -32,12 +32,83 @@ impl Item {
...
+                session
+            )));
+        };
+
+        let mut inner = self.inner.lock().await;
+
+        match session.aes_key() {
+            Some(key) => {
+                let iv = secret.1;
+                let secret_encrypted = secret.2;
```
bilelmoussaoui, line 100:
```suggestion
                let {session, iv, encrypted_secret, _content_type} = secret;
```
move it to the top

bilelmoussaoui: we will handle content_type later on in the future.

server/src/service.rs
```diff
@@ -131,10 +131,38 @@ impl Service {
...
+                        Ok(secret) => {
+                            secrets.insert(item.path().clone(), secret.0);
+                            // To avoid iterating through all the remaining collection, if the items
+                            // secrets are already retrieved.
+                            if secrets.len() == items.len() {
+                                break 'outer;
+                            }
+                        }
+                        // Based on gnome-keyring-daemon behavior, to avoid error out when the item
+                        // is Locked.
```
bilelmoussaoui, line 153:
```suggestion
                        // Avoid erroring out if an item is locked.
```

server/src/service.rs
```diff
@@ -131,10 +131,38 @@ impl Service {
...
+        let mut secrets = HashMap::new();
+        let collections = self.collections.lock().await;
+
+        'outer: for collection in collections.iter() {
+            for item in &items {
+                if let Some(item) = collection.item_from_path(item).await {
+                    match item.get_secret(session.clone()).await {
+                        Ok(secret) => {
+                            secrets.insert(item.path().clone(), secret.0);
+                            // To avoid iterating through all the remaining collection, if the items
```
bilelmoussaoui, line 146:
```suggestion
                            // To avoid iterating through all the remaining collections, if the items
```

server/src/service.rs
```diff
@@ -131,10 +131,38 @@ impl Service {
...
-        todo!()
+        let mut secrets = HashMap::new();
+        let collections = self.collections.lock().await;
+
+        'outer: for collection in collections.iter() {
+            for item in &items {
+                if let Some(item) = collection.item_from_path(item).await {
+                    match item.get_secret(session.clone()).await {
+                        Ok(secret) => {
+                            secrets.insert(item.path().clone(), secret.0);
```
bilelmoussaoui, line 145:
```suggestion
                        Ok((secret,)) => {
                            secrets.insert(item.path().clone(), secret);
```

server/src/service.rs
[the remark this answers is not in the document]
bilelmoussaoui: Nothing per see, just trying to understand what is the most efficient way to do all of this.

bilelmoussaoui (commented):
almost there

bilelmoussaoui:
The commit message and it description could use some work, in general. Currently you just duplicate the commit message "title" with the "description". In the description, add any technical information or detail or leave it empty.

server/src/item.rs
warusadura: Done, thanks!

server/src/service.rs

server/src/service.rs

server/src/service.rs

PR #164, opened 2024-11-28
server: CreateItem and Delete item implementation

warusadura:
Please see the commits.

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -18,3 +18,4 @@ tokio = { workspace = true, features = ["full"] }
 tracing = "0.1"
 tracing-subscriber.workspace = true
 zbus.workspace = true
+zeroize.workspace = true
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -75,11 +75,55 @@ impl Collection {
     #[zbus(out_args("item", "prompt"))]
     pub async fn create_item(
         &self,
-        _properties: Properties,
-        _secret: SecretInner,
-        _replace: bool,
-    ) -> Result<(OwnedObjectPath, ObjectPath), ServiceError> {
-        todo!()
+        properties: Properties,
+        secret: SecretInner,
+        replace: bool,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+        #[zbus(signal_emitter)] signal_emitter: zbus::object_server::SignalEmitter<'_>,
+    ) -> Result<(OwnedObjectPath, OwnedObjectPath), ServiceError> {
+        let SecretInner(session, iv, secret, _content_type) = secret;
+        let label = properties.label();
+        // Safe to unwrap as an item always has attributes
+        let attributes = properties.attributes().unwrap();
+
+        let Some(session) = self.service.session(&session).await else {
+            tracing::error!("The session `{}` does not exist.", session);
+            return Err(ServiceError::NoSession(format!(
+                "The session `{}` does not exist.",
+                session
+            )));
+        };
+
+        let secret = match session.aes_key() {
+            Some(key) => oo7::crypto::decrypt(secret, &key, &iv),
+            None => zeroize::Zeroizing::new(secret),
+        };
+
+        let item = self
+            .keyring
+            .create_item(label, &attributes, secret, replace)
+            .await
+            .map_err(|err| {
+                ServiceError::ZBus(zbus::Error::FDO(Box::new(zbus::fdo::Error::Failed(
+                    format!("Failed to create a new item {err}."),
+                ))))
+            })?;
+
+        let n_items = *self.item_index.read().await;
+        let item = item::Item::new(item, false, self.service.clone(), &self.path, n_items);
+        *self.item_index.write().await = n_items + 1;
+
+        self.items.lock().await.push(item.clone());
+
+        let path = item.path().clone();
+        object_server.at(path.clone(), item).await?;
+
+        Self::item_created(&signal_emitter, &path).await?;
+        self.items_changed(&signal_emitter).await?;
+
+        tracing::info!("Item `{}` created.", &path);
+
+        Ok((path, OwnedObjectPath::default()))
     }
 
     #[zbus(property, name = "Items")]
@@ -120,7 +164,7 @@ impl Collection {
     #[zbus(signal, name = "ItemCreated")]
     async fn item_created(
         signal_emitter: &SignalEmitter<'_>,
-        item: OwnedObjectPath,
+        item: &OwnedObjectPath,
     ) -> zbus::Result<()>;
 
     #[zbus(signal, name = "ItemDeleted")]
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -168,9 +168,9 @@ impl Collection {
     ) -> zbus::Result<()>;
 
     #[zbus(signal, name = "ItemDeleted")]
-    async fn item_deleted(
+    pub async fn item_deleted(
         signal_emitter: &SignalEmitter<'_>,
-        item: OwnedObjectPath,
+        item: &OwnedObjectPath,
     ) -> zbus::Result<()>;
 
     #[zbus(signal, name = "ItemChanged")]
@@ -295,6 +295,38 @@ impl Collection {
         }
 
         *self.item_index.write().await = n_items;
+
+        Ok(())
+    }
+
+    pub async fn delete_item(&self, path: &OwnedObjectPath) -> Result<(), ServiceError> {
+        let Some(item) = self.item_from_path(path).await else {
+            return Err(ServiceError::NoSuchObject(format!(
+                "Item `{}` does not exist.",
+                path
+            )));
+        };
+
+        if item.is_locked().await {
+            return Err(ServiceError::IsLocked(format!(
+                "Cannot delete a locked item `{}`",
+                path
+            )));
+        }
+
+        let attributes = item.attributes().await;
+        self.keyring.delete(&attributes).await.map_err(|err| {
+            ServiceError::ZBus(zbus::Error::FDO(Box::new(zbus::fdo::Error::Failed(
+                format!("Failed to deleted item {err}."),
+            ))))
+        })?;
+
+        let mut items = self.items.lock().await;
+        items.retain(|item| item.path() != path);
+        drop(items);
+
+        let signal_emitter = self.service.signal_emitter(&self.path)?;
+        self.items_changed(&signal_emitter).await?;
 
         Ok(())
     }
--- a/server/src/item.rs
+++ b/server/src/item.rs
@@ -10,7 +10,7 @@ use oo7::{
     portal,
 };
 use tokio::sync::Mutex;
-use zbus::zvariant::{ObjectPath, OwnedObjectPath};
+use zbus::zvariant::OwnedObjectPath;
 
 use crate::{collection::Collection, Service};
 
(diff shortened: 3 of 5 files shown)


server/src/collection.rs
```diff
@@ -75,11 +75,60 @@ impl Collection {
     #[zbus(out_args("item", "prompt"))]
     pub async fn create_item(
```
bilelmoussaoui, line 76: Please split Create & Delete into separate commits.

server/src/collection.rs
```diff
@@ -75,11 +75,60 @@ impl Collection {
...
+            tracing::error!("The session `{}` does not exist.", session);
+            return Err(ServiceError::NoSession(format!(
+                "The session `{}` does not exist.",
+                session
+            )));
+        };
+
+        let secret = match session.aes_key() {
+            Some(key) => oo7::crypto::decrypt(secret, &key, &iv),
+            None => zeroize::Zeroizing::new(secret),
```
bilelmoussaoui, line 99: You don't have to wrap this in Zeroize, using to_vec() above should be enough I think.

server/src/collection.rs
```diff
@@ -75,11 +75,60 @@ impl Collection {
...
+        };
+
+        let item = match self
+            .keyring
+            .create_item(label, &attributes, secret, replace)
+            .await
+        {
+            Ok(item) => item,
+            Err(_) => todo!(), // TODO: may be new ServiceError
+        };
```
bilelmoussaoui, line 109: You can't add new ServiceError, that would not be compatible with the current specification. Instead, I would tracing::error the error message or so.

server/src/collection.rs
```diff
@@ -75,11 +75,60 @@ impl Collection {
...
+            Arc::clone(&self.manager),
+            &self.path,
+            n_items,
+        );
+        *self.item_index.write().await = n_items + 1;
+
+        self.items.lock().await.push(item.clone());
+
+        Self::item_created(&signal_emitter, item.path()).await?;
+        self.items_changed(&signal_emitter).await?;
```
bilelmoussaoui, line 124: should be emitted after the item is served.

server/src/collection.rs
```diff
@@ -75,11 +75,60 @@ impl Collection {
...
+            .await
+        {
+            Ok(item) => item,
+            Err(_) => todo!(), // TODO: may be new ServiceError
+        };
+
+        let n_items = *self.item_index.read().await;
+        let item = item::Item::new(
+            item,
+            self.is_locked().await,
```
bilelmoussaoui, line 114: Can you create an item on a locked collection? i don't think so, so it is safe to assume this would be unlocked. Leave a comment explaining why and add the usual check if the collection.is_locked & error message

server/src/collection.rs
```diff
@@ -257,4 +306,23 @@ impl Collection {
 
         Ok(())
     }
+
+    pub async fn delete_item(&self, path: &OwnedObjectPath) -> Result<(), Error> {
+        let Some(item) = self.item_from_path(path).await else {
+            tracing::debug!("Item: {} does not exist.", path);
+            return Err(Error::NoSuchObject);
```
bilelmoussaoui, line 313: Shouldn't the NoSuchObject take a string ?

server/src/collection.rs
```diff
@@ -257,4 +306,23 @@ impl Collection {
...
+        };
+
+        let attributes = item.attributes().await;
+        self.keyring.delete(&attributes).await?; // TODO: we need a new method?
```
bilelmoussaoui, line 317: For what?

server/src/item.rs
```diff
@@ -28,8 +28,39 @@ pub struct Item {
...
     #[zbus(out_args("prompt"))]
-    pub async fn delete(&self) -> Result<ObjectPath, ServiceError> {
-        todo!()
+    pub async fn delete(
+        &self,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<OwnedObjectPath, ServiceError> {
+        let manager = self.manager.lock().await;
+
+        let Some(collection) = manager.collection(&self.collection_path) else {
```
bilelmoussaoui, line 37: Same, you cannot delete an item from a locked collection

server/src/item.rs
```diff
@@ -28,8 +28,39 @@ pub struct Item {
...
+            )));
+        };
+
+        drop(manager);
+        match collection.delete_item(&self.path).await {
+            Ok(_) => {
+                tracing::info!("Item `{}` deleted.", &self.path);
+            }
+            Err(_) => {
+                return Err(ServiceError::NoSuchObject(format!(
```
bilelmoussaoui, line 50: Is this the only possible error? I don't think so. It would be nice if delete_item returned a ServiceError already as it is only used here and then you can just call collection.delete_item(&self.path).await?;

server/src/item.rs
```diff
@@ -28,8 +28,39 @@ pub struct Item {
...
+                    "Item `{}` does not exist.",
+                    &self.path
+                )));
+            }
+        }
+
+        object_server.remove::<Self, _>(&self.path).await?;
+
+        let manager = self.manager.lock().await;
```
bilelmoussaoui, line 59: Why you drop the manager above just to get it again further?

server/src/service_manager.rs
```diff
@@ -5,20 +5,23 @@ use std::{collections::HashMap, sync::Arc};
...
 
-use crate::session::Session;
+use crate::{collection::Collection, session::Session};
 
 #[derive(Debug)]
 pub struct ServiceManager {
     connection: zbus::Connection,
     // sessions mapped to their corresponding object path on the bus
     sessions: HashMap<OwnedObjectPath, Arc<Session>>,
+    // collections mapped to their corresponding object path on the bus
```
bilelmoussaoui, line 15: I am not a fan of duplicating this information between Service and ServiceManager. Service already knows about the collections and I don't think it makes sense to move them from there to ServiceManager. 

I see that we are pushing everything into ServiceManager because it is easy but that leads to spaghetti code. Instead, I think another approach would be to give a back reference to the Service in the ServiceManager then you would let the Service handle both the Collections and the sessions as it is the object responsible for them.

server/src/collection.rs
warusadura: `None => secret.to_vec().into()`

server/src/collection.rs
warusadura: Understood, thanks!

warusadura: Ah, I still need to return an error from here. Can't use `.map_err` either, missing a trait impl.

bilelmoussaoui: You can, by using ServiceError::Zbus and then creating a custom zbus error message

bilelmoussaoui: basically zbus::Error::FDO(zbus::fdo::Error::Failed(format!("Failed to do {e}")))`

warusadura: got it, thank you!

server/src/service_manager.rs
warusadura: Sorry for the delay. Implemented through: https://github.com/bilelmoussaoui/oo7/pull/165

server/src/collection.rs

server/src/collection.rs
warusadura: Yeah. I think passing `false` here is enough. Because, when calling CreateItem, a call to Service.Unlock happens before, automatically. After the Service.Unlock call returns CreateItem is on.

bilelmoussaoui: >  Because, when calling CreateItem, a call to Service.Unlock happens before, automatically. After the Service.Unlock call returns CreateItem is on.

What do you mean? that is not true at all. The user is expected to unlock the collection, it is not done automatically

warusadura: I meant when calling CreateItem, regardless of the current locked status, a call to Service.Unlock happens beforehand (I'm not calling this so, automatically). When that call returns CreateItem starts. This is what I see with Bustle (with gnome-keyring-daemon and now  and with oo7-daemon)  :)

server/src/collection.rs
warusadura: Done.

server/src/item.rs
warusadura: Done.

server/Cargo.toml
```diff
@@ -18,3 +18,4 @@ tokio = { workspace = true, features = ["full"] }
 tracing = "0.1"
 tracing-subscriber.workspace = true
 zbus.workspace = true
+zeroize.workspace = true
```
bilelmoussaoui, line 21: This is no longer used

server/src/collection.rs
```diff
@@ -75,11 +75,57 @@ impl Collection {
...
-        todo!()
+        properties: Properties,
+        secret: SecretInner,
+        replace: bool,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+        #[zbus(signal_emitter)] signal_emitter: zbus::object_server::SignalEmitter<'_>,
+    ) -> Result<(OwnedObjectPath, OwnedObjectPath), ServiceError> {
+        let SecretInner(session, iv, secret, _content_type) = secret;
+        let label = properties.label();
+        let attributes = properties.attributes().unwrap();
```
bilelmoussaoui, line 86:
```suggestion
// Safe to unwrap as an item always has attributes
        let attributes = properties.attributes().unwrap();
```

server/src/collection.rs
```diff
@@ -75,11 +75,57 @@ impl Collection {
...
+                session
+            )));
+        };
+
+        let secret = match session.aes_key() {
+            Some(key) => oo7::crypto::decrypt(secret, &key, &iv),
+            None => zeroize::Zeroizing::new(secret),
+        };
+
+        let item = match self
```
bilelmoussaoui, line 101: Use map_err instead.

server/src/collection.rs
bilelmoussaoui: This got reverted.

server/src/collection.rs
```diff
@@ -254,4 +300,39 @@ impl Collection {
...
+
+        let attributes = item.attributes().await;
+        match self.keyring.delete(&attributes).await {
+            Ok(_) => tracing::debug!("Item `{}` deleted.", path),
+            Err(err) => {
+                return Err(ServiceError::ZBus(zbus::Error::FDO(Box::new(
+                    zbus::fdo::Error::Failed(format!("Failed to deleted item {err}.")),
+                ))))
+            }
+        };
```
bilelmoussaoui, line 327: Same here

server/src/collection.rs

server/src/collection.rs
warusadura: Sorry for the delay.

So, the issue is `delete` remove all the items that matches the attributes. All the items created using Seahorse have the same attributes: `"xdg:schema" "org.gnome.keyring.Note"`. So, using this method removes everything :)

server/Cargo.toml
warusadura: Sorry, I don't understand. Also, [this](https://github.com/bilelmoussaoui/oo7/pull/164#discussion_r1864965430)

server/src/collection.rs
warusadura: The same happens with Key Rack.

server/Cargo.toml
bilelmoussaoui: As mentioned before, you don't to use zeroize in that specific case but it is okay.

server/src/collection.rs
bilelmoussaoui: Please open an issue then. We have to fix this asap.

server/src/collection.rs
bilelmoussaoui: I don't get what you mean still. Who calls unlock?

bilelmoussaoui:
Let us merge this, we can iterate with time to fix little details.

PR #165, opened 2024-12-01
server: Remove ServiceManager

warusadura:
Although it was easier to keep the sessions and other utility functions stored inside the ServiceManager, storing them inside the Service makes more sense. Because, the Service is responsible for handling them.


server/src/service_manager.rs
```diff
@@ -28,22 +28,24 @@ impl ServiceManager {
...
-
-    pub fn insert_session(&mut self, path: OwnedObjectPath, session: Arc<Session>) {
-        self.sessions.insert(path, session);
+    pub fn service(&self) -> Result<Arc<Service>, ServiceError> {
+        if let Some(service) = &self.service {
+            Ok(Arc::clone(service))
+        } else {
+            Err(ServiceError::ZBus(zbus::Error::FDO(Box::new(
+                zbus::fdo::Error::Failed("Service isn't initialized yet.".to_string()),
+            ))))
```
bilelmoussaoui, line 37: This should be unrechable code instead of something that returns an error.

warusadura: Thanks for the prompt review. Sorry, I don't understand :)

bilelmoussaoui: It is a programmer error to call any method that needs the server before the server instance is up and ready. So there is no error to report back, if that happens, you should panic using a macro called unreachable or so.

server/src/service.rs
```diff
@@ -256,9 +258,19 @@ impl Service {
         let object_server = connection.object_server();
         let service = Self {
             collections: Default::default(),
+            sessions: Default::default(),
             manager: Arc::new(Mutex::new(ServiceManager::new(connection.clone()))),
         };
 
+        {
+            let service_clone = Arc::new(service.clone());
```
bilelmoussaoui, line 266: What is the use of Arc here id the service is clonable anyways?

server/src/service_manager.rs
warusadura: Understood, thanks!

server/src/service.rs
warusadura: Mmm.. we can't `Arc::clone` without a `Arc::new` right?

bilelmoussaoui: Why you need Arc::clone anyways?

bilelmoussaoui: Cloning the service would just increment the ref count of every instance in it, so there is no need for another Arc wrapper.

warusadura: Oh, I see. Thanks!

warusadura: A quick question: Isn't Arc::clone is really fast and uses less memory compared to a normal clone? And putting a Arc wrapped Service in ServiceManager might be good, because we're accessing it many times in different places. May be I'm wrong :)

bilelmoussaoui: This is not a normal clone though. Clone can only be implemented if all the inner items implement the Clone traits. In this case, all the inner items are wrapped in Arc. So basically you will get the same thing without the extra Arc.

warusadura: Yes. Understood, thanks :)

warusadura: Thanks again for the explanation. Done.

server/src/service_manager.rs
```diff
@@ -1,24 +1,23 @@
...
-use crate::session::Session;
+use crate::Service;
 
 #[derive(Debug)]
 pub struct ServiceManager {
     connection: zbus::Connection,
-    // sessions mapped to their corresponding object path on the bus
-    sessions: HashMap<OwnedObjectPath, Arc<Session>>,
+    service: Option<Service>,
     session_index: Arc<RwLock<u32>>,
```
bilelmoussaoui, line 14: Do we know what else that could be added here to the service manager ? Because it feels like we can just completely get rid of it now.

warusadura: Ha ha :)

So, we're going to pass Service instances instead of ServiceManagers to the other parts right?

bilelmoussaoui: Yes

bilelmoussaoui (approved):

PR #173, opened 2024-12-09
Add prompt implementation for service methods lock & unlock

warusadura:
Please see the commits.

### DIFF
--- a/server/src/gnome/prompter.rs
+++ b/server/src/gnome/prompter.rs
@@ -1,0 +1,276 @@
+// org.gnome.keyring.Prompter
+// https://gitlab.gnome.org/GNOME/gcr/-/blob/main/gcr/org.gnome.keyring.Prompter.xml
+
+use clap::error::Result;
+use oo7::{dbus::ServiceError, Key};
+use serde::{Deserialize, Serialize};
+use tokio::sync::RwLock;
+use zbus::{
+    interface, proxy,
+    zvariant::{
+        DeserializeDict, NoneValue, Optional, OwnedObjectPath, OwnedValue, SerializeDict, Type,
+        Value,
+    },
+};
+
+use crate::{
+    gnome::secret_exchange,
+    prompt::{Prompt, PromptRole},
+    service::Service,
+};
+
+#[derive(Debug, DeserializeDict, SerializeDict, Type)]
+#[zvariant(signature = "dict")]
+// GcrPrompt properties: https://gitlab.gnome.org/GNOME/gcr/-/blob/main/gcr/gcr-prompt.c#L95
+pub struct Properties {
+    title: Option<String>,
+    message: Option<String>,
+    description: Option<String>,
+    warning: Option<String>,
+    #[zvariant(rename = "password-new")]
+    password_new: Option<bool>,
+    #[zvariant(rename = "password-strength")]
+    password_strength: Option<i32>,
+    #[zvariant(rename = "choice-label")]
+    choice_label: Option<String>,
+    #[zvariant(rename = "choice-chosen")]
+    choice_chosen: Option<bool>,
+    #[zvariant(rename = "caller-window")]
+    caller_window: Option<String>,
+    #[zvariant(rename = "continue-label")]
+    continue_label: Option<String>,
+    #[zvariant(rename = "cancel-label")]
+    cancel_label: Option<String>,
+}
+
+impl Properties {
+    fn for_lock(object: &str, window_id: &str) -> Self {
+        Self {
+            title: None,
+            message: Some("Lock Keyring".to_owned()),
+            description: Some(format!("Confirm locking '{}' Keyring", object)),
+            warning: None,
+            password_new: None,
+            password_strength: None,
+            choice_label: None,
+            choice_chosen: None,
+            caller_window: Some(window_id.to_owned()),
+            continue_label: Some("Lock".to_owned()),
+            cancel_label: Some("Cancel".to_owned()),
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, Type)]
+#[serde(rename_all = "lowercase")]
+#[zvariant(signature = "s")]
+// Possible values for PromptReady reply parameter
+pub enum Reply {
+    No,
+    Yes,
+}
+
+impl NoneValue for Reply {
+    type NoneType = String;
+
+    fn null_value() -> Self::NoneType {
+        String::new()
+    }
+}
+
+impl TryFrom<String> for Reply {
+    type Error = String;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        match value.as_str() {
+            "no" => Ok(Reply::No),
+            "yes" => Ok(Reply::Yes),
+            _ => Err("Invalid value".to_string()),
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, Type)]
+#[serde(rename_all = "lowercase")]
+#[zvariant(signature = "s")]
+// Possible values for PerformPrompt type parameter
+pub enum PromptType {
+    Confirm,
+    Password,
+}
+
+// org.gnome.keyring.internal.Prompter
+
+#[proxy(
+    default_service = "org.gnome.keyring.SystemPrompter",
+    interface = "org.gnome.keyring.internal.Prompter",
+    default_path = "/org/gnome/keyring/Prompter"
+)]
+pub trait Prompter {
+    fn begin_prompting(&self, callback: &OwnedObjectPath) -> Result<(), ServiceError>;
+
+    fn perform_prompt(
+        &self,
+        callback: OwnedObjectPath,
+        type_: PromptType,
+        properties: Properties,
+        exchange: &str,
+    ) -> Result<(), ServiceError>;
+
+    fn stop_prompting(&self, callback: OwnedObjectPath) -> Result<(), ServiceError>;
+}
+
+// org.gnome.keyring.internal.Prompter.Callback
+
+#[derive(Debug)]
+pub struct PrompterCallback {
+    window_id: String,
+    private_key: Key,
+    public_key: Key,
+    aes_key: RwLock<Option<Key>>,
+    exchange: RwLock<String>,
+    service: Service,
+    prompt_path: OwnedObjectPath,
+    path: OwnedObjectPath,
+}
+
+#[interface(name = "org.gnome.keyring.internal.Prompter.Callback")]
+impl PrompterCallback {
+    pub async fn prompt_ready(
+        &self,
+        reply: Optional<Reply>,
+        _properties: Properties,
+        exchange: &str,
+        #[zbus(connection)] connection: &zbus::Connection,
+    ) -> Result<(), ServiceError> {
+        let Some(prompt) = self.service.prompt(&self.prompt_path).await else {
+            return Err(ServiceError::NoSuchObject(
+                "Prompt does not exist.".to_string(),
+            ));
+        };
+
+        let prompter = PrompterProxy::new(connection).await?;
+        let path = self.path.clone();
+        let prompt_role = prompt.role();
+        let prompt_path = prompt.path().clone();
+        let prompt_objects = prompt.objects().to_owned();
+
+        match *reply {
+            Some(Reply::No) => {
+                // Second PromptReady call and the prompt is dismissed
+                tracing::debug!("Prompt is being dismissed.");
+
+                tokio::spawn(async move { prompter.stop_prompting(path).await });
+
+                let signal_emitter = self.service.signal_emitter(prompt_path)?;
+                let result = Value::new::<Vec<OwnedObjectPath>>(vec![])
+                    .try_to_owned()
+                    .unwrap();
+
+                tokio::spawn(async move { Prompt::completed(&signal_emitter, true, result).await });
+            }
+            Some(Reply::Yes) => {
+                // Second PromptReady call with the final exchange
+                match prompt_role {
+                    PromptRole::Lock => {
+                        let service = self.service.clone();
+                        let objects = prompt_objects.clone();
+
+                        tokio::spawn(async move {
+                            let _ = service.set_locked(true, &objects, true).await;
+                        });
+                    }
+                    PromptRole::Unlock => todo!(),
+                    PromptRole::CreateCollection => todo!(),
+                }
+
+                tokio::spawn(async move { prompter.stop_prompting(path).await });
+
+                let signal_emitter = self.service.signal_emitter(prompt_path)?;
+                let result = Value::new(prompt_objects).try_to_owned().unwrap();
+
+                tokio::spawn(async move {
+                    tracing::debug!("Prompt completed.");
+                    Prompt::completed(&signal_emitter, false, result).await
+                });
+            }
+            None => {
+                // First PromptReady call
+                let server_exchange = secret_exchange::begin(&self.public_key);
+                *self.exchange.write().await = server_exchange.clone();
+
+                let label = self
+                    .object_label(&prompt_objects)
+                    .await
+                    .unwrap_or(String::new());
+                let path = self.path.clone();
+
+                let (properties, prompt_type) = match prompt_role {
+                    PromptRole::Lock => (
+                        Properties::for_lock(&label, &self.window_id),
+                        PromptType::Confirm,
+                    ),
+                    PromptRole::Unlock => todo!(),
+                    PromptRole::CreateCollection => todo!(),
+                };
+
+                tokio::spawn(async move {
+                    prompter
+                        .perform_prompt(path, prompt_type, properties, &server_exchange)
+                        .await
+                });
+            }
+        }
+        Ok(())
+    }
+
+    pub async fn prompt_done(
+        &self,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<(), ServiceError> {
+        if let Some(prompt) = self.service.prompt(&self.prompt_path).await {
+            let path = prompt.path();
+            object_server.remove::<Prompt, _>(path).await?;
+            self.service.remove_prompt(path).await;
+        }
+        object_server.remove::<Self, _>(&self.path).await?;
+
+        Ok(())
+    }
+}
+
+impl PrompterCallback {
+    pub async fn new(
+        window_id: &str,
+        service: Service,
+        prompt_path: OwnedObjectPath,
+    ) -> Result<Self, oo7::crypto::Error> {
+        let index = service.prompt_index().await;
+        let private_key = Key::generate_private_key()?;
+        Ok(Self {
+            window_id: window_id.to_owned(),
+            public_key: crate::gnome::crypto::generate_public_key(&private_key)?,
+            aes_key: Default::default(),
+            exchange: Default::default(),
+            path: OwnedObjectPath::try_from(format!("/org/gnome/keyring/Prompt/p{index}")).unwrap(),
+            private_key,
+            service,
+            prompt_path,
+        })
+    }
+
+    pub fn path(&self) -> &OwnedObjectPath {
+        &self.path
+    }
+
+    pub async fn object_label(&self, objects: &Vec<OwnedObjectPath>) -> Option<String> {
+        for object in objects {
+            if let Some(collection) = self.service.collection_from_path(object).await {
+                return Some(collection.label().await);
+            } else if let Some(item) = self.service.item_from_path(object).await {
+                return Some(item.label().await);
+            }
+        }
+        None
+    }
+}
--- a/server/src/gnome/prompter.rs
+++ b/server/src/gnome/prompter.rs
@@ -7,10 +7,7 @@ use serde::{Deserialize, Serialize};
 use tokio::sync::RwLock;
 use zbus::{
     interface, proxy,
-    zvariant::{
-        DeserializeDict, NoneValue, Optional, OwnedObjectPath, OwnedValue, SerializeDict, Type,
-        Value,
-    },
+    zvariant::{DeserializeDict, NoneValue, Optional, OwnedObjectPath, SerializeDict, Type, Value},
 };
 
 use crate::{
@@ -56,6 +53,25 @@ impl Properties {
             choice_chosen: None,
             caller_window: Some(window_id.to_owned()),
             continue_label: Some("Lock".to_owned()),
+            cancel_label: Some("Cancel".to_owned()),
+        }
+    }
+
+    fn for_unlock(object: &str, warning: &str, window_id: &str) -> Self {
+        Self {
+            title: Some("Unlock Keyring".to_owned()),
+            message: Some("Authentication required".to_owned()),
+            description: Some(format!(
+                "An application wants access to the keyring '{}', but it is locked",
+                object
+            )),
+            warning: Some(warning.to_owned()),
+            password_new: None,
+            password_strength: None,
+            choice_label: None,
+            choice_chosen: None,
+            caller_window: Some(window_id.to_owned()),
+            continue_label: Some("Unlock".to_owned()),
             cancel_label: Some("Cancel".to_owned()),
         }
     }
@@ -171,6 +187,11 @@ impl PrompterCallback {
             }
             Some(Reply::Yes) => {
                 // Second PromptReady call with the final exchange
+                let label = self
+                    .object_label(&prompt_objects)
+                    .await
+                    .unwrap_or(String::new());
+
                 match prompt_role {
                     PromptRole::Lock => {
                         let service = self.service.clone();
@@ -180,7 +201,61 @@ impl PrompterCallback {
                             let _ = service.set_locked(true, &objects, true).await;
                         });
                     }
-                    PromptRole::Unlock => todo!(),
+                    PromptRole::Unlock => {
+                        let Some(aes_key) = &*self.aes_key.read().await else {
+                            return Err(ServiceError::ZBus(zbus::Error::FDO(Box::new(
+                                zbus::fdo::Error::Failed(
+                                    "Failed to retrieve AES key for SecretExchange.".to_string(),
+                                ),
+                            ))));
+                        };
+
+                        // Verify the secret
+                        if let Some(secret) = secret_exchange::retrieve(exchange, aes_key) {
+                            match oo7::file::Keyring::open(&label, secret).await {
+                                Ok(_) => {
+                                    tracing::debug!("{} keyring secret matches.", &label);
+                                }
+                                Err(oo7::file::Error::IncorrectSecret) => {
+                                    tracing::error!("{} keyring incorrect secret.", &label);
+
+                                    let properties = Properties::for_unlock(
+                                        &label,
+                                        "The unlock password was incorrect",
+                                        &self.window_id,
+                                    );
+                                    let server_exchange = self.exchange.read().await.clone();
+
+                                    tokio::spawn(async move {
+                                        prompter
+                                            .perform_prompt(
+                                                path,
+                                                PromptType::Password,
+                                                properties,
+                                                &server_exchange,
+                                            )
+                                            .await
+                                    });
+
+                                    return Ok(());
+                                }
+                                Err(err) => {
+                                    return Err(ServiceError::ZBus(zbus::Error::FDO(Box::new(
+                                        zbus::fdo::Error::Failed(format!(
+                                            "Failed to unlock {label} Keyring: {err}."
+                                        )),
+                                    ))))
+                                }
+                            }
+                        }
+
+                        let service = self.service.clone();
+                        let objects = prompt_objects.clone();
+
+                        tokio::spawn(async move {
+                            let _ = service.set_locked(false, &objects, true).await;
+                        });
+                    }
                     PromptRole::CreateCollection => todo!(),
                 }
 
@@ -210,7 +285,22 @@ impl PrompterCallback {
                         Properties::for_lock(&label, &self.window_id),
                         PromptType::Confirm,
                     ),
-                    PromptRole::Unlock => todo!(),
+                    PromptRole::Unlock => {
+                        let aes_key = secret_exchange::handshake(&self.private_key, exchange)
+                            .map_err(|err| {
+                                ServiceError::ZBus(zbus::Error::FDO(Box::new(
+                                    zbus::fdo::Error::Failed(format!(
+                                        "Failed to generate AES key for SecretExchange {err}."
+                                    )),
+                                )))
+                            })?;
+                        *self.aes_key.write().await = Some(aes_key);
+
+                        (
+                            Properties::for_unlock(&label, "", &self.window_id),
+                            PromptType::Password,
+                        )
+                    }
                     PromptRole::CreateCollection => todo!(),
                 };
 
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -1,6 +1,5 @@
 mod collection;
 mod error;
-#[allow(unused)]
 mod gnome;
 mod item;
 mod prompt;
--- a/server/src/prompt.rs
+++ b/server/src/prompt.rs
@@ -1,25 +1,104 @@
 // org.freedesktop.Secret.Prompt
 
 use oo7::dbus::ServiceError;
-use zbus::{interface, object_server::SignalEmitter, zvariant::OwnedValue};
+use zbus::{
+    interface,
+    object_server::SignalEmitter,
+    zvariant::{OwnedObjectPath, OwnedValue},
+};
 
-#[derive(Debug)]
-pub struct Prompt {}
+use crate::{
+    gnome::prompter::{PrompterCallback, PrompterProxy},
+    service::Service,
+};
+
+#[derive(Debug, Clone)]
+#[allow(unused)]
+pub enum PromptRole {
+    Lock,
+    Unlock,
+    CreateCollection,
+}
+
+#[derive(Debug, Clone)]
+pub struct Prompt {
+    service: Service,
+    objects: Vec<OwnedObjectPath>,
+    role: PromptRole,
+    path: OwnedObjectPath,
+}
 
 #[interface(name = "org.freedesktop.Secret.Prompt")]
 impl Prompt {
-    pub async fn prompt(&self, _window_id: &str) -> Result<(), ServiceError> {
-        todo!()
+    pub async fn prompt(
+        &self,
+        // TODO: Replace with ashpd::WindowIdentifierType
+        window_id: &str,
+        #[zbus(connection)] connection: &zbus::Connection,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<(), ServiceError> {
+        let callback = PrompterCallback::new(window_id, self.service.clone(), self.path.clone())
+            .await
+            .map_err(|err| {
+                ServiceError::ZBus(zbus::Error::FDO(Box::new(zbus::fdo::Error::Failed(
+                    format!("Failed to create PrompterCallback {err}."),
+                ))))
+            })?;
+
+        let path = callback.path().clone();
+        let connection = connection.clone();
+
+        object_server.at(&path, callback).await?;
+        tracing::debug!("Prompt `{}` created.", self.path);
+
+        // Starts GNOME System Prompting.
+        // Spawned separately to avoid blocking the early return of the current
+        // execution.
+        let prompter = PrompterProxy::new(&connection).await?;
+        tokio::spawn(async move { prompter.begin_prompting(&path).await });
+
+        Ok(())
     }
 
-    pub async fn dismiss(&self) -> Result<(), ServiceError> {
-        todo!()
+    pub async fn dismiss(
+        &self,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<(), ServiceError> {
+        object_server.remove::<Self, _>(&self.path).await?;
+        self.service.remove_prompt(&self.path).await;
+
+        Ok(())
     }
 
     #[zbus(signal, name = "Completed")]
-    async fn completed(
+    pub async fn completed(
         signal_emitter: &SignalEmitter<'_>,
         dismissed: bool,
         result: OwnedValue,
     ) -> zbus::Result<()>;
+}
+
+impl Prompt {
+    pub async fn new(service: Service, objects: Vec<OwnedObjectPath>, role: PromptRole) -> Self {
+        let index = service.prompt_index().await;
+        Self {
+            path: OwnedObjectPath::try_from(format!("/org/freedesktop/secrets/prompt/p{index}"))
+                .unwrap(),
+            service,
+            objects,
+            role,
+        }
+    }
+
+    pub fn path(&self) -> &OwnedObjectPath {
+        &self.path
+    }
+
+    pub fn role(&self) -> PromptRole {
+        self.role.clone()
+    }
+
+    pub fn objects(&self) -> &[OwnedObjectPath] {
+        &self.objects
+    }
 }
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -18,7 +18,13 @@ use zbus::{
     zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Value},
 };
 
-use crate::{collection::Collection, error::Error, session::Session};
+use crate::{
+    collection::Collection,
+    error::Error,
+    item,
+    prompt::{Prompt, PromptRole},
+    session::Session,
+};
 
 #[derive(Debug, Clone)]
 pub struct Service {
@@ -29,6 +35,8 @@ pub struct Service {
     // sessions mapped to their corresponding object path on the bus
     sessions: Arc<Mutex<HashMap<OwnedObjectPath, Session>>>,
     session_index: Arc<RwLock<u32>>,
+    prompts: Arc<Mutex<HashMap<OwnedObjectPath, Prompt>>>,
+    prompt_index: Arc<RwLock<u32>>,
 }
 
 #[zbus::interface(name = "org.freedesktop.Secret.Service")]
@@ -130,8 +138,20 @@ impl Service {
     pub async fn unlock(
         &self,
         objects: Vec<OwnedObjectPath>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
-        let (unlocked, _not_unlocked) = self.set_locked(false, &objects).await?;
+        let (unlocked, not_unlocked) = self.set_locked(false, &objects, false).await?;
+        if !not_unlocked.is_empty() {
+            let prompt = Prompt::new(self.clone(), not_unlocked, PromptRole::Unlock).await;
+            let path = prompt.path().clone();
+            self.prompts
+                .lock()
+                .await
+                .insert(path.clone(), prompt.clone());
+
+            object_server.at(&path, prompt).await?;
+            return Ok((unlocked, path));
+        }
 
         Ok((unlocked, OwnedObjectPath::default()))
     }
@@ -140,8 +160,20 @@ impl Service {
     pub async fn lock(
         &self,
         objects: Vec<OwnedObjectPath>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
-        let (locked, _not_locked) = self.set_locked(true, &objects).await?;
+        let (locked, not_locked) = self.set_locked(true, &objects, false).await?;
+        if !not_locked.is_empty() {
+            let prompt = Prompt::new(self.clone(), not_locked, PromptRole::Lock).await;
+            let path = prompt.path().clone();
+            self.prompts
+                .lock()
+                .await
+                .insert(path.clone(), prompt.clone());
+
+            object_server.at(&path, prompt).await?;
+            return Ok((locked, path));
+        }
 
         Ok((locked, OwnedObjectPath::default()))
     }
@@ -273,6 +305,8 @@ impl Service {
             connection: connection.clone(),
             sessions: Default::default(),
             session_index: Default::default(),
+            prompts: Default::default(),
+            prompt_index: Default::default(),
         };
 
         object_server
@@ -318,6 +352,8 @@ impl Service {
         &self,
         locked: bool,
         objects: &[OwnedObjectPath],
+        // Whether the method is called from a prompt
+        has_prompt: bool,
     ) -> Result<(Vec<OwnedObjectPath>, Vec<OwnedObjectPath>), ServiceError> {
         let mut without_prompt = Vec::new();
         let mut with_prompt = Vec::new();
@@ -335,8 +371,9 @@ impl Service {
                         );
                         without_prompt.push(object.clone());
                     } else {
-                        // TODO: remove this once the prompt implementation is complete.
-                        collection.set_locked(locked).await?;
+                        if has_prompt {
+                            collection.set_locked(locked).await?;
+                        }
                         with_prompt.push(object.clone());
                     }
                     break;
@@ -354,8 +391,9 @@ impl Service {
                         item.set_locked(locked).await?;
                         without_prompt.push(object.clone());
                     } else {
-                        // TODO: remove this once the prompt implementation is complete.
-                        item.set_locked(locked).await?;
+                        if has_prompt {
+                            item.set_locked(locked).await?;
+                        }
                         with_prompt.push(object.clone());
                     }
                     break;
@@ -377,6 +415,18 @@ impl Service {
         for collection in collections.iter() {
             if collection.path() == path {
                 return Some(collection.clone());
+            }
+        }
+
+        None
+    }
+
+    pub async fn item_from_path(&self, path: &OwnedObjectPath) -> Option<item::Item> {
+        let collections = self.collections.lock().await;
+
+        for collection in collections.iter() {
+            if let Some(item) = collection.item_from_path(path).await {
+                return Some(item);
             }
         }
 
@@ -409,5 +459,20 @@ impl Service {
         *self.session_index.write().await = n_sessions;
 
         n_sessions
+    }
+
+    pub async fn prompt_index(&self) -> u32 {
+        let n_prompts = *self.prompt_index.read().await + 1;
+        *self.prompt_index.write().await = n_prompts;
+
+        n_prompts
+    }
+
+    pub async fn prompt(&self, path: &OwnedObjectPath) -> Option<Prompt> {
+        self.prompts.lock().await.get(path).cloned()
+    }
+
+    pub async fn remove_prompt(&self, path: &OwnedObjectPath) {
+        self.prompts.lock().await.remove(path);
     }
 }
--- a/server/src/gnome/mod.rs
+++ b/server/src/gnome/mod.rs
@@ -1,2 +1,3 @@
 pub mod crypto;
+pub mod prompter;
 pub mod secret_exchange;


client/src/crypto/native.rs
```diff
@@ -81,6 +81,17 @@ pub(crate) fn generate_public_key(private_key: impl AsRef<[u8]>) -> Vec<u8> {
     public_key_uint.to_bytes_be()
 }
 
+pub(crate) fn generate_public_key_for_secret_exchange(private_key: impl AsRef<[u8]>) -> Vec<u8> {
+    let private_key_uint = BigUint::from_bytes_be(private_key.as_ref());
+    let dh_generator = {
```
bilelmoussaoui, line 86: Should use LazyLock nowadays.

client/src/crypto/native.rs
```diff
@@ -110,6 +121,35 @@ pub(crate) fn generate_aes_key(
...
+    private_key: impl AsRef<[u8]>,
+    server_public_key: impl AsRef<[u8]>,
+) -> Zeroizing<Vec<u8>> {
+    let server_public_key_uint = BigUint::from_bytes_be(server_public_key.as_ref());
+    let private_key_uint = BigUint::from_bytes_be(private_key.as_ref());
+    let common_secret = powm_for_secret_exchange(&server_public_key_uint, private_key_uint);
+
+    let mut common_secret_bytes = common_secret.to_bytes_be();
+    let mut common_secret_padded = vec![0; 192 - common_secret_bytes.len()];
+    // inefficient, but ok for now
```
bilelmoussaoui, line 134: what is inefficient about it

client/src/crypto/native.rs
```diff
@@ -233,3 +273,42 @@ fn powm(base: &BigUint, mut exp: BigUint) -> BigUint {
 
     result
 }
+
+fn powm_for_secret_exchange(base: &BigUint, mut exp: BigUint) -> BigUint {
+    let modulus = {
+        // for key exchange
+        static DH_PRIME: OnceLock<BigUint> = OnceLock::new();
```
bilelmoussaoui, line 280: LazyLock

client/src/crypto/openssl.rs
```diff
@@ -94,6 +94,19 @@ pub(crate) fn generate_public_key(private_key: impl AsRef<[u8]>) -> Vec<u8> {
...
 
+pub(crate) fn generate_public_key_for_secret_exchange(private_key: impl AsRef<[u8]>) -> Vec<u8> {
+    let private_key_bn = BigNum::from_slice(private_key.as_ref()).unwrap();
+    Dh::from_pqg(
+        BigNum::get_rfc3526_prime_1536().unwrap(),
+        None,
+        BigNum::from_u32(2).unwrap(),
+    )
+    .and_then(|key| key.set_private_key(private_key_bn))
+    .unwrap()
```
bilelmoussaoui, line 105: See my latest PR, don't unwrap here

client/src/crypto/openssl.rs
```diff
@@ -127,6 +140,39 @@ pub(crate) fn generate_aes_key(
...
+    server_public_key: impl AsRef<[u8]>,
+) -> Zeroizing<Vec<u8>> {
+    let private_key_bn = BigNum::from_slice(private_key.as_ref()).unwrap();
+    let server_public_key_bn = BigNum::from_slice(server_public_key.as_ref()).unwrap();
+    let mut common_secret_bytes = Dh::from_pqg(
+        BigNum::get_rfc3526_prime_1536().unwrap(),
+        None,
+        BigNum::from_u32(2).unwrap(),
+    )
+    .and_then(|key| key.set_private_key(private_key_bn))
```
bilelmoussaoui, line 154: Same here

client/src/crypto/openssl.rs
```diff
@@ -127,6 +140,39 @@ pub(crate) fn generate_aes_key(
...
+    .and_then(|key| key.compute_key(&server_public_key_bn))
+    .unwrap();
+
+    let mut common_secret_padded = vec![0; 192 - common_secret_bytes.len()];
+    // inefficient, but ok for now
```
bilelmoussaoui, line 159: Same here

server/Cargo.toml
```diff
@@ -9,6 +9,7 @@ rust-version.workspace = true
 version.workspace = true
 
 [dependencies]
+base64 = "0.22.1"
```
bilelmoussaoui, line 12:
```suggestion
base64 = "0.22"
```

server/src/main.rs
```diff
@@ -2,6 +2,9 @@ mod collection;
 mod error;
 mod item;
 mod prompt;
+mod prompter;
+mod prompter_callback;
+mod secret_exchange;
```
bilelmoussaoui, line 7: Those two seems to be specific to the gnome prompting mechanism, so probably should go under a gnome sub-module.

server/src/prompt.rs
```diff
@@ -1,19 +1,40 @@
...
+    path: OwnedObjectPath,
+}
 
 #[interface(name = "org.freedesktop.Secret.Prompt")]
 impl Prompt {
-    pub async fn prompt(&self, _window_id: &str) -> Result<(), ServiceError> {
-        todo!()
+    pub async fn prompt(
+        &self,
+        _window_id: &str,
```
bilelmoussaoui, line 22: This should use a ashpd::WindowIdentifierType here

bilelmoussaoui: i know it is not used by the gnome implementation but it will eventually used by something better.

server/src/prompter.rs
```diff
@@ -0,0 +1,29 @@
...
+    interface = "org.gnome.keyring.internal.Prompter",
+    default_path = "/org/gnome/keyring/Prompter"
+)]
+pub trait Prompter {
+    fn begin_prompting(&self, callback: OwnedObjectPath) -> Result<(), ServiceError>;
+
+    fn perform_prompt(
+        &self,
+        callback: OwnedObjectPath,
+        type_: &str, // 'password' or 'confirm', put this in Enum?
```
bilelmoussaoui, line 23: Yes, this has to be turned into an enum to avoid random strings being passed.

server/src/prompter.rs
```diff
@@ -0,0 +1,29 @@
...
+        properties: HashMap<&str, OwnedValue>,
```
bilelmoussaoui, line 24: I assume the properties are fixed/known dictionnaries, if so, they must be defined as structs.

server/src/prompter.rs
```diff
@@ -0,0 +1,29 @@
...
+        exchange: &str,
```
bilelmoussaoui, line 25: What is the content of this? is it really a string?

server/src/prompter.rs
```diff
@@ -0,0 +1,29 @@
...
+use std::collections::HashMap;
+
+use oo7::dbus::ServiceError;
+use zbus::{
+    proxy,
+    zvariant::{OwnedObjectPath, OwnedValue},
+};
+
+#[proxy(
+    default_service = "org.gnome.keyring.SystemPrompter",
```
bilelmoussaoui, line 13: gnome specific -> gnome module

server/src/prompter_callback.rs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    pub async fn prompt_done(&self) -> Result<(), ServiceError> {
+        Ok(())
+    }
+}
+
+impl PrompterCallback {
+    pub async fn new(service: Service) -> Self {
+        let index = service.session_index().await;
```
bilelmoussaoui, line 37: session or prompt_index?

server/src/secret_exchange.rs
```diff
@@ -0,0 +1,131 @@
+// SecretExchange: Exchange secrets between processes in an unexposed way.
+
+// Initial C implementation: https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-secret-exchange.c
+
+// The initial implementation of SecretExchange/GCRSecretExchange uses a KeyFile
+// to encode/parse the payload. And in this implementation the payload is based
+// on a HashMap.
```
bilelmoussaoui, line 7:
```suggestion
// The initial implementation of SecretExchange/GCRSecretExchange uses a KeyFile
// to encode/parse the payload. In this implementation the payload is based
// on a HashMap.
```

server/src/secret_exchange.rs
```diff
@@ -0,0 +1,131 @@
...
+
+const SECRET: &str = "secret";
+const PUBLIC: &str = "public";
+const PRIVATE: &str = "private";
+const IV: &str = "iv";
+const PROTOCOL: &str = "[sx-aes-1]\n";
+const CIPHER_TEXT_LEN: usize = 16;
+
+#[derive(Debug)]
+pub struct SecretExchange {
```
bilelmoussaoui, line 25: Zeroize the type.

server/src/secret_exchange.rs
```diff
@@ -0,0 +1,131 @@
...
+    // Creates the initial payload containing caller public_key
+    pub fn begin(&self) -> String {
+        let map = HashMap::from([(PUBLIC, self.public_key.as_ref())]);
+
+        encode(&map)
+    }
+
+    // Creates the shared secret: an AES key
+    pub fn create_shared_secret(&self, exchange: &str) -> String {
+        let decoded = decode(exchange).unwrap();
```
bilelmoussaoui, line 40: Avoid unwrap, use expect if the error is never supposed to happen with a comment why it is not

server/src/secret_exchange.rs
```diff
@@ -0,0 +1,131 @@
...
+// Retrieves the secret from final secret exchange string
+pub(crate) fn retrieve_secret(exchange: &str, aes_key: &str) -> Option<Zeroizing<Vec<u8>>> {
+    let decoded = decode(exchange).unwrap();
+    let secret = decoded.get(SECRET);
+
+    // If we cancel an ongoing prompt call, the final exchange won't have the secret
+    // or IV. The following is to avoid `Option::unwrap()` on a `None` value
+    secret?;
+
+    let secret = secret.unwrap();
```
bilelmoussaoui, line 100:
```suggestion
    // If we cancel an ongoing prompt call, the final exchange won't have the secret
    // or IV. The following is to avoid `Option::unwrap()` on a `None` value
    let secret = decoded.get(SECRET)?;

```

server/src/secret_exchange.rs
```diff
@@ -0,0 +1,131 @@
...
+    if secret.len() != CIPHER_TEXT_LEN {
+        // to avoid a short secret/cipher-text causing an UnpadError during decryption
+        let false_secret: Vec<u8> = vec![0, 1];
+        return Some(Zeroizing::new(false_secret));
```
bilelmoussaoui, line 104:
```suggestion
        let false_secret = vec![0, 1];
        return Some(Zeroizing::new(false_secret));
```

server/src/secret_exchange.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    let iv = decoded.get(IV).unwrap();
+    let decoded = decode(aes_key).unwrap();
+    let aes_key = Key::new(decoded.get(PRIVATE).unwrap().to_vec());
```
bilelmoussaoui, line 109:
```suggestion
    let iv = decoded.get(IV)?;
    let decoded = decode(aes_key)?;
    let aes_key = Key::new(decoded.get(PRIVATE)?.to_vec());
```

server/src/secret_exchange.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    Some(crypto::decrypt(secret, &aes_key, iv))
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_secret_exchange() {
```
bilelmoussaoui, line 119: should the opposite also be tested?

server/src/service.rs
```diff
@@ -114,8 +115,16 @@ impl Service {
     pub async fn unlock(
         &self,
         objects: Vec<OwnedObjectPath>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
-        let (unlocked, _not_unlocked) = self.set_locked(false, &objects).await?;
+        let (unlocked, not_unlocked) = self.set_locked(false, &objects).await?;
+        if !not_unlocked.is_empty() {
+            let prompt = Prompt::new(self.clone()).await;
```
bilelmoussaoui, line 122: The prompt also needs a way to pass it a future that would be executed when the prompt is done

bilelmoussaoui (commented):
also needs a rebase

warusadura:
Thanks for the prompt review @bilelmoussaoui

client/src/crypto/native.rs
```diff
@@ -1,6 +1,6 @@
 use std::{
     ops::{Mul, Rem, Shr},
-    sync::LazyLock,
+    sync::{LazyLock, OnceLock},
```
bilelmoussaoui, line 3: This lacks an openssl implementation as well as that is what most distros will likely use the daemon with

warusadura: Will do, thanks!

client/src/crypto/native.rs
warusadura: Thanks, done.

client/src/crypto/openssl.rs
warusadura: Implemented, thanks!

server/Cargo.toml
warusadura: Done.

server/src/main.rs

server/src/secret_exchange.rs
warusadura: Yeah, thanks!

server/src/secret_exchange.rs
warusadura: Done.

warusadura:
Sorry for the delay @bilelmoussaoui 

I have removed the prompt implementation for now. Just wanted get the crypto part and secret_exchange impl cleared before moving into the Prompt impl. I have removed a lot of unwraps, but I'm still not sure this is any good. Very much like to hear what you think of the current impl :)

server/src/gnome/secret_exchange.rs
```diff
@@ -0,0 +1,174 @@
...
+            return None;
+        }
+        map.insert(key, encoded);
+    }
+
+    Some(map)
+}
+
+// Retrieves the secret from final secret exchange string
+pub(crate) fn retrieve_secret(exchange: &str, aes_key: &str) -> Option<Zeroizing<Vec<u8>>> {
```
bilelmoussaoui, line 102: This should probably return a oo7::Secret.

server/src/gnome/secret_exchange.rs
```diff
@@ -0,0 +1,174 @@
...
+
+const SECRET: &str = "secret";
+const PUBLIC: &str = "public";
+const PRIVATE: &str = "private";
+const IV: &str = "iv";
+const PROTOCOL: &str = "[sx-aes-1]\n";
+const CIPHER_TEXT_LEN: usize = 16;
+
+#[derive(Debug, Zeroize)]
+pub struct SecretExchange {
```
bilelmoussaoui, line 25: I am still not sure how the SecretExchange is supposed to be used. The current API, even if used internally only, is not really nice. why there is a need for a SecretExchange struct at all? who would call begin/create_shared_secret? and what about retrieve_secret? I would need to see how it is used in a real-case scenario to provide any meaningful feedback

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,205 @@
...
+    default_path = "/org/gnome/keyring/Prompter"
+)]
+pub trait Prompter {
+    fn begin_prompting(&self, callback: &OwnedObjectPath) -> Result<(), ServiceError>;
+
+    fn perform_prompt(
+        &self,
+        callback: OwnedObjectPath,
+        type_: &str,
+        properties: HashMap<&str, OwnedValue>,
```
bilelmoussaoui, line 33: Type safety, like mentioned before.

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct PrompterCallback {
+    service: Service,
+    path: OwnedObjectPath,
+}
+
+#[interface(name = "org.gnome.keyring.internal.Prompter.Callback")]
+impl PrompterCallback {
+    pub async fn prompt_ready(
+        &self,
+        reply: &str,
```
bilelmoussaoui, line 51: What the reply contains?

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,205 @@
...
+        properties: HashMap<&str, OwnedValue>,
```
bilelmoussaoui, line 52: Type safety, like above

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,205 @@
...
+            return Err(ServiceError::NoSuchObject(
+                "Prompt does not exist.".to_string(),
+            ));
+        };
+
+        match prompt.role() {
+            PromptRole::Lock => {
+                if properties.is_empty() {
+                    // First PromptReady call
+                    let mut properties: HashMap<&str, zvariant::OwnedValue> = HashMap::new();
```
bilelmoussaoui, line 66: Same as above

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+        object_server.remove::<Self, _>(&self.path).await?;
+
+        Ok(())
+    }
+}
+
+impl PrompterCallback {
+    pub async fn new(service: Service) -> Self {
+        let index = service.prompt_index().await;
```
bilelmoussaoui, line 159: You should create a PrompterProxy here and keepe it around for the functions below.

server/src/prompt.rs
```diff
@@ -1,25 +1,92 @@
...
-pub struct Prompt {}
+use crate::{
+    gnome::prompter::{PrompterCallback, PrompterProxy},
+    service::Service,
+};
+
+#[derive(Debug, Clone)]
+pub enum PromptRole {
+    Lock,
+}
```
bilelmoussaoui, line 18: What else this enum might contain? fill it with other types and add a todo! for the match rule

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,205 @@
...
+                    ));
+                } else {
+                    // Second PromptReady call with the final exchange
+                    let service = self.service.clone();
+                    let objects = prompt.objects().clone();
+                    let result = Value::new(&objects).try_to_owned().unwrap();
+
+                    tokio::spawn(async move {
+                        let _ = service.set_locked(true, &objects, true).await;
+                    });
```
bilelmoussaoui, line 122: Instead of passing the objects, this should pass a future directly that you can spawn here. Keeping the logic that is supposed to be in Service::set_locked, stay there.

server/src/service.rs
```diff
@@ -410,4 +433,23 @@ impl Service {
...
+        let prompts = self.prompts.lock().await;
+
+        prompts.last().cloned()
+    }
+
+    pub async fn remove_prompt(&self) {
+        // prompts should always contain one item during a prompt related operation and
+        // it should be cleaned up afterwards.
+        self.prompts.lock().await.pop();
+    }
```
bilelmoussaoui, line 454: If there would always be one single prompt (which i don't believe is true, except for maybe GNOME specifically) as the API clearly allows modal keyring dialogs so every app could have an ongoing prompt operation.

server/src/service.rs
```diff
@@ -318,6 +336,7 @@ impl Service {
         &self,
         locked: bool,
         objects: &[OwnedObjectPath],
+        prompt: bool,
```
bilelmoussaoui, line 339: why is this needed?

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,205 @@
...
+                    );
+                    properties.insert("caller-window", Value::new("").try_to_owned().unwrap());
+                    properties.insert("cancel-label", Value::new("Cancel").try_to_owned().unwrap());
+
+                    let connection = connection.clone();
+                    let path = self.path.clone();
+
+                    tokio::spawn(PrompterCallback::perform_prompt(
+                        connection, path, "confirm", properties, exchange,
+                    ));
```
bilelmoussaoui, line 94: Not sure why this has to be in PrompterCallback,

server/src/gnome/prompter.rs
warusadura: Done, thanks!
Quick question: why using a HashMaps isn't type safe? I couldn't find an answer for this from the zbus book.

server/src/gnome/prompter.rs
warusadura: Good idea. But I can't access `self` inside `tokio::spawn`. All these following functions are associated functions.

server/src/prompt.rs
warusadura: Done, thanks!

server/src/gnome/secret_exchange.rs
warusadura: Yeah, it should. Done.

server/src/prompter_callback.rs
warusadura: Fixed this, thanks!

server/src/prompter.rs
warusadura: Done.

server/src/gnome/prompter.rs
warusadura: 'yes' or 'no'.

bilelmoussaoui: Make it an enum then.

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,226 @@
...
+                if properties.is_empty() {
+                    // First PromptReady call
+                    let secret_exchange = SecretExchange::new().map_err(|err| {
+                        ServiceError::ZBus(zbus::Error::FDO(Box::new(zbus::fdo::Error::Failed(
+                            format!("Failed to generate SecretExchange {err}."),
+                        ))))
+                    })?;
+                    let exchange = secret_exchange.begin();
+
+                    let properties = Properties {
```
bilelmoussaoui, line 95: As those properties are static, move them to new constructors in Properties impl. Like, for_lock, for_unlock and so on.

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,226 @@
...
+                        title: "".to_string(),
+                        choice_label: "".to_string(),
+                        description: "Confirm locking \"login\", Keyring.".to_string(),
+                        message: "Lock Keyring".to_string(),
+                        caller_window: "".to_string(),
+                        warning: "".to_string(),
```
bilelmoussaoui, line 101: Optional ones should be `Option<String>`.

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,226 @@
...
+        let Some(prompt) = self.service.prompt().await else {
+            return Err(ServiceError::NoSuchObject(
+                "Prompt does not exist.".to_string(),
+            ));
+        };
+
+        match prompt.role() {
+            PromptRole::Lock => {
+                if properties.is_empty() {
+                    // First PromptReady call
```
bilelmoussaoui, line 87: Shouldn't most of the prompting code be exactly the same no matter the PromptRole? the only thing that i see being different is the Properties and the future to execute once the prompt is done

server/src/gnome/prompter.rs
bilelmoussaoui: This was marked as resolved yet I still see properties being `HashMap<&str, Value>` instead of the newly added struct

bilelmoussaoui: also the properties passed are never used, how come?

server/src/gnome/prompter.rs
bilelmoussaoui: What i meant here is that you can pass an arbitrary dict when in reality it is a struct but as the only way to pass that over dbus is to use a dict, we still shouldn't use a dict from Rust.

warusadura: Understood, thanks and sorry for the delay.

server/src/gnome/prompter.rs
warusadura: Done, thanks!

server/src/gnome/prompter.rs
warusadura: Done, thanks!

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,287 @@
...
+use crate::{
+    gnome::secret_exchange::SecretExchange,
+    prompt::{Prompt, PromptRole},
+    service::Service,
+};
+
+#[derive(Debug, DeserializeDict, SerializeDict, Type)]
+#[zvariant(signature = "a{sv}")]
+// System prompt properties
+pub struct Properties {
```
warusadura, line 21: With this approach, during the second PromptReady call return I'm getting the following error, `Unexpected non-0 padding byte 1`. I can't figure out why. I think this is related to [this](https://gitlab.freedesktop.org/dbus/zbus/-/issues/82). Then again all the fields here are properly renamed using `#[zvariant(rename = "")`

server/src/prompter.rs
warusadura: Done.

server/src/prompter.rs
warusadura: Yes. The content:
```
'[sx-aes-1]\npublic=GfG2V4/Q/IgZpA9mV3QLnmFiUBdHSwIaNkZJpn//vuQLn1P2ujRT03ZBOJpKZsq/IYnWbTDfuRDue+kq7cWlp6KEdsv+6ZM57kWE7V4UnBcEtiuZB/DwuGmVL1JaKdqWbs9AVN3D+l6ZFrYek0WZM2YcOiGWUUD7TzhUai6RFVkgKC+KC3BQpldDefZCb5jxiYldC6gRy3+AQBnkpRqlT0eeh/9ecL4FieiRhfYJpUmooMUQY9Y7fFycVbl5aGw5\n'
```

server/src/gnome/prompter.rs
bilelmoussaoui: Check what gets send over the bus with busctl and see what is going on? Probably a mismatched type

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,287 @@
...
+    title: Option<String>,
+    #[zvariant(rename = "choice-label")]
+    choice_label: Option<String>,
+    description: Option<String>,
+    message: Option<String>,
+    #[zvariant(rename = "caller-window")]
+    caller_window: Option<String>,
```
bilelmoussaoui, line 28: This should be a WindowIdentifierType from ashpd

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,287 @@
...
+    cancel_label: Option<String>,
+}
+
+#[derive(Debug, Type)]
+#[zvariant(signature = "s")]
+// Possible values for PromptReady reply parameter
+pub enum Reply {
+    No,
+    Yes,
+    Empty,
```
bilelmoussaoui, line 48: What does empty corresponds to? Please link in the code to where those answers are defined

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+const NO: &str = "no";
```
bilelmoussaoui, line 51: No need for those consts

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,287 @@
...
+            EMPTY => Ok(Self::Empty),
+            err => Err(serde::de::Error::custom(format!("Invalid reply {err}"))),
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, Type)]
+#[serde(rename_all = "lowercase")]
+#[zvariant(signature = "s")]
+// Possible values for PerformPrompt type parameter
```
bilelmoussaoui, line 85: Should this implement the serialisation traits or the derived ones are correct?

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,287 @@
...
+    service: Service,
+    path: OwnedObjectPath,
+}
+
+#[interface(name = "org.gnome.keyring.internal.Prompter.Callback")]
+impl PrompterCallback {
+    pub async fn prompt_ready(
+        &self,
+        reply: Reply,
+        _properties: Properties,
```
bilelmoussaoui, line 124: Why is this unused? Already mentioned it before but you marked the thread as resolved except not really...

server/src/prompt.rs
```diff
@@ -1,25 +1,95 @@
...
 impl Prompt {
-    pub async fn prompt(&self, _window_id: &str) -> Result<(), ServiceError> {
-        todo!()
+    pub async fn prompt(
+        &self,
+        _window_id: &str,
+        #[zbus(connection)] connection: &zbus::Connection,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
+    ) -> Result<(), ServiceError> {
+        let callback = PrompterCallback::new(self.service.clone()).await;
```
bilelmoussaoui, line 39: The prompt callback creates the prompt path again, just pass it here instead

server/src/gnome/prompter.rs
warusadura: Will do, thanks.

server/src/gnome/prompter.rs
warusadura: Yes, the derived ones are correct. I had to implement the serialisation and de-serialisation for `Reply` because of the Empty variant.

server/src/gnome/prompter.rs
warusadura: Done, thanks!

server/src/gnome/prompter.rs
warusadura: Don't know anything about ashpd. I'll look into it :)

server/src/gnome/prompter.rs
warusadura: I'll take another look.

server/src/gnome/prompter.rs
warusadura: Empty corresponds to: `''`. The first PromptReady call's `reply` argument is always `''` and also the `properties` dict is empty.

image

server/src/gnome/prompter.rs
warusadura: I think this is correct. I have just pushed the unlock commit (wip) and the code looks mostly the same with a couple of changes.

server/src/gnome/prompter.rs
bilelmoussaoui: Check if you can use the zvariant::Optional type here and drop the Empty variant.

warusadura: Will do, thanks!

server/src/gnome/prompter.rs
bilelmoussaoui: You don't have to do anything special here, just use that type.

bilelmoussaoui:
cc @ueno, could you review & ack the crypto code please?

client/src/key.rs
```diff
@@ -47,6 +47,14 @@ impl Key {
         Ok(Self::new(crypto::generate_public_key(private_key)?))
     }
 
+    pub fn generate_public_key_for_secret_exchange(
```
bilelmoussaoui, line 50: Those new methods needs tests, similar to the previous ones to ensure we have the same behaviour in both native & openssl crypto.

client/src/crypto/native.rs
```diff
@@ -86,6 +86,16 @@ pub(crate) fn generate_public_key(private_key: impl AsRef<[u8]>) -> Result<Vec<u
     Ok(public_key_uint.to_bytes_be())
 }
 
+pub(crate) fn generate_public_key_for_secret_exchange(
```
bilelmoussaoui, line 89: As those APIs are only useful for the secret exchange, which is only used by the GNOME implementation I don't think we should have them here and instead move them to the server side under /gnome. Especially that we have a `Key::new`

server/src/gnome/prompter.rs
warusadura: hey. I tried `reply: Optional<Reply>` approach and I think the current approach with Empty variant is bit better or shorter. Because all we have to do is impl Serialize and Deserialize. When using Optional, I had to impl `NoneValue` and bunch of String, &str traits. I just stopped at: trait `FromStr` is not implemented for `Reply`. Should I keep `Empty` variant or continue with `Optional`?

server/src/gnome/prompter.rs
warusadura: I tried everything I can think of. I think this is happening due to a bug in zbus. Probably related to zbus issues #1193 and #320.

server/src/gnome/prompter.rs
bilelmoussaoui: Use zvariant::Optional please. and you shouldn't have to implement that many traits but if you are struggling with it, you can leave it as is and i will take care of it before merging.

warusadura: Ok, thanks :)

server/src/gnome/prompter.rs
bilelmoussaoui: Can you write a minimal reproducer of the issue? I am using the exact same thing in ashpd and it works fine.

warusadura: Will do.

> I am using the exact same thing in ashpd and it works fine.

Yeah, I did see that.

bilelmoussaoui: Actually the issues you mentioned are about the OwnedValue/Value derive macros which you don't use here. So the issue is very likely in the types you are using inside Properties struct. Maybe they are all passed as strings?

warusadura: So [these](https://gitlab.gnome.org/GNOME/gcr/-/blob/main/gcr/gcr-prompt.c#L95) are the GcrPrompt properties.

bilelmoussaoui: Show me the dbus traffic instead.

warusadura: https://upload.disroot.org/r/uCEPb8Pu#Z6OOnI1W4Fx5uALHyb8s5IIVaaDAngKl5tPm5m0tk04=

bilelmoussaoui: Can you confirm this is caused by the properties and not the the reply param of PromptReady?

bilelmoussaoui: Then keep the Properties struct there and add a todo comment about it so we can investigate it further.

warusadura: got it, thanks! I think I'll have to keep this diff as well. I can't test anything without it :)

warusadura: Just locally.

client/src/crypto/native.rs
warusadura: Yeah, it make sense to move the secret exchange related new crypto impls to /gnome. But, we don't move the new functions in keys.rs right? Also, is it ok to have the server side as a dependency for the client side?

bilelmoussaoui: I didn't say anything like that, the client is a library which cannot depend on the server which is a binary...

client/src/crypto/native.rs
warusadura: I was just following the existing implementations.

server/src/gnome/prompter.rs
warusadura: Removed, thanks!

server/src/gnome/prompter.rs
warusadura: Done.

client/src/key.rs
warusadura: Done.

client/src/crypto/native.rs
warusadura: Sorry, I miss understood. Moved secret exchange cryto impl to gnome/crypto/.

server/src/gnome/prompter.rs
```diff
@@ -55,6 +55,24 @@ impl Properties {
             cancel_label: Some("Cancel".to_string()),
         }
     }
+
+    fn for_unlock() -> Self {
+        Self {
+            title: Some("Unlock Keyring".to_string()),
+            choice_label: None,
+            description: Some(
+                "An application wants access to the keyring 'login', but it is locked".to_string(),
```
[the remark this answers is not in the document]
warusadura: Yeah, fixed this thanks!

server/src/gnome/prompter.rs
warusadura: hey. I tried this and I get the following error: `the trait bound `WindowIdentifierType: Serialize` is not satisfied`

server/src/service.rs
```diff
@@ -137,8 +139,17 @@ impl Service {
     pub async fn unlock(
         &self,
         objects: Vec<OwnedObjectPath>,
+        #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(Vec<OwnedObjectPath>, OwnedObjectPath), ServiceError> {
-        let (unlocked, _not_unlocked) = self.set_locked(false, &objects, false).await?;
+        let (unlocked, not_unlocked) = self.set_locked(false, &objects, false).await?;
```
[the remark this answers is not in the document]
warusadura: I have cleaned up a lot issues you mentioned and I'm looking into this now.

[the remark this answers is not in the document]
warusadura: I don't really understand why I need to pass a Future to the Prompt and what it is suppose to do :)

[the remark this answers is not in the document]
bilelmoussaoui: The future is an async function that would be executed once the prompt is done. So you don't have to pass the list of objects to lock/unlock to the prompt mechanism but rather the function that you would execute. by doing so, you can keep all the locking/unlocking logic in the same place and the prompting logic would be generic enough that it could be even tested on CI if needed.

server/src/gnome/prompter.rs
bilelmoussaoui: easy to add upstream, will take care of that :-)

server/src/gnome/keys.rs
```diff
@@ -0,0 +1,70 @@
+use oo7::Key;
```
bilelmoussaoui, line 1: this file is not really useful, move it content to gnome/crypto/mod.rs

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,407 @@
...
+    continue_label: Option<String>,
+    #[zvariant(rename = "cancel-label")]
+    cancel_label: Option<String>,
+}
+
+impl Properties {
+    fn for_lock(object: &str) -> Self {
+        Self {
+            title: None,
+            message: Some(String::from("Lock Keyring")),
```
bilelmoussaoui, line 51: Use to_owned instead of String::from

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,407 @@
...
+            description: Some(format!(
+                "An application wants access to the keyring '{}', but it is locked",
+                object
+            )),
+            warning: Some(String::from(warning)),
+            password_new: None,
+            password_strength: None,
+            choice_label: None,
+            choice_chosen: None,
+            caller_window: None,
```
bilelmoussaoui, line 77: The caller window is something you should pass as a parameter as well

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,407 @@
...
+
+impl NoneValue for Reply {
+    type NoneType = String;
+
+    fn null_value() -> Self::NoneType {
+        String::new()
+    }
+}
+
+impl TryFrom<String> for Reply {
```
bilelmoussaoui, line 101: What uses this ?

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,407 @@
...
+    pub async fn prompt(
+        &self,
+        locked: bool,
+        reply: Optional<Reply>,
+        exchange: &str,
+        connection: zbus::Connection,
+        prompt_path: OwnedObjectPath,
+        objects: Vec<OwnedObjectPath>,
+    ) -> Result<(), ServiceError> {
+        if let Some(reply) = reply.as_ref() {
```
bilelmoussaoui, line 275: why this and the match at the same time?

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,407 @@
...
+    ) -> Result<(), ServiceError> {
+        Prompt::completed(&signal_emitter, dismissed, result).await?;
+        tracing::debug!("Prompt completed.");
+
+        Ok(())
+    }
+
+    pub async fn prompt(
+        &self,
+        locked: bool,
```
bilelmoussaoui, line 268: Why the locked here when you know the role already?

server/src/gnome/prompter.rs
warusadura: Thanks :)

server/src/gnome/prompter.rs
warusadura: Done.

server/src/gnome/prompter.rs
warusadura: This is to fix the following error:

```
error[E0277]: the trait bound `Reply: From<std::string::String>` is not satisfied
   --> server/src/gnome/prompter.rs:140:1
    |
140 | #[interface(name = "org.gnome.keyring.internal.Prompter.Callback")]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<std::string::String>` is not implemented for `Reply`
```

server/src/gnome/prompter.rs
warusadura: Reply (has just two variants now) is wrapped around `zvariant::Optional`. So, here if there is a value we'll match it, otherwise `else` block will handle the NoneValue which is `''`.

server/src/gnome/prompter.rs
warusadura: Yeah, this is set based on the role. `prompt` handles both lock & unlock.

server/src/gnome/prompter.rs
warusadura: Done. Does this offers any performance improvement?

server/src/gnome/prompter.rs
bilelmoussaoui: Optional implements deref to Option so you should use that.

server/src/gnome/prompter.rs
bilelmoussaoui: What i said is that you have access to the role here, the new param is useless

server/src/gnome/prompter.rs
bilelmoussaoui: None, it is the same thing internally

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,407 @@
...
+        Self {
+            title: None,
+            message: Some("Lock Keyring".to_owned()),
+            description: Some(format!("Confirm locking '{}' Keyring", object)),
+            warning: None,
+            password_new: None,
+            password_strength: None,
+            choice_label: None,
+            choice_chosen: None,
+            caller_window: Some("".to_owned()),
```
bilelmoussaoui, line 58: That is not what i meant... I meant the constructor should take a param for this as well.

warusadura:
Hey, currently the server side openssl based tests fails with "the trait `From<ErrorStack>` is not implemented for `oo7::crypto::Error`". I think this is caused by the server side [Cargo.toml](https://github.com/bilelmoussaoui/oo7/blob/6f06a64ee8fd466e918b113ffdad6f9beafd5f53/server/Cargo.toml#L19) and I'm not sure how to fix this.

I ran the tests with: `cargo test --no-default-features --features openssl_crypto`

server/src/gnome/prompter.rs
warusadura: Done.

server/src/gnome/keys.rs
warusadura: Done.

bilelmoussaoui:
> Hey, currently the server side openssl based tests fails with "the trait `From<ErrorStack>` is not implemented for `oo7::crypto::Error`". I think this is caused by the server side [Cargo.toml](https://github.com/bilelmoussaoui/oo7/blob/6f06a64ee8fd466e918b113ffdad6f9beafd5f53/server/Cargo.toml#L19) and I'm not sure how to fix this.
> 
> I ran the tests with: `cargo test --no-default-features --features openssl_crypto`

you have to enable the corresponding oo7 feature as well, in server/Cargo.toml.

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,416 @@
...
+            let secret_exchange = SecretExchange::new().map_err(|err| {
+                ServiceError::ZBus(zbus::Error::FDO(Box::new(zbus::fdo::Error::Failed(
+                    format!("Failed to generate SecretExchange {err}."),
+                ))))
+            })?;
+            let server_exchange = secret_exchange.begin();
+            self.service
+                .set_secret_exchange_pub_key(server_exchange.clone())
+                .await;
+
```
bilelmoussaoui, line 379: why this information is stored in the service and not in the prompter itself? i don't think the gnome specific bits needs to be exposed in Service.

server/Cargo.toml
```diff
@@ -9,13 +9,28 @@ rust-version.workspace = true
...
 zeroize.workspace = true
+
+[features]
+default = ["native_crypto"]
+native_crypto = [
+    "dep:hkdf",
+    "dep:sha2",
+]
+
+openssl_crypto = ["dep:openssl"]
```
bilelmoussaoui, line 36:
```suggestion
default = ["native_crypto"]
native_crypto = [
    "dep:hkdf",
    "dep:sha2",
    "oo7/native_crypto",
]

openssl_crypto = ["dep:openssl", "oo7/openssl_crypto"]
```

server/src/gnome/prompter.rs
warusadura: Done, thanks!

server/src/gnome/prompter.rs
warusadura: Done, thanks!

server/src/gnome/prompter.rs
warusadura: Yeah, I agree. Removed these from the Service.

server/Cargo.toml
warusadura: Thanks! I tried this, but I'm getting `error: You can't enable both openssl_crypto & native_crypto features at once`. Also, tried disabling oo7 default features. It didn't work either.

server/src/prompt.rs
warusadura: But, they are two different paths. Prompt: `/org/freedesktop/secrets/prompt/p{i}` and PrompterCallback: `/org/gnome/keyring/Prompt/p{i}`.

server/src/gnome/prompter.rs
warusadura: Thanks!

server/Cargo.toml
bilelmoussaoui: How are you building it for it to fail this way?

warusadura: `cargo build --no-default-features --features openssl_crypto`

server/src/gnome/prompter.rs
warusadura: https://github.com/bilelmoussaoui/oo7/pull/173#discussion_r1876066717

I tried this (also the above) and now I'm getting the following error,
Screenshot From 2025-02-18 15-19-36

warusadura: image

A6GibKm: Please paste here the windows id that was used.

warusadura:
```
diff --git a/Cargo.lock b/Cargo.lock
index 9969255..e35cb09 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1224,6 +1224,7 @@ dependencies = [
 name = "oo7-daemon"
 version = "0.4.0"
 dependencies = [
+ "ashpd",
  "base64",
  "clap",
  "enumflags2",
diff --git a/server/Cargo.toml b/server/Cargo.toml
index dd20f6b..d677d4f 100644
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -9,6 +9,7 @@ rust-version.workspace = true
 version.workspace = true
 
 [dependencies]
+ashpd = { workspace = true, features = ["backend"] }
 base64 = "0.22"
 clap.workspace = true
 enumflags2 = "0.7"
diff --git a/server/src/gnome/prompter.rs b/server/src/gnome/prompter.rs
index da3fda3..4f66e55 100644
--- a/server/src/gnome/prompter.rs
+++ b/server/src/gnome/prompter.rs
@@ -3,6 +3,7 @@
 
 use std::sync::Arc;
 
+use ashpd::WindowIdentifierType;
 use clap::error::Result;
 use oo7::dbus::ServiceError;
 use serde::{Deserialize, Serialize};
@@ -40,7 +41,7 @@ pub struct Properties {
     #[zvariant(rename = "choice-chosen")]
     choice_chosen: Option<bool>,
     #[zvariant(rename = "caller-window")]
-    caller_window: Option<String>,
+    caller_window: Option<WindowIdentifierType>,
     #[zvariant(rename = "continue-label")]
     continue_label: Option<String>,
     #[zvariant(rename = "cancel-label")]
@@ -48,7 +49,7 @@ pub struct Properties {
 }
 
 impl Properties {
-    fn for_lock(object: &str, window_id: &str) -> Self {
+    fn for_lock(object: &str, window_id: &WindowIdentifierType) -> Self {
         Self {
             title: None,
             message: Some("Lock Keyring".to_owned()),
@@ -58,13 +59,13 @@ impl Properties {
             password_strength: None,
             choice_label: None,
             choice_chosen: None,
-            caller_window: Some(window_id.to_owned()),
+            caller_window: Some(window_id.clone()),
             continue_label: Some("Lock".to_owned()),
             cancel_label: Some("Cancel".to_owned()),
         }
     }
 
-    fn for_unlock(object: &str, warning: &str, window_id: &str) -> Self {
+    fn for_unlock(object: &str, warning: &str, window_id: &WindowIdentifierType) -> Self {
         Self {
             title: Some("Unlock Keyring".to_owned()),
             message: Some("Authentication required".to_owned()),
@@ -77,7 +78,7 @@ impl Properties {
             password_strength: None,
             choice_label: None,
             choice_chosen: None,
-            caller_window: Some(window_id.to_owned()),
+            caller_window: Some(window_id.clone()),
             continue_label: Some("Unlock".to_owned()),
             cancel_label: Some("Cancel".to_owned()),
         }
@@ -146,7 +147,7 @@ pub trait Prompter {
 // org.gnome.keyring.internal.Prompter.Callback
 
 pub struct PrompterCallback {
-    window_id: String,
+    window_id: WindowIdentifierType,
     // SecretExchange public_key
     secret_exchange_pub_key: Arc<RwLock<String>>,
     // SecretExchange aes_key
@@ -215,13 +216,13 @@ impl PrompterCallback {
 }
 
 impl PrompterCallback {
-    pub async fn new(window_id: &str, service: Service) -> Self {
+    pub async fn new(window_id: WindowIdentifierType, service: Service) -> Self {
         let index = service.prompt_index().await;
         Self {
-            window_id: window_id.to_owned(),
             secret_exchange_pub_key: Default::default(),
             secret_exchange_aes_key: Default::default(),
             path: OwnedObjectPath::try_from(format!("/org/gnome/keyring/Prompt/p{index}")).unwrap(),
+            window_id,
             service,
         }
     }
diff --git a/server/src/prompt.rs b/server/src/prompt.rs
index ca48004..e8cdd57 100644
--- a/server/src/prompt.rs
+++ b/server/src/prompt.rs
@@ -32,7 +32,7 @@ pub struct Prompt {
 impl Prompt {
     pub async fn prompt(
         &self,
-        window_id: &str,
+        window_id: ashpd::WindowIdentifierType,
         #[zbus(connection)] connection: &zbus::Connection,
         #[zbus(object_server)] object_server: &zbus::ObjectServer,
     ) -> Result<(), ServiceError> {

```

bilelmoussaoui: the diff you pasted doesn't include the call to PromptCallback::new which probably is passing the wrong WindowIdentifierType. Instead, make those calls pass Option<T> and pass None when you don't have a window identifier like in GNOME instead of passing Some("") which is wrong.

warusadura: Thanks for the reply.
No, the diff there. And there is no `Some("")` in this pr anymore.
So, the `window_id` parameter is coming from the Prompt method `prompt`. We don't pass this. This method get called when we create and serve a Prompt on the object tree. All the time the value of this parameter is `''`. And we can't use `window_id: Option<ashpd::WindowIdentifierType>`. This causes "Signature mismatch: got `s`, expected `ms`" error.

bilelmoussaoui: You can't use Option<T> obviously, like I mentioned previously for the prompt reply, it must be Optional<T>

warusadura: Tried `Optional` too :) "the trait bound `WindowIdentifierType: Default` is not satisfied"

bilelmoussaoui: anyways at this point focus at finishing the basics of the implementation, i will take care of cleaning things afterwards.

warusadura: Understood, thanks! Will keep the `window_id: &str` for now (easy for testing and compiling). I'll add a comment regarding WindowIdentifierType.

server/src/service.rs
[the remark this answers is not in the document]
warusadura: Thanks for the reply, as I have mentioned [during](https://github.com/bilelmoussaoui/oo7/pull/150#discussion_r1852419482) the lock/unlock implementation, the `set_locked()` method does two things.
1: filter out and return object paths that were locked/unlocked without a prompt. We also update the locked state of an item based on the collection's locked state.
2: the actual locking/unlocking happens with a prompt (inside prompter impl). We shouldn't update the locked state right away (this is what we have now).

I agree, it does make sense (on the surface) to pass a future to the Prompt and spawn it when needed. But, during the service method Lock/Unlock `set_locked()` should return a value (objects paths that were locked/unlocked without a prompt) first. Based on that value we decide whether to issue a prompt or not. And since we already pass `service` to the Prompt, I really don't feel we need this :) The extra parameter to `set_locked()`, `has_prompt: bool` makes sure the call to `set_locked()` comes within the prompter impl based on only that we update the locked state. The locking/unlocking logic is exactly the same (generic enough).

the following is what changed (extra parameter plus) in `set_locked()` ,

```
-                    	// TODO: remove this once the prompt implementation is complete.
-                    	collection.set_locked(locked).await?;
+                    	if has_prompt {
+                        	collection.set_locked(locked).await?;
+  
```

Is passing a list of object paths to Prompt constructor a bad design?

server/src/service.rs
warusadura: https://github.com/bilelmoussaoui/oo7/pull/173#discussion_r1966813203

server/src/service.rs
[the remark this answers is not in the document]
bilelmoussaoui: I haven't read any technical blockers at all but just a personal preference you have. The PR is any ways is not ready to be merged given the various unresolved yet ignored review comments.

server/src/service.rs
warusadura: Yeah, this is correct and I have removed this restriction.

server/src/gnome/prompter.rs
warusadura: What do you mean? Maybe the confusion is caused by the fact that I'm using the same name  as `Prompter.perform_prompt`. `PrompterCallback::perform_prompt` is just a wrapper calling `Prompter.perform_prompt` with other code to avoid code duplication.

server/Cargo.toml
warusadura: Fixed this with https://github.com/bilelmoussaoui/oo7/pull/206

warusadura:
Hey Daiki can you please review the crypto impls (`gnome/secret_exchange.rs` and `gnome/crypto/`) when you have some free time. I just fixed an issue found in my openssl based implementation. With that everything works :) In case you wanted to run the openssl based tests you might need this [diff](https://gist.github.com/warusadura/7cc39750421f38d9bf0581758b516dc9).

@ueno

server/src/gnome/prompter.rs
warusadura: https://github.com/bilelmoussaoui/oo7/pull/173#discussion_r1966813203

server/src/service.rs
warusadura: Addressed [here](https://github.com/bilelmoussaoui/oo7/pull/173#discussion_r1966813203)

server/src/prompt.rs
```diff
@@ -1,25 +1,97 @@
...
+        let path = callback.path().clone();
+        let connection = connection.clone();
+
+        object_server.at(&path, callback).await?;
+        tracing::debug!("Prompt `{}` created.", self.path);
+
+        tokio::spawn(async move {
+            let prompter = PrompterProxy::new(&connection).await.unwrap();
+            prompter.begin_prompting(&path).await.unwrap();
+        });
```
bilelmoussaoui, line 51: Needs a comment why this has to be a separate spawned future

server/src/prompt.rs
```diff
@@ -1,25 +1,97 @@
...
+        }
+    }
+
+    pub fn path(&self) -> &OwnedObjectPath {
+        &self.path
+    }
+
+    pub fn role(&self) -> &PromptRole {
+        &self.role
+    }
```
bilelmoussaoui, line 92:
```suggestion
    pub fn role(&self) -> PromptRole {
        self.role
    }
```
it is a copy type

server/src/prompt.rs
```diff
@@ -1,25 +1,97 @@
...
+
+    pub fn objects(&self) -> &Vec<OwnedObjectPath> {
+        &self.objects
+    }
```
bilelmoussaoui, line 96:
```suggestion
    pub fn objects(&self) -> &[OwnedObjectPath] {
        &self.objects
    }
```
I still truly believe the objects don't make sense as part of the prompting but let us deal with that separately.

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,450 @@
...
+        exchange: &str,
+        connection: zbus::Connection,
+        prompt_path: OwnedObjectPath,
+        objects: Vec<OwnedObjectPath>,
+    ) -> Result<(), ServiceError> {
+        let locked = match prompt_role {
+            PromptRole::Lock => true,
+            PromptRole::Unlock => false,
+            PromptRole::CreateCollection => todo!(),
+        };
```
bilelmoussaoui, line 310: I am not sure this is needed. The role is enough.

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,450 @@
...
+        &self.path
+    }
+
+    pub async fn object_label(&self, objects: &Vec<OwnedObjectPath>) -> Option<String> {
+        for object in objects {
+            if let Some(collection) = self.service.collection_from_path(object).await {
+                return Some(collection.label().await);
+            }
+        }
+        None
```
bilelmoussaoui, line 242: The objects could be items, not necessarily collections, no?

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,450 @@
...
+    pub async fn secret_exchange_pub_key(&self) -> String {
+        self.secret_exchange_pub_key.read().await.clone()
+    }
+
+    pub async fn set_secret_exchange_aes_key(&self, aes_key: String) {
+        *self.secret_exchange_aes_key.write().await = aes_key;
+    }
+
+    pub async fn secret_exchange_aes_key(&self) -> String {
+        self.secret_exchange_aes_key.read().await.clone()
```
bilelmoussaoui, line 258: no need for all these methods if they are going to be called once internally. Drop them.

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+
+    pub async fn perform_prompt(
```
bilelmoussaoui, line 261: All of these static methods have 0 usefulness at all. Drop all of them.

server/src/gnome/prompter.rs
warusadura: removed, thanks!

server/src/prompt.rs
warusadura: done.

server/src/gnome/prompter.rs
warusadura: Yes. This is fixed now, thanks!

server/src/gnome/prompter.rs
warusadura: Actually they do :) Consider `PrompterCallback::perform_prompt()` wrapper method. It's invoked in two places. So, this avoids code duplication in two places. And also with this approach I can avoid calling `.unwrap()` inside `tokio::spawn(async move` and avoid the following error, "^ cannot use the `?` operator in an async block that returns `()`"

server/src/prompt.rs
warusadura: Done. Yeah I agree, thanks!

server/src/gnome/prompter.rs
warusadura: Sorry, I don't get it.

server/src/prompt.rs
warusadura: It's not right?

server/src/gnome/prompter.rs
bilelmoussaoui: Well, you should make the spawned t ask return a Result instead of () https://docs.rs/tokio/latest/tokio/task/fn.spawn.html

server/src/gnome/prompter.rs
bilelmoussaoui: I don't get what you wrote here neither. What I said is there is 0 reason to convert an enum that already conveys the information properly to a boolean

bilelmoussaoui:
Can you split the first two & the last commit into a separate PR, I will ping Daiki by email to review only that and we can merge it. I will provide a better approach for the secret exchange API as well so we can land that and figure out the remaining of the prompting here given that this PR is already too huge in terms of review comments.

warusadura:
> Can you split the first two & the last commit into a separate PR, I will ping Daiki by email to review only that and we can merge it. I will provide a better approach for the secret exchange API as well so we can land that and figure out the remaining of the prompting here given that this PR is already too huge in terms of review comments.

Yeah, done. Sorry for the delay, I've been dealing with [this](https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg2024463.html) issue in my destro.
 
I'll remove the split commits and rebase this once #213 is merged.

bilelmoussaoui:
This needs a rebase and please split the lock and unlock changes from the additional common changes for prompting so they can be reviewed easily

server/src/prompt.rs
warusadura: I'm getting the following error,
```
move occurs because `self.role` has type `prompt::PromptRole`, which does not implement the `Copy` trait
```
and the compiler suggest to do a clone. Should I do this or, keep it as it is?

server/src/gnome/prompter.rs
warusadura: Thanks. I think we should keep these wrapper methods, they help a lot with code duplication.

server/src/prompt.rs
bilelmoussaoui: That is what i mentioned above.

server/src/gnome/prompter.rs
bilelmoussaoui: I haven't seen any code duplication so far.

server/src/prompt.rs
warusadura: Done, thanks!

server/src/gnome/prompter.rs
warusadura: :) In `prompter.rs` file search for `stop_prompting`, `perform_prompt`, `prompt_completed` and `prompt`. Each of these static wrapper methods get executed twice.

server/src/service.rs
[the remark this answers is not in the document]
warusadura: We decided to resolve this (passing list of objects around) later: https://github.com/bilelmoussaoui/oo7/pull/173#discussion_r1976572719

server/src/gnome/prompter.rs
warusadura: I think this is resolved too :)

server/src/gnome/prompter.rs
```diff
@@ -0,0 +1,445 @@
...
+                return Some(collection.label().await);
+            } else if let Some(item) = self.service.item_from_path(object).await {
+                return Some(item.label().await);
+            }
+        }
+        None
+    }
+
+    // Wrappers around PrompterProxy methods
+    pub async fn begin_prompting(
```
bilelmoussaoui, line 243: All of those methods are still here.

warusadura: Removed all of them.

bilelmoussaoui:
merged most of it in https://github.com/bilelmoussaoui/oo7/pull/224, if you notice anything left/broken, patches welcome. 

Closing this one

warusadura:
Thanks!

PR #191, opened 2025-02-01
Fix dbus service algorithm fallback

ylxdzsw:
The current error matching is incorrect, as the actual error type returned when algorithm unsupported is `DBus(Service(ZBus(MethodError(OwnedErrorName("org.freedesktop.DBus.Error.NotSupported"), Some(""), Msg { .. }))))`

It is wrapped in the `ServiceError` type. Refer to this line: https://github.com/bilelmoussaoui/oo7/blob/main/client/src/dbus/api/service.rs#L101 

P.S. the two error variant names have inconsistent capitalization: [ServiceError::Z**B**us](https://github.com/bilelmoussaoui/oo7/blob/main/client/src/dbus/error.rs#L11) vs [Error::Z**b**us](https://github.com/bilelmoussaoui/oo7/blob/main/client/src/dbus/error.rs#L23). They actually helped me debugging this.


client/src/dbus/service.rs
```diff
@@ -51,7 +51,7 @@ impl<'a> Service<'a> {
     pub async fn new() -> Result<Service<'a>, Error> {
         let service = match Self::encrypted().await {
             Ok(service) => Ok(service),
-            Err(Error::Zbus(zbus::Error::MethodError(_, _, _))) => Self::plain().await,
```
A6GibKm, line 54: I think it would be more safe to have both match branches.

ylxdzsw: I'm fine with keeping both. Do you want me to change it in this pull request?

bilelmoussaoui: i would keep both in the same commit yes, just to avoid any future regressions

bilelmoussaoui:
> P.S. the two error variant names have inconsistent capitalization: [ServiceError::ZBus](https://github.com/bilelmoussaoui/oo7/blob/main/client/src/dbus/error.rs#L11) vs [Error::Zbus](https://github.com/bilelmoussaoui/oo7/blob/main/client/src/dbus/error.rs#L23). They actually helped me debugging this.

Would you mind making them consistent in a separate commit here? we will do a new breaking API release in the future so this change is fine

ylxdzsw:
I have changed the first commit to keep both branches and added another commit to rename Error:ZBus. (I chose the upper case as it is used more often, e.g. in https://github.com/bilelmoussaoui/oo7/blob/main/client/src/error.rs )

bilelmoussaoui:
the changes looks lgtm, please fix rustfmt complains and will merge afterwards

client/src/dbus/service.rs
```diff
@@ -51,7 +51,8 @@ impl<'a> Service<'a> {
...
+            Err(Error::ZBus(zbus::Error::MethodError(_, _, _))) => Self::plain().await,
+            Err(Error::Service(ServiceError::ZBus(zbus::Error::MethodError(_, _, _)))) => Self::plain().await,
```
bilelmoussaoui, line 55:
```suggestion
            Err(Error::ZBus(zbus::Error::MethodError(_, _, _))) | Err(Error::Service(ServiceError::ZBus(zbus::Error::MethodError(_, _, _)))) => Self::plain().await,
```
maybe? or would that be too unreadable?

A6GibKm: In this very specific case I would say that it is less readable and the diff is also less readable, considering the rhs is just a single call.

ylxdzsw: I also think it less readable to merge the two already long lines together. I tried to replace `_, _, _` to `..` but the second branch still exceeds the width limit and results in the two branches being formatted differently (the first branch is inline while the second is put into a block). I have updated the pull request with my best effort solution.

PR #192, opened 2025-02-01
Update rand and getrandom

A6GibKm:
Closes: https://github.com/bilelmoussaoui/oo7/pull/189, https://github.com/bilelmoussaoui/oo7/pull/187.


client/src/file/api/mod.rs
```diff
@@ -329,6 +329,10 @@ impl TryFrom<&[u8]> for Keyring {
     }
 }
 
+fn salt() -> Vec<u8> {
```
bilelmoussaoui, line 332: I don't think it is necessary to add this function

A6GibKm: Removed that commit.

PR #206, opened 2025-02-26
Disable oo7 default-features in workspace.dependencies

warusadura:
This is to stop oo7/native_crypto feature being enabled when using oo7/openssl_crypto on the server side (or the opposite).

This also resolves the following error when setting oo7 `default-features = false` in workspace members.
"`default-features` is ignored for oo7, since `default-features` was not specified for `workspace.dependencies.oo7`, this could become a hard error in the future".

### DIFF
--- a/portal/Cargo.toml
+++ b/portal/Cargo.toml
@@ -14,7 +14,7 @@ version.workspace = true
 [dependencies]
 ashpd = {workspace = true, features = ["backend", "tracing"]}
 clap.workspace = true
-oo7.workspace = true
+oo7 = { workspace = true, default-features = true }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 tracing.workspace = true
 tracing-subscriber.workspace = true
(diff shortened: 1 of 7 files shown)


portal/Cargo.toml
```diff
@@ -14,7 +14,7 @@ version.workspace = true
 [dependencies]
 ashpd = {workspace = true, features = ["backend", "tracing"]}
 clap.workspace = true
-oo7.workspace = true
+oo7 = { workspace = true, default-features = true }
```
bilelmoussaoui, line 17: shouldn't this enable the tokio feature? have you tested that the portal & cli works with your changes?

warusadura: tokio feature is [covered](https://github.com/bilelmoussaoui/oo7/blob/cbc3fa9b8df33634595c9d846dc70fc9b00338ed/client/Cargo.toml#L59) with the `default-features = true` right? I have just tested the cli and it works. I'm not sure how to test the portal.

PR #213, opened 2025-03-11
server: Add SecretExchange implementation

warusadura:
Please see the commits.

Commits are split from #173


.github/workflows/CI.yml
```diff
@@ -53,6 +56,13 @@ jobs:
         run: |
           cargo test --manifest-path ./client/Cargo.toml --no-default-features --features tokio --features openssl_crypto
 
+      - name: Test Server crypto/gnome (native)
```
bilelmoussaoui, line 59: Test Server (native)

same below. and squash this commit with the first one.

warusadura: Done, thanks!

server/src/gnome/secret_exchange.rs
```diff
@@ -0,0 +1,174 @@
...
+
+const SECRET: &str = "secret";
+const PUBLIC: &str = "public";
+const PRIVATE: &str = "private";
+const IV: &str = "iv";
+const PROTOCOL: &str = "[sx-aes-1]\n";
+const CIPHER_TEXT_LEN: usize = 16;
+
+#[derive(Debug, Zeroize)]
+pub struct SecretExchange {
```
bilelmoussaoui, line 25: After giving this some thoughts and looking at the current usage in the prompting code. I would drop this struct and also drop the begin method. Instead, inside PrompterCallback you would store the `Key`s directly.

Then have the following methods:
- `secret_exchange::begin(public_key: &Key)`
- `secret_exchange::handshake(client_private_key: &Key, server_public_key: &Key)`
- `secret_exchange::retrieve(exchange: &str, aes_key: &Key)` 

As a first step, a second step would consists of creating a wrapper type for the Exchange that holds a String internally and have an easy way to convert From/To a HashMap to simplify the serialization/deserialization.

warusadura:
> Please resolve the conflicts with Cargo.lock.

done.

warusadura:
I'm not sure what to do about the `never used` clippy warning.

bilelmoussaoui:
Just silence them for now by adding a #[allow(unused)] just before the mod gnome;

warusadura:
Thanks, done.

bilelmoussaoui:
Let us land this then, we can improve things and get the whole crypto code reviewed before a potential release.

Issue #207, opened 2025-03-05
After upgrading to 0.4.0, some users get a `file::Error::IncorrectSecret` error


zecakeh:
For some users the upgrade is fine, but for others, they get this error which seems to mean that oo7 fails to decrypt the items in the file? See [fractal#1603](https://gitlab.gnome.org/World/fractal/-/issues/1603).

Downgrading back to 0.3.x works so its looks like it's a regression.

bilelmoussaoui:
That indeed would give us more insights

bilelmoussaoui:
@zecakeh I published a 0.4.1 with the added log. Please re-switch back to 0.4 in Fractal and see what causes the issue. Thanks!

bilelmoussaoui:
@zecakeh does this happen when using the secret portal or when interacting with the host keyring ?

zecakeh:
With the secret portal

bilelmoussaoui:
Interesting. I guess the next debugging step we can do here is to see how many items have this issue (if there are more than one).

zecakeh:
So we added more logs to look into it. The users we tested with expected to have 2 sessions. There are 3 encrypted items, only one of them failing to decrypt. The 2 that succeed to decrypt are the expected sessions.

Looking at the map of hashed attributes in the encrypted item, it looks like a very old item, that could even possibly have been created by libsecret before we were using oo7. So we have no idea why it's still here.

zecakeh:
Maybe not error, but return a list/count of items that failed to decrypt?

zecakeh:
Except if all items failed to decrypt maybe

bilelmoussaoui:
Return it where? because you are using the generic API that can either be baked by the secret portal or the dbus backend...

bilelmoussaoui:
An alternative approach would be to raise that error only if n_broken_items > n_not_broken_items, add a new method that would skip the secret check so you can get access to the inner file backend, add a new method to the file backend that would remove all the items that fail to decrypt.

Would that cover all your needs?

theCapypara:
Hello! I also got reports of users of my app running into this issue - however I am already on 0.4.3. I am unfortunately not at all familiar with the ins and outs of oo7 or the secrets backend or what this error actually means. I'd need some guidance to know if this is an issue of oo7, my app or the setup that my users have.

https://github.com/theCapypara/field-monitor/issues/18

bilelmoussaoui:
Left a comment with some clarification about the situation there.

[...]

PR #219, opened 2025-03-19
client: Relax IncorrectSecret condition

bilelmoussaoui:
Fixes https://github.com/bilelmoussaoui/oo7/issues/207

cc @zecakeh

### DIFF
--- a/client/src/file/mod.rs
+++ b/client/src/file/mod.rs
@@ -807,9 +807,18 @@ mod tests {
 
     #[tokio::test]
     async fn change_secret() -> Result<(), Error> {
-        let path = PathBuf::from("../../tests/test_rekeying.keyring");
+        let data_dir = tempdir()?;
+        let v0_dir = data_dir.path().join("keyrings");
+        let v1_dir = v0_dir.join("v1");
+        fs::create_dir_all(&v1_dir).await?;
 
-        let keyring = Keyring::load(&path, strong_key()).await?;
+        let fixture_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("fixtures")
+            .join("default.keyring");
+        let keyring_path = v1_dir.join("default.keyring");
+        fs::copy(&fixture_path, &keyring_path).await?;
+
+        let keyring = Keyring::load(&keyring_path, Secret::blob("test")).await?;
         let attributes = HashMap::from([("attr", "value")]);
         let item_before = keyring
             .create_item("test", &attributes, "password", false)
@@ -819,7 +828,7 @@ mod tests {
         keyring.change_secret(secret).await?;
 
         let secret = Secret::blob("new_secret");
-        let keyring = Keyring::load(&path, secret).await?;
+        let keyring = Keyring::load(&keyring_path, secret).await?;
         let item_now = keyring.lookup_item(&attributes).await?.unwrap();
 
         assert_eq!(item_before.label(), item_now.label());
@@ -829,7 +838,7 @@ mod tests {
         // No items were broken during the secret change
         assert_eq!(keyring.delete_broken_items().await?, 0);
 
-        fs::remove_file(path).await?;
+        fs::remove_file(keyring_path).await?;
 
         Ok(())
     }
--- a/client/src/file/mod.rs
+++ b/client/src/file/mod.rs
@@ -95,9 +95,39 @@ impl Keyring {
     /// * `path` - The path to the file backend.
     /// * `secret` - The service key, usually retrieved from the Secrets portal.
     pub async fn load(path: impl AsRef<Path>, secret: Secret) -> Result<Self, Error> {
+        Self::load_inner(path, secret, true).await
+    }
+
+    /// Load from a keyring file.
+    ///
+    /// # Arguments
+    ///
+    /// * `path` - The path to the file backend.
+    /// * `secret` - The service key, usually retrieved from the Secrets portal.
+    ///
+    /// # Safety
+    ///
+    /// The secret is not validated to be the correct one to decrypt the keyring
+    /// items. Allowing the API user to write new items with a different
+    /// secret on top of previously added items with a different secret.
+    ///
+    /// As it is not a supported behaviour, this API is mostly meant for
+    /// recovering broken keyrings.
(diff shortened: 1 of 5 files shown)


client/src/file/mod.rs
```diff
@@ -118,15 +148,34 @@ impl Keyring {
...
 
-        let key = keyring.derive_key(&secret)?;
-        for encrypted_item in &mut keyring.items {
-            if let Err(_err) = encrypted_item.clone().decrypt(&key) {
+        if validate_items {
+            let key = keyring.derive_key(&secret)?;
+
+            let mut n_broken_items = 0;
+            let mut n_valid_items = 0;
+            for encrypted_item in &mut keyring.items {
```
zecakeh, line 156: It doesn't look like you actually need the `mut`.

client/src/file/mod.rs
```diff
@@ -475,6 +524,29 @@ impl Keyring {
...
+            if let Err(_err) = encrypted_item.clone().decrypt(&key) {
+                broken_items.push(index);
+            }
+        }
+        let n_broken_items = broken_items.len();
+        for index in broken_items {
+            keyring.items.remove(index);
+        }
+        drop(keyring);
+        Ok(n_broken_items)
```
zecakeh, line 548: Should this call `self.write().await` before returning to persist the changes to the file? It seems like all other methods that make modifications in the Keyring do.

zecakeh (commented):
The API looks good, I'm going to try to use it in Fractal.

client/src/file/mod.rs
```diff
@@ -118,15 +148,34 @@ impl Keyring {
...
+                {
+                    tracing::error!(
+                        "The file contains {n_broken_items} broken items and {n_valid_items} valid ones."
+                    );
+                    tracing::info!("Please switch to Keyring::load_unchecked to load the keyring with the secret validation and restore your valid items in a new keyring.");
+                }
                 return Err(Error::IncorrectSecret);
             }
+
+            drop(key);
```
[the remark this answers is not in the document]
bilelmoussaoui: it was there before and doesn't hurt to have it explicitly.

client/src/file/mod.rs
```diff
@@ -95,6 +95,36 @@ impl Keyring {
...
+    ///
+    /// # Safety
+    ///
+    /// The secret is not validated to be the correct one to decrypt the keyring
+    /// items. Allowing the API user to write new items with a different
+    /// secret on top of previously added items with a different secret.
+    ///
+    /// As it is not a supported behaviour, this API is mostly meant for
+    /// recovering broken keyrings.
+    pub async unsafe fn load_unchecked(
```
zecakeh, line 116: Could we also have a `load_default_unchecked` maybe, for simplicity?

client/src/file/mod.rs
```diff
@@ -118,15 +148,34 @@ impl Keyring {
...
+                if let Err(_err) = encrypted_item.clone().decrypt(&key) {
+                    #[cfg(feature = "tracing")]
+                    tracing::error!("Failed to decrypt item, incorrect secret?: {_err}");
+                    n_broken_items += 1;
+                } else {
+                    n_valid_items += 1;
+                }
+            }
+
+            if n_broken_items > n_valid_items {
```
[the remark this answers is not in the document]
bilelmoussaoui: Indeed, done that.

client/src/file/mod.rs
zecakeh: Well that's the case that we have. I'm not sure why. Maybe the user migrated their keyring to a new system so they got a different secret, but it was not catched at the time?

bilelmoussaoui: No idea how did any of that happen.

bilelmoussaoui: The problem is that i also have a user here https://gitlab.gnome.org/World/Authenticator/-/issues/452 reporting a similar one. Although, I don't know how many broken items they had. 

The thing about making it always fail if there is any  invalid item makes the load API not work around those items which seemed like good enough for me, imho.

zecakeh: My plan was to add a one-time migration in Fractal that removes broken items automatically, so we don't have to worry about it after upgrading oo7. But if it reoccurs after that, it's clearly a problem.

That said if you want to keep the behavior you had, it's also fine by me. There's a probability that users will open an issue because they see the same error every time that they open the app though. Or maybe just put it as a warning, because it doesn't actually block the application.

bilelmoussaoui: I have switched the behaviour back to the one I had. We can see what the users end up reporting with their logs and figure out a plan from there instead of over-doing it now.

If this looks lgtm, I will merge and release it as 0.4.2

client/src/file/mod.rs
```diff
@@ -118,15 +160,35 @@ impl Keyring {
...
-            if let Err(_err) = encrypted_item.clone().decrypt(&key) {
+        if validate_items {
+            let key = keyring.derive_key(&secret)?;
+
+            let mut n_broken_items = 0;
+            let mut n_valid_items = 0;
+            for encrypted_item in &keyring.items {
+                if let Err(_err) = encrypted_item.clone().decrypt(&key) {
+                    #[cfg(feature = "tracing")]
+                    tracing::error!("Failed to decrypt item, incorrect secret?: {_err}");
```
zecakeh, line 171: I was talking about putting this one as `warn`.

bilelmoussaoui: Indeed. Fixed

client/src/keyring.rs
```diff
@@ -25,8 +25,21 @@ pub enum Keyring {
 }
 
 impl Keyring {
+    /// Create a new instance of the Keyring that automatically removes the
+    /// broken items from the file backend keyring.
+    ///
+    /// This method will probably be removed in future versions if the
+    /// mis-behaviour is tracked and fixed.
+    pub async fn with_auto_delete_broken_items() -> Result<Self> {
```
bilelmoussaoui, line 33: @zecakeh this should make things easier for us, but the name is very bad...

zecakeh: Ah nice !

PR #222, opened 2025-03-23
wip: server: Fix sessions not closing

warusadura:
wip


server/src/service.rs
```diff
@@ -410,4 +416,50 @@ impl Service {
...
     }
+
+    pub async fn watch_clients(&self) -> Result<(), Error> {
+        // wip: The idea is store client information when a client calls a method.
+        // Watch for NameOwnerChanged signal from the same client.
+        // Clean up session and resources based on that.
+        let connect_rule = zbus::MatchRule::builder()
+            .msg_type(zbus::message::Type::MethodCall)
+            .path_namespace("/org/freedesktop/secrets")?
+            .build();
```
bilelmoussaoui, line 427: This seems like too broad, a client has to start a session somewhere first no? so that is where i would start the tracking :-)

warusadura: Thanks for the prompt review. Yeah, I agree :)

server/src/service.rs
```diff
@@ -76,6 +80,14 @@ impl Service {
             .await
             .insert(path.clone(), session.clone());
 
+        if let Some(sender) = header.sender() {
+            tracing::info!("Client {} connected.", sender);
+            self.clients
+                .lock()
+                .await
+                .insert(sender.as_str().to_owned(), path.clone());
```
bilelmoussaoui, line 88: Why do you convert the sender type to a string?

bilelmoussaoui: Also why doesn't this keep a ref to the session instead?

server/src/service.rs
```diff
@@ -458,4 +474,42 @@ impl Service {
 
         Ok(signal_emitter)
     }
+
+    pub async fn remove_client(&self, old_owner: &str) -> Option<OwnedObjectPath> {
+        self.clients.lock().await.remove(old_owner)
```
bilelmoussaoui, line 479: Make this return the session directly instead of the session path

server/src/service.rs
```diff
@@ -458,4 +474,42 @@ impl Service {
...
+    }
+
+    pub async fn cleanup_sessions_on_client_disconnect(&self) -> Result<(), Error> {
+        let rule = zbus::MatchRule::builder()
+            .msg_type(zbus::message::Type::Signal)
+            .sender("org.freedesktop.DBus")?
+            .member("NameOwnerChanged")?
+            .build();
```
bilelmoussaoui, line 487: This is too broad, you will get too many NameOwnerChanged, not only for secrets service only.

server/src/service.rs
```diff
@@ -458,4 +474,42 @@ impl Service {
...
+        let mut stream = zbus::MessageStream::for_match_rule(rule, &self.connection, None).await?;
+
+        while let Some(message) = stream.try_next().await? {
+            let Ok((_name, old_owner, new_owner)) =
+                message.body().deserialize::<(String, String, String)>()
+            else {
+                continue;
+            };
+
```
bilelmoussaoui, line 496: Is it possible that the owner changes? if not, maybe assert new_owner is NULL (once yo uupdate the matching rule)

server/src/service.rs
```diff
@@ -458,4 +474,42 @@ impl Service {
...
+            if new_owner.is_empty() {
+                if let Some(session_path) = self.remove_client(&old_owner).await {
+                    if let Some(session) = self.session(&session_path).await {
+                        match session.close(self.object_server()).await {
```
bilelmoussaoui, line 500: this is weird, instead remove the object_server param and make that close function use the object server from self.service.object_server or so.

server/src/service.rs
```diff
@@ -458,4 +474,42 @@ impl Service {
...
+                            Ok(_) => tracing::info!(
+                                "Client {} disconnected, session: {} closed.",
+                                old_owner,
+                                session_path
+                            ),
+                            Err(_) => tracing::info!("Failed to close session: {}.", session_path),
```
bilelmoussaoui, line 506: This should not be just info!

server/src/service.rs
```diff
@@ -458,4 +474,42 @@ impl Service {
...
+
+        while let Some(message) = stream.try_next().await? {
+            let Ok((_name, old_owner, new_owner)) =
+                message.body().deserialize::<(String, String, String)>()
+            else {
+                continue;
+            };
+
+            if new_owner.is_empty() {
+                if let Some(session_path) = self.remove_client(&old_owner).await {
```
bilelmoussaoui, line 498: Log that the client is already disconnected if self.remove_client returns None

server/src/service.rs
```diff
@@ -458,4 +474,42 @@ impl Service {
 
         Ok(signal_emitter)
     }
+
+    pub async fn remove_client(&self, old_owner: &str) -> Option<OwnedObjectPath> {
+        self.clients.lock().await.remove(old_owner)
+    }
+
+    pub async fn cleanup_sessions_on_client_disconnect(&self) -> Result<(), Error> {
```
bilelmoussaoui, line 482: Just name this on_client_disconnect.

server/Cargo.toml
```diff
@@ -22,6 +22,7 @@ rpassword = "7.3"
 serde.workspace = true
 sha2 = { version = "0.10", optional = true }
 tokio = { workspace = true, features = ["full"] }
+tokio-stream = "0.1"
```
bilelmoussaoui, line 25: Do you need this dependency? for what?

warusadura:
Sorry for the late response, will address review comments this week :)

server/Cargo.toml
warusadura: This is for `tokio_stream::StreamExt;`
I can't call `stream.try_next()` without this trait.

server/src/service.rs
warusadura: sure.

server/src/service.rs
bilelmoussaoui: by using add_arg no ?

server/src/service.rs
bilelmoussaoui: I feel like storing both the session twice is too much, maybe we should combine the path / sender into a key for finding the corresponding session

server/src/service.rs
warusadura: Removed remove_client()

server/src/service.rs

server/src/service.rs
warusadura: May be error! ?

server/src/service.rs
warusadura: Yeah, I agree. Session now stores the client_name and introduced session_from_client_name

warusadura: > Why do you convert the sender type to a string?

UniqueName requires named lifetime parameter :)

server/src/service.rs
warusadura: now, new_owner is NULL is handled by `.arg(2, "")?`

server/src/service.rs
```diff
@@ -458,4 +467,50 @@ impl Service {
...
+            .build();
+        let mut stream = zbus::MessageStream::for_match_rule(rule, &self.connection, None).await?;
+
+        while let Some(message) = stream.try_next().await? {
+            let Ok((_name, old_owner, new_owner)) =
+                message.body().deserialize::<(String, String, String)>()
+            else {
+                continue;
+            };
+
```
bilelmoussaoui, line 499: can this ever fail to deserialize? also add the assertion that new_owner is empty, because you have it in the match rule...

server/src/service.rs
```diff
@@ -458,4 +467,50 @@ impl Service {
...
+            if new_owner.is_empty() {
+                if let Some(session) = self.session_from_client_name(&old_owner).await {
+                    match session.close().await {
+                        Ok(_) => tracing::info!(
+                            "Client {} disconnected, session: {} closed.",
+                            old_owner,
+                            session.path()
+                        ),
+                        Err(_) => tracing::error!("Failed to close session."),
```
bilelmoussaoui, line 508: Log the error please.

server/src/service.rs
bilelmoussaoui: There is an OwnedUniqueName. Please don't mark threads as resolved when they are not.

PR #234, opened 2025-04-28
server: match items if search attributes are a subset

### LINKED ISSUE #233: `oo7` fails to match secrets when searched attributes are a subset of stored attributes
I was testing `oo7` as a replacement for `gnome-keyring` on my system. I noticed that signing in to my GitHub account with `zed` (which uses the Secret Service API) would write a secret to the collection, but it couldn’t find that secret again after being closed and reopened.

When `zed` wrote the secret, the attributes looked like this:

```json
{
  "url": "https://zed.dev",
  "username": "..."
}
```

However, it searched for the secret using only a subset of the attributes:

```json
{
  "url": "https://zed.dev"
}
```

I'm not familiar enough with the FreeDesktop specification, but it seems reasonable to expect that items with a superset of the searched attributes should still match.

noxpardalis:
Change attribute comparison to match on lookups where the search attributes form a subset of the item's attributes.

Fixes #233

### DIFF
--- a/server/src/collection.rs
+++ b/server/src/collection.rs
@@ -229,7 +229,16 @@ impl Collection {
         let mut items = Vec::new();
 
         for item in self.items.lock().await.iter() {
-            if item.attributes().await == *attributes {
+            let item_attributes = item.attributes().await;
+
+            // Check if the (key, value) pairs in the requested attributes are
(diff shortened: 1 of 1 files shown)


A6GibKm (approved):
lgtm! Thanks!

PR #251, opened 2025-06-02
portal: Dont ask for app tokens with schema

A6GibKm:
If gnome-keyring creates a secret without a schema, due to https://github.com/bilelmoussaoui/oo7/issues/233, the search would return no values and a new secret would be stored by the app. This would result in the app getting the wrong secret when using the portal.

Secrets stored by gnome-keyring do not have a schema set. Moreover

It could be argued that https://github.com/bilelmoussaoui/oo7/issues/233 is the real issue at place, but the generic schema is the default schema and assumed if no schema is set and it does not make sense to search for secrets having it.


bilelmoussaoui:
i am fine merging this

PR #253, opened 2025-06-02
Systemd creds

A6GibKm:
This allows users to manually declare a credential so that the service can automatically unlock the session keyring.

**cc** @AdrianVovk


AdrianVovk:
At  quick glance over, this looks reasonable to me

server/src/main.rs
```diff
@@ -34,18 +38,42 @@ async fn main() -> Result<(), Error> {
...
+    let secret = if args.login {
         let password = rpassword::prompt_password("Enter the login password: ")?;
         if password.is_empty() {
             tracing::error!("Login password can't be empty.");
             return Err(Error::EmptyPassword);
         }
-        secret = Some(oo7::Secret::text(password));
-    }
+        Some(oo7::Secret::text(password))
+    } else if let Ok(credential_dir) = std::env::var("CREDENTIALS_DIRECTORY") {
```
bilelmoussaoui, line 70: Can you add a very brief documentation about this to the server/README?

server/src/main.rs
```diff
@@ -34,18 +38,42 @@ async fn main() -> Result<(), Error> {
...
+        // FIXME We should not exit the service if the provided password is not
+        // the correct one in this branch. The credential might run out of sync
+        // and this is only a fallback.
+        let mut contents = Vec::new();
+        let cred_path = Path::new(&credential_dir).join("oo7-keyring-encryption-password");
+
+        match tokio::fs::File::open(&cred_path).await {
+            Ok(mut cred_file) => {
+                tracing::info!("Unlocking session keyring with user's systemd credentials");
+                cred_file.read_to_end(&mut contents).await?;
```
bilelmoussaoui, line 78: use tokio::fs::read instead?

server/src/main.rs
```diff
@@ -34,18 +38,42 @@ async fn main() -> Result<(), Error> {
...
         }
-        secret = Some(oo7::Secret::text(password));
-    }
+        Some(oo7::Secret::text(password))
+    } else if let Ok(credential_dir) = std::env::var("CREDENTIALS_DIRECTORY") {
+        // We try to unlock the login keyring with a system credential.
+        //
+        // FIXME We should not exit the service if the provided password is not
+        // the correct one in this branch. The credential might run out of sync
+        // and this is only a fallback.
```
bilelmoussaoui: I guess this means we need a way to tell the server that this password might not be correct...

server/src/main.rs
A6GibKm: This is `tokio::io::AsyncReadExt::read_to_end`. `read` in general has the issue that it reads an undefined amount of bytes.

server/src/main.rs
A6GibKm: Or just wrapping the secret in an `enum { Infallible(Secret), Fallible(secret) }` could be done to distinguish them and unlocking them with or without `?` depending on which enum was used.

bilelmoussaoui: Doing something like that would be nice indeed.

A6GibKm: updated.

A6GibKm:
Perhaps the secret should be named `oo7.keyring-encryption-password` instead (the dot denotes a namespace).

bilelmoussaoui:
> Perhaps the secret should be named `oo7.keyring-encryption-password` instead (the dot denotes a namespace).

sounds good to me

server/src/main.rs
A6GibKm: Done.

server/README.md
```diff
@@ -4,6 +4,30 @@
...
+can be stored as an encrypted blob in the user's credstore via
+
+``` sh
+mkdir -p ${XDG_CONFIG_HOME:-~/.config}/credstore.encrypted
+systemd-ask-password -n | systemd-creds encrypt --user --name=oo7.keyring-encryption-password - ${XDG_CONFIG_HOME:-~/.config}/credstore.encrypted/oo7.keyring-encryption-password
+```
+
+**WARNING**: Any user which able to read this file and access the TPM (e.g the
+`root` user) can decrypt this blob. This can be mitigated if using an encrypted
+home with [systemd-homed](https://systemd.io/HOME_DIRECTORY/).
```
bilelmoussaoui:
```suggestion
**WARNING**: Any user capable of reading this file and access the TPM (e.g the
`root` user) can decrypt this blob. This can be mitigated if using an encrypted
home with [systemd-homed](https://systemd.io/HOME_DIRECTORY/).
```

bilelmoussaoui:
otherwise lgtm, once the the CI passes

PR #259, opened 2025-06-03
Make services dbus activable

A6GibKm:
They were before, but simply by running the binaries. This makes it so that they start the service instead. The /usr/bin/false comes from:

```
$ cat /usr/share/dbus-1/system-services/org.freedesktop.hostname1.service
[D-BUS Service]
Name=org.freedesktop.hostname1
Exec=/bin/false
User=root
SystemdService=dbus-org.freedesktop.hostname1.service
```

this needs testing.


portal/data/meson.build
```diff
@@ -1,29 +1,37 @@
+fs = import('fs')
+
 configure_file(
-  input: 'oo7-portal.desktop.in',
+  input: '@0@.desktop.in'.format(service_name),
```
bilelmoussaoui, line 4: can we rename the files themselves instead?

bilelmoussaoui:
Not everything under  /usr/share/dbus-1/system-services uses /bin/false, why?

portal/data/meson.build
A6GibKm: wdym? This change functionally does not do anything since service_name is oo7-portal here. It is just so life is easier if we rename them.

bilelmoussaoui:
The first two commits are doing too many things at once. Can you split the renaming bits into a separate commit? maybe

A6GibKm:
Done.

PR #263, opened 2025-06-05
server: Use connection builder

A6GibKm:
cc @zeenix does this look good to you? It uses https://github.com/A6GibKm/zbus/tree/add-builder-for-flags.


zeenix:
> cc @zeenix does this look good to you?

No, since it completely ignores the important issue I mentioned on Matrix. :) I'll discuss that in your zbus PR.

PR #265, opened 2025-06-08
Add more logging and always ask for DoNotQueue

A6GibKm:


server/src/service.rs
```diff
@@ -292,6 +292,13 @@ impl Service {
         flags: BitFlags<zbus::fdo::RequestNameFlags>,
     ) -> Result<(), Error> {
         let connection = zbus::Connection::session().await?;
+
+        if let Some(unique_name) = connection.unique_name() {
+            tracing::info!("Got unique name {unique_name} on the bus");
+        }
+
+        setup_dbus_proxy(&connection).await?;
```
bilelmoussaoui: Maybe spawn the task here to avoid two functions instead of one? Also the name is confusing

A6GibKm: Suggestions welcome regarding the function names. Indeed, they are not good. 

Regarding the spawning, I think it is better to spawn after we get the name_acquired_stream, otherwise the might be a race between the call requesting the name and we listening for the call.

A6GibKm: Renamed both fns.

server/src/service.rs
```diff
@@ -295,6 +295,13 @@ impl Service {
         flags: BitFlags<zbus::fdo::RequestNameFlags>,
     ) -> Result<(), Error> {
         let connection = zbus::Connection::session().await?;
+
+        if let Some(unique_name) = connection.unique_name() {
+            tracing::info!("Got unique name {unique_name} on the bus");
```
bilelmoussaoui:
```suggestion
            tracing::info!("Unique name {unique_name} acquired");
```

server/src/service.rs
```diff
@@ -520,3 +527,28 @@ impl Service {
         Ok(signal_emitter)
     }
 }
+
+async fn connect_name_acquired(connection: &zbus::Connection) -> Result<(), zbus::Error> {
+    let proxy = zbus::fdo::DBusProxy::new(connection).await?;
+    let name_acquired_stream = proxy.receive_name_acquired().await?;
+
+    tokio::spawn(async move {
```
bilelmoussaoui, line 535: I still believe the spawn should be called for connect_name_acquired.

server/src/service.rs
```diff
@@ -520,3 +527,28 @@ impl Service {
...
+        if let Err(err) = name_acquired_callback(name_acquired_stream).await {
+            tracing::error!("Could not read name acquired signal: {err}");
```
bilelmoussaoui: can this ever fail?

server/src/service.rs
```diff
@@ -520,3 +527,28 @@ impl Service {
...
+    Ok(())
+}
+
+async fn name_acquired_callback(
+    mut name_acquired_stream: zbus::fdo::NameAcquiredStream,
+) -> Result<(), zbus::Error> {
+    if let Some(name_acquired) = name_acquired_stream.next().await {
+        let name_acquired_args = name_acquired.args()?;
+        let name = name_acquired_args.name();
+        tracing::info!("Acquired requested name {name} on the bus");
```
bilelmoussaoui: Use the same logging message as above

server/src/service.rs
A6GibKm: No idea to be honest. The only thing that could fail is https://docs.rs/zbus/2.0.1/zbus/fdo/struct.NameAcquired.html#method.args returning an error. cc @zeenix

server/src/service.rs
A6GibKm: If `connection.request_name_with_flags` finishes before the NameAquiredStream from `proxy.receive_name_acquired()` finishes, then the callback will never return. See https://github.com/dbus2/zbus/pull/1401#discussion_r2136417956.

Wrapping `receive_name_acquired` in a spawn might result in this race.

A6GibKm: I settled for inlining the function inside the `spawn`. Perhaps could be simplified into a simple `if let Some(Ok(name_acquired_args))` depending on the conditions in which `args()` can fail.

bilelmoussaoui (approved):

PR #268, opened 2025-06-14
Create locked/unlocked keyring variants

A6GibKm:


server/src/item.rs
```diff
@@ -14,11 +14,17 @@ use zbus::zvariant::OwnedObjectPath;
 
 use crate::{Service, collection::Collection, error::custom_service_error};
 
+#[derive(Debug, Clone)]
+enum InnerItem {
+    Locked(file::LockedItem),
+    Unlocked(file::Item),
+}
```
bilelmoussaoui: Should we have this enum in client/file and rename `file::Item` to `file::UnlockedItem` or so?

server/src/item.rs
```diff
@@ -115,62 +123,95 @@ impl Item {
...
     pub async fn attributes(&self) -> HashMap<String, String> {
-        self.inner
-            .lock()
-            .await
-            .attributes()
-            .iter()
-            .map(|(k, v)| (k.to_owned(), v.to_string()))
-            .collect()
+        match &*self.inner.lock().await {
+            InnerItem::Locked(_inner) => todo!(),
```
bilelmoussaoui: I assume this is because we can't return custom errors from properties, can we return zbus::fdo errors? if so we can map the error for now. Otherwise maybe return Default::default() would be enough with a todo message pointing to the corresponding zbus issue.

client/src/file/locked_keyring.rs
```diff
@@ -0,0 +1,136 @@
...
+        if v1_path.exists() {
+            return Self::load(v1_path).await;
+        }
+
+        let v0_path = api::Keyring::path(name, api::LEGACY_MAJOR_VERSION)?;
+        if v0_path.exists() {
+            // Migrating a locked keyring is more involved without having the
+            // secret.
+            todo!();
+        };
```
bilelmoussaoui, line 137: There is a fundamental difference between v0 & v1. v0 encrypts the whole keyring, v1 encrypts individual items. So we cannot really achieve this.

I am leaning towards not providing the open api for locked keyring and only do that for cases where you also provide a secret.

client/src/file/locked_keyring.rs
```diff
@@ -0,0 +1,136 @@
...
+#[derive(Debug)]
+pub struct LockedKeyring {
+    pub(super) keyring: Arc<RwLock<api::Keyring>>,
+    pub(super) path: Option<PathBuf>,
+    pub(super) mtime: Mutex<Option<std::time::SystemTime>>,
+}
+
+impl LockedKeyring {
+    /// Unlocks a keyring and validates it
+    pub async fn unlock(self, secret: Secret) -> Result<Keyring, Error> {
```
bilelmoussaoui: Given that we also have a Keyring inside `api::Keyring`, maybe we should rename this one to UnlockedKeyring.

client/src/file/locked_item.rs
```diff
@@ -0,0 +1,16 @@
...
+#[derive(Clone, Debug)]
+pub struct LockedItem {
+    pub(crate) inner: api::EncryptedItem,
+}
+
+impl LockedItem {
+    // TODO should return the original item if we fail to decrypt it.
+    pub fn unlock(self, key: &Key) -> Result<Item, file::Error> {
+        self.inner.decrypt(key)
+    }
```
bilelmoussaoui: I guess that is because we consume self. Why would we want to keep the original locked item if we can't even decrypt it? how can we return both the file::Error and the original item?

server/src/item.rs
bilelmoussaoui: If so, we should do the same for the keyring type.

bilelmoussaoui (commented):
I think the approach here looks good. Just some minor things here & there. 

We definitely need a test case to ensure that
- Opening a locked keyring and calling unlock with the correct secret gives you an unlocked keyring.
- Same for items

client/src/file/locked_keyring.rs
A6GibKm: My thoughts exactly, i opted for this to reduce the diff, i can either do a followup or put a commit at the end

server/src/item.rs
bilelmoussaoui: Renamed both in the latest commits.

client/src/file/locked_keyring.rs
bilelmoussaoui: Renamed in the latest commits.

client/src/file/locked_item.rs
A6GibKm: Yes, we don't want to lose the original item, but considering that there might be no way to unlock it after that point it might be OK to just lose it?

A6GibKm: One thing is that if there is an item in a object path, it should be stable, this might be a reason to keep it around even if one cannot do anything with it.

bilelmoussaoui: Does LockedItem implements Clone? i guess it is fine :)

client/src/file/locked_keyring.rs
A6GibKm: The migration step should be moved for a time after we have access to the password, or we somehow prompt the user for password for the migration+unlock.

bilelmoussaoui: I mean, this is for backward compatibility purposes only, we can take care of that on the server side.

bilelmoussaoui: We can for now provide a method that also does the migration from v0 to v1 that takes the Secret. The server side can use that instead of the common use case, the file backend backed by the portal.

client/src/file/locked_item.rs
bilelmoussaoui: No longer relevant for now

bilelmoussaoui:
Merged to main in some form

PR #279, opened 2025-07-25
client: Stop giving mac as Zeroizing

A6GibKm:
A mac does not contain sensitive data (no more than a hash at least), it does not need zeroizing on drop.


ueno (approved):
LGTM! I suspect that we could hide the content of MAC behind a newtype to make sure the values are always compared in a constant-time manner (i.e. using `memcmp::eq`), or at least suggesting that, though it might require changes in the other parts of the API.

PR #281, opened 2025-07-25
client: encrypted_item: Remove unnecesary allocs

A6GibKm:
split_off allocates a new Vec for its return value and it needs the vec to be mutable.


ueno (approved):
LGTM

PR #282, opened 2025-07-26
client: file: Add {,de}serialization test for Item

A6GibKm:
We also add the first commit from https://github.com/bilelmoussaoui/oo7/pull/268 for no particular reason that to make it smaller.

### DIFF
--- a/client/src/file/item.rs
+++ b/client/src/file/item.rs
@@ -190,3 +190,63 @@ impl TryFrom<&[u8]> for Item {
         Ok(item)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_file_item_serialization() {
+        let key = Key::new(vec![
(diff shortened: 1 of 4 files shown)


client/src/file/item.rs
```diff
@@ -184,3 +190,63 @@ impl TryFrom<&[u8]> for Item {
...
+            ]
+        );
+
+        let decrypted = encrypted.decrypt(&key).unwrap();
+
+        // The decrypted item matches the original one but with the content-type
+        // attribute set.
+        item.attributes.insert(
+            crate::CONTENT_TYPE_ATTRIBUTE.to_string(),
+            AttributeValue::from("application/octet-stream"),
```
bilelmoussaoui: Should use ContentType::Blob, maybe you could add a From<ContentType> for AttributeValue or so.

bilelmoussaoui:
Otherwise lgtm

client/src/file/item.rs
A6GibKm: Done, but I don't think is it interesting enough to add a From.

PR #284, opened 2025-07-28
client: Add new Mac struct for constant time eq

A6GibKm:
Prevents [timing attacks](https://en.wikipedia.org/wiki/Timing_attack) that can be used to guess possible macs. The subtle crate is already a dependency of digest.

### DIFF
--- a/client/src/crypto/native.rs
+++ b/client/src/crypto/native.rs
@@ -132,11 +132,11 @@ pub(crate) fn compute_mac(data: impl AsRef<[u8]>, key: &Key) -> Result<crate::Ma
 pub(crate) fn verify_mac(
     data: impl AsRef<[u8]>,
     key: &Key,
-    expected: impl AsRef<[u8]>,
+    expected_mac: impl AsRef<[u8]>,
 ) -> Result<bool, super::Error> {
     let mut mac = MacAlg::new_from_slice(key.as_ref()).unwrap();
     mac.update(data.as_ref());
-    Ok(mac.verify_slice(expected.as_ref()).is_ok())
+    Ok(mac.verify_slice(expected_mac.as_ref()).is_ok())
 }
 
 pub(crate) fn verify_checksum_md5(digest: impl AsRef<[u8]>, content: impl AsRef<[u8]>) -> bool {
--- a/client/src/crypto/openssl.rs
+++ b/client/src/crypto/openssl.rs
@@ -153,9 +153,12 @@ pub(crate) fn compute_mac(data: impl AsRef<[u8]>, key: &Key) -> Result<Mac, supe
 pub(crate) fn verify_mac(
     data: impl AsRef<[u8]>,
     key: &Key,
-    expected: impl AsRef<[u8]>,
+    expected_mac: impl AsRef<[u8]>,
 ) -> Result<bool, super::Error> {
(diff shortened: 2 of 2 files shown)


ueno (dismissed):
Looks nice!

A6GibKm:
Thanks!

There is a small (public) api wart I just noticed. Ill fix it tomorrow

client/src/mac.rs
```diff
@@ -0,0 +1,41 @@
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "native_crypto")]
+use subtle::ConstantTimeEq;
+use zbus::zvariant::Type;
+
+#[derive(Deserialize, Serialize, Type, Debug, Clone)]
+pub struct Mac(Vec<u8>);
+
+impl Mac {
+    pub fn new(inner: Vec<u8>) -> Self {
```
bilelmoussaoui: Is this needed? If yes, make it a From impl

client/src/lib.rs
```diff
@@ -18,13 +18,15 @@ use std::collections::HashMap;
...
 mod key;
+mod mac;
 mod migration;
 
 #[cfg(feature = "unstable")]
 #[cfg_attr(docsrs, doc(cfg(feature = "unstable")))]
 pub use key::Key;
 #[cfg(not(feature = "unstable"))]
 pub(crate) use key::Key;
+pub use mac::Mac;
```
bilelmoussaoui, line 29: should be pub(crate) no?

client/src/mac.rs
A6GibKm: I made it `pub(crate)`. We can use `Mac(bytes)` if you prefer. 

The wart I mentioned is that this is not intended to be constructed by the user (since it requires length verification), therefore it needs a public method to be compared against a slice.

client/src/lib.rs
A6GibKm: No, it is exposed by the `AsAttributes::hash`  (pub) trait method :(

client/src/mac.rs
```diff
@@ -0,0 +1,43 @@
...
+
+#[derive(Deserialize, Serialize, Type, Debug, Clone)]
+pub struct Mac(Vec<u8>);
+
+impl Mac {
+    pub(crate) fn new(inner: Vec<u8>) -> Self {
+        Mac(inner)
+    }
+
+    #[cfg(feature = "native_crypto")]
```
bilelmoussaoui: Can you move this inside the function?

client/src/mac.rs
```diff
@@ -0,0 +1,43 @@
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "native_crypto")]
+use subtle::ConstantTimeEq;
+use zbus::zvariant::Type;
+
+#[derive(Deserialize, Serialize, Type, Debug, Clone)]
+pub struct Mac(Vec<u8>);
```
bilelmoussaoui, line 10: Can you add a single line comment about this?

A6GibKm: OK

client/src/mac.rs
A6GibKm: Sure, however it requires a dummy {}

A6GibKm:
Added documentation.

PR #301, opened 2025-09-15
service: Refactor connection build and service.collection field set

warusadura:
For some reason, even though the call to `Service.connection.set()` succeeds, afterwards calls to `Service.connection.get()` always return None.

This fixes this issue.

### DIFF
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -298,27 +298,28 @@ impl Service {
     pub async fn run(secret: Option<Secret>, request_replacement: bool) -> Result<(), Error> {
         let service = Self {
             collections: Default::default(),
-            connection: OnceLock::new(),
+            connection: Default::default(),
             sessions: Default::default(),
             session_index: Default::default(),
             prompts: Default::default(),
             prompt_index: Default::default(),
         };
 
-        let connection = zbus::connection::Builder::session()?
+        let connection_builder = zbus::connection::Builder::session()?
             .allow_name_replacements(true)
             .replace_existing_names(request_replacement)
             .name(oo7::dbus::api::Service::DESTINATION.as_deref().unwrap())?
             .serve_at(
                 oo7::dbus::api::Service::PATH.as_deref().unwrap(),
                 service.clone(),
-            )?
-            .build()
-            .await?;
+            )?;
 
-        service.connection.set(connection.clone()).unwrap();
-        let object_server = connection.object_server();
+        service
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
> For some reason, even though the call to Service.connection.set() succeeds, afterwards calls to Service.connection.get() always return None.

Where that happens?

bilelmoussaoui:
Can you get the backtrace?

bilelmoussaoui:
That is not the actual fix though. If you look at https://github.com/bilelmoussaoui/oo7/blob/main/server/src/service.rs#L308-L317 we do serve the Service as part of creating the connection, which means the methods can be called at that point of time, like Lock/Unlock although the connection is not yet set on the field.

So the solution is to move the serve_at after setting the connection field.

warusadura:
Yeah, understood. Thanks @bilelmoussaoui ! Will update the pr.

warusadura:
> Have you tested your changes ? does it work as expected now?

Yes. No more, `Option::unwrap()` on a `None` value` error. Lock method call exists nicely.

server/src/service.rs
```diff
@@ -309,15 +309,18 @@ impl Service {
...
             .await?;
 
         service.connection.set(connection.clone()).unwrap();
+
         let object_server = connection.object_server();
+        object_server
+            .at(
+                oo7::dbus::api::Service::PATH.as_deref().unwrap(),
+                service.clone(),
+            )
```
bilelmoussaoui: There is a difference in behaviour though, serve_at when called the moment of building the connection, will replace the existing instance of the served interface. Where ObjectServer::at doesn't. So basically the request_replacement no longer does what it is supposed to do.

warusadura: oh! I didn't know that, thanks! Let me double check on this.

server/src/service.rs
```diff
@@ -298,27 +298,31 @@ impl Service {
...
+            .name(oo7::dbus::api::Service::DESTINATION.as_deref().unwrap())?;
+
+        service
+            .connection
+            .set(
+                connection_builder
+                    .serve_at(
+                        oo7::dbus::api::Service::PATH.as_deref().unwrap(),
+                        service.clone(),
+                    )?
```
bilelmoussaoui: why is this not done above?

warusadura: Yeah.... I can do that :)

Issue #302, opened 2025-09-15
Server issues/bugs


warusadura:
- [x] The prompt (for lock and unlock) is not working.
 - [x]  Sessions aren't properly closing when the clients leave.

warusadura:
Yes, the first part is complete.

PR #333, opened 2025-10-19
server: Fix sessions are not closing when clients leave

warusadura:
Call to message.body().deserialize() fails with
"Invalid unique name" error when trying to deserialize an empty string ("") as an OwnedUniqueName. Changing the type of new_owner to String fixes this issue.

This completes: #302

### DIFF
--- a/server/src/service.rs
+++ b/server/src/service.rs
@@ -516,10 +516,9 @@ impl Service {
             .build();
         let mut stream = zbus::MessageStream::for_match_rule(rule, self.connection(), None).await?;
         while let Some(message) = stream.try_next().await? {
-            let Ok((_name, old_owner, new_owner)) =
-                message
-                    .body()
-                    .deserialize::<(String, OwnedUniqueName, OwnedUniqueName)>()
+            let Ok((_name, old_owner, new_owner)) = message
+                .body()
+                .deserialize::<(String, OwnedUniqueName, String)>()
             else {
                 continue;
             };


server/src/service.rs
```diff
@@ -516,10 +516,9 @@ impl Service {
             .build();
         let mut stream = zbus::MessageStream::for_match_rule(rule, self.connection(), None).await?;
         while let Some(message) = stream.try_next().await? {
-            let Ok((_name, old_owner, new_owner)) =
-                message
-                    .body()
-                    .deserialize::<(String, OwnedUniqueName, OwnedUniqueName)>()
+            let Ok((_name, old_owner, new_owner)) = message
+                .body()
+                .deserialize::<(String, OwnedUniqueName, String)>()
```
bilelmoussaoui, line 521:
```suggestion
                .deserialize::<(String, OwnedUniqueName, Option<OwnedUniqueName>)>()
```
doesn't help?

warusadura: No, this doesn't help. And get we also get `Signature mismatch`.

bilelmoussaoui: Not even Optional?

warusadura: Ah.. my roaming connection is slow. Sorry, what do you mean?

bilelmoussaoui: https://docs.rs/zbus/latest/zbus/fdo/struct.NameOwnerChangedArgs.html

warusadura: No, it doesn't work.

```
Signature mismatch: got `(sss)`, expected `(ssms)`
[server/src/service.rs:527:25] "{}" = "{}"
[server/src/service.rs:527:25] err = Variant(
    SignatureMismatch(
        Structure(
            Dynamic {
                fields: [
                    Str,
                    Str,
                    Str,
                ],
            },
        ),
        "`(ssms)`",
    ),
)

```

bilelmoussaoui: What changes did you try

PR #338, opened 2025-10-24
Add PAM support

### LINKED ISSUE #247: PAM integration
So that the keyring can be auto-unlocked when logging in

bilelmoussaoui:
Closes #247 

This is done by adding a PAM cdylib project that receives the secret once the user is authenticated and send it over to the running daemon if any over a socket. The server listens on that socket when started and tries to unlock the collections with the received secret if possible.

This is has not been tested on an actual system yet.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -28,6 +28,31 @@ encrypted home with [systemd-homed](https://systemd.io/HOME_DIRECTORY/).
 
 See the manual page `systemd.exec(5)` for more details.
 
+## Debugging and Logs
+
+When running `oo7-daemon` as a systemd user service, logs can be viewed using `journalctl`:
+
+```sh
+journalctl -b -u oo7-daemon --user
+```
+
+The daemon supports verbose logging via the `--verbose` flag, which enables debug-level logs:
+
+```sh
+# Edit the service file to add --verbose
+systemctl --user edit oo7-daemon.service
+```
+
+PAM module logs are also available in the system journal and can be filtered:
+
+```sh
+# View all logs from the current session (includes PAM module)
+journalctl -b --user
+
+# Search for PAM-related entries
+journalctl -b --user | grep pam_oo7
+```
+
 ## License
 
 The project is released under the MIT license.


pam/README.md
```diff
@@ -0,0 +1,59 @@
...
+password   required     pam_unix.so
+session    required     pam_unix.so
+session    optional     pam_systemd.so
+```
+
+## Configuration
+
+### Environment Variables
+
+- `OO7_PAM_SOCKET`: Path to the Unix domain socket for daemon communication
```
A6GibKm, line 75: The README misses a short section explaining how this all fits together. Who owns the socket (user, group, permission bits, process), who writes to it and what do they write to it, who reads it, etc.

pam/meson.build
```diff
@@ -0,0 +1,25 @@
+project(
+  'pam_oo7',
+  'rust',
+  version: '0.6.0',
+  meson_version: '>= 0.61.0',
```
A6GibKm, line 5: By now I would require 1.7.0 on new projects.

pam/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+    if user_ptr.is_null() {
+        return Err(PAM_SYSTEM_ERR);
+    }
+
+    let user_cstr = unsafe { CStr::from_ptr(user_ptr) };
+    Ok(user_cstr.to_string_lossy().into_owned())
+}
+
+/// Get the authentication token
+unsafe fn get_auth_token(pamh: *mut pam_handle_t) -> Result<Zeroizing<Vec<u8>>, c_int> {
```
A6GibKm, line 44: Perhaps, in a followup, but the error should have a wrapper enum

pam/src/protocol.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+impl PamResponse {
+    #[cfg(test)]
+    pub fn to_bytes(&self) -> Result<Vec<u8>, zvariant::Error> {
+        let ctxt = Context::new_dbus(zvariant::LE, 0);
+        to_bytes(ctxt, self).map(|data| data.to_vec())
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, zvariant::Error> {
```
A6GibKm: I think this should take `Vec<u8>` to take ownership of the data, but I would need to double check zvariant.

server/src/pam_listener.rs
```diff
@@ -0,0 +1,191 @@
...
+        let collections = self.service.collections.lock().await;
+
+        for (_path, collection) in collections.iter() {
+            if collection.is_locked().await {
+                tracing::debug!("Attempting to unlock collection: {}", collection.path());
+
+                // Try to unlock with the provided secret
+                if let Err(e) = collection.set_locked(false, Some(secret.clone())).await {
+                    collection.dispatch_items().await?;
+                    tracing::debug!("Failed to unlock collection {}: {}", collection.path(), e);
```
A6GibKm, line 152: Did you pass this through clippy? I think it complains nowadays if variables (like `e` here) are not inlined in the macro.

server/src/pam_listener.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        let secret = Secret::from(message.secret);
+        self.user_secrets
+            .write()
+            .await
+            .insert(message.username.clone(), secret.clone());
+
+        // Try to unlock the default collection with this secret
+        match self.try_unlock_collections(&secret).await {
+            Ok(_) => {
```
A6GibKm, line 119: Does it make sense to keep listening to the socket after we already unlocked the keyring?

pam/meson.build
bilelmoussaoui: That is what we use everywhere else, we can bump it in a separate PR.

pam/src/lib.rs
bilelmoussaoui: Given that we only have two calls where we use those FFI constants, maybe it is okay to keep as is for now. Can be improved in the future if needed.

pam/src/protocol.rs
bilelmoussaoui: It doesn't need to, borrowing is fine here unless you have a reason for doing so that i didn't thought of

server/src/pam_listener.rs
bilelmoussaoui: Yes I did.

server/src/pam_listener.rs
bilelmoussaoui: My initial thought was to keep listening in case the user logs out and then logs in again or other user authentication operations that might trigger PAM. It is okay to keep around i would say

PR #379, opened 2026-01-18
Update ashpd to test reduced deps

bilelmoussaoui:


Cargo.toml
```diff
@@ -26,13 +26,13 @@ exclude = ["org.freedesktop.Secrets.xml"]
 
 [workspace.dependencies]
 zvariant = { version = "5.8", default-features = false, features = ["gvariant", "serde_bytes"]}
-ashpd = {version = "0.12", default-features = false}
+ashpd = {git = "https://github.com/bilelmoussaoui/ashpd", default-features = false}
```
A6GibKm: cargo-deny fails because this is missing a `version = "0.13"`.

bilelmoussaoui: yes, i won't merge this until i release a new version of ashpd

A6GibKm: I mean yes, but it can also be added to this branch for the sake of seeing if CI passes.

PR #495, opened 2026-06-21
server/prompter: Set cli prompter to use non-blocking sockets

bilelmoussaoui:


A6GibKm (approved):
Fixes `oo7-cli unlock` when there is no session running.

PR #515, opened 2026-07-09
cargo: Bump msrv to 1.95

A6GibKm:
git-credentials-oo7 uses a if-let guard which were [introduced](https://blog.rust-lang.org/2026/04/16/Rust-1.95.0/#if-let-guards-in-matches) in 1.95.


A6GibKm:
Will fix ci in a bit. @bilelmoussaoui the alternative would be to remove the single use of the let if and avoid the bump wdyt?

bilelmoussaoui:
Doing the bump is fine, we will do a new api breaking release anyways

PR #517, opened 2026-07-13
client: Accept MIME parameters in content types

caniko:
`ContentType` currently accepts only the two bare media types. That rejects valid Secret Service values with MIME parameters: the specification itself uses [`text/plain; charset=utf8`](https://specifications.freedesktop.org/secret-service/latest/types.html#type-Secret), and [go-keyring sends that exact value](https://github.com/zalando/go-keyring/blob/a8cdfe320cc8bc0534c895a648dc9214715a9da5/secret_service/secret_service.go#L34-L40).

This keeps the existing Text/Blob model, but compares only the base media type. Matching is trimmed and case-insensitive, outgoing serialization stays unchanged, and unknown base types still report the original full value.

I hit this through kubelogin: writing its token cache failed before kubectl could return cluster credentials.

Tested with:

- the full client test suite with native crypto
- the full client test suite with OpenSSL crypto
- a go-keyring Set/Get/Delete round trip against oo7

### DIFF
--- a/client/src/secret.rs
+++ b/client/src/secret.rs
@@ -34,10 +34,18 @@ impl FromStr for ContentType {
     type Err = String;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s {
-            "text/plain" => Ok(Self::Text),
-            "application/octet-stream" => Ok(Self::Blob),
-            e => Err(format!("Invalid content type: {e}")),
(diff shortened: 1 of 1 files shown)


caniko:
Tested on my own NixOS flake deployed with Cosmic DE

bilelmoussaoui:
thank you!

PR #521, opened 2026-07-14
Server: Deduplicate strings

tragivicky:
These strings are exactly the same as in `/server/src/gnome/prompter.rs` file, except for the brackets used. Change the strings to use same brackets. This will allow gettext to deduplicate them.


bilelmoussaoui:
Thanks!

tragivicky:
Sigh, I didn't checked properly and I introduced new string… Oops😅️ Will follow up with another MR, sorry for my carelessness…

PR #522, opened 2026-07-14
Server: Properly deduplicate string

tragivicky:
In #521 I accidentally introduced new string. This commit fixes this.

### DIFF
--- a/server/src/gnome/prompter.rs
+++ b/server/src/gnome/prompter.rs
@@ -178,7 +178,7 @@ impl Properties {
             message: Some(gettext("Authentication required")),
             description: Some(
                 formatx!(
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
Can we use the same quotes style everywhere?

tragivicky:
Sure. I'm in bed right now tho, so will do it in the morning, unless someone wants to force push to my branch, then feel free to do so. The other string that would need fixing is in `/server/src/gnome/prompter.rs`.

tragivicky:
I tried to fix this, but I have absolutely 0 idea why github feels the need to include completely unrelated commit here. I didn't even touched it, so why is this shown here?

bilelmoussaoui:
your changes contain an unrelated commit, please rebase

PR #529, opened 2026-07-16
capability: Clarify what insecure memory means

A6GibKm:


server/src/capability.rs
```diff
@@ -122,12 +122,12 @@ pub fn drop_unnecessary_capabilities() -> Result<(), rustix::io::Errno> {
             )?;
         }
         CapabilityState::None => {
-            tracing::warn!("No process capabilities, insecure memory might get used");
```
bilelmoussaoui, line 125: Maybe we should turn this into debug or info ? Not sure though

A6GibKm: I am not sure yet. Ideally this is only a warning if it is actionable and it seems it is not if the binary is run as the user.

bilelmoussaoui: could you reformat so the CI passes

A6GibKm: Will do today, but maybe wait untll i can test the setcap thingy, if it works as I hope we can keep the warn: it is ok as it would indicate a packaging issue.

A6GibKm:
So:
```
$ id --user
1000
$ id --group
1000
$ sudo install --mode 755 --owner root --group root /path/to/built/oo7-daemon /usr/local/libexec/oo7-daemon
$ /usr/local/libexec/oo7-daemon --verbose
2026-07-17T16:17:22.925954Z DEBUG oo7_daemon: Running in verbose mode
2026-07-17T16:17:22.925983Z  WARN oo7_daemon::capability: No process capabilities, insecure memory might get used
$ /sbin/setcap 'cap_ipc_lock=ep' /usr/local/libexec/oo7-daemon
$ /sbin/getcap /usr/local/libexec/oo7-daemon
/usr/local/libexec/oo7-daemon cap_ipc_lock=ep
$ /usr/local/libexec/oo7-daemon --verbose
2026-07-17T16:18:42.038410Z DEBUG oo7_daemon: Running in verbose mode
2026-07-17T16:18:42.039138Z  INFO oo7_daemon::capability: Successfully locked all memory pages
```

when installing the binary, caps can be added and we are able to lock the memory. This could be done on all binaries we ship. There is also the `i` flag for 'Inheritable' capabilities (see `capabilities(7)`) but I am not sure if we do any `execve(2)` call (probably not).

server/src/capability.rs
```diff
@@ -122,12 +122,17 @@ pub fn drop_unnecessary_capabilities() -> Result<(), rustix::io::Errno> {
...
+            // NOTE: For uses that landed here. The binary is probably missing
```
bilelmoussaoui: Users

bilelmoussaoui (dismissed):

PR #537, opened 2026-07-23
client: Accept `application/json` as `Text`

diwangs:
`ContentType` currently does not accept `application/json` MIME type, which breaks apps like [aws-vault](https://github.com/ByteNess/aws-vault). 

This PR adds `application/json` as an acceptable `Text` secret (since JSON is always a valid UTF-8) and keeps the Text/Blob distinction and MIME parameter parsing intact.

Tested with `aws-vault` in my NixOS setup:
- Adding secret with it no longer throws `aws-vault: error: add: Invalid content type: application/json`
- Exporting secret and requesting STS token through it works normally

A more generic fix for this type of errors might be to map every unrecognized content type to `Blob` instead of rejecting them. 

Let me know what you think!

### DIFF
--- a/client/src/secret.rs
+++ b/client/src/secret.rs
@@ -42,6 +42,7 @@ impl FromStr for ContentType {
 
         if media_type.eq_ignore_ascii_case("text/plain")
             || media_type.eq_ignore_ascii_case("text/utf8")
+            || media_type.eq_ignore_ascii_case("application/json")
         {
             Ok(Self::Text)
         } else if media_type.eq_ignore_ascii_case("application/octet-stream") {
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
Right, defaulting to blob seems sane but i kind of intentionally made it crash on early versions of the daemon to collect what mimetypes are out there in the wild so i can design the API better in the future.

diwangs:
Understandable. In that case, consider `application/json` as another data point. Do you have any comment about the PR itself?

bilelmoussaoui:
> Understandable. In that case, consider `application/json` as another data point. Do you have any comment about the PR itself?

that would require pulling serde_json, so for now let us keep it as is and we will see later.

PR #549, opened 2026-08-14
server: replace only items with identical attributes

### LINKED ISSUE #541: `CreateItem(replace=true)` deletes items whose attributes are a superset of the given ones
> **Disclosure: this report was researched and written by an AI agent** (Claude, run by the account owner). The reproducer below was executed against a throwaway keyring file and the output is copied verbatim, and the source references were read directly — but a human has not independently re-derived the analysis. Please treat the reasoning with the scepticism you'd apply to any unfamiliar contributor, and I'm happy to supply anything further.

## Summary

`CreateItem` with `replace = true` removes **every** item whose attributes are a *superset* of the attributes being stored, not just items with the *same* attributes. Storing one credential under a coarser attribute set silently destroys unrelated, more specific credentials.

The spec wording for the `replace` argument is:

> Whether to replace an item with the **same** attributes or not.

and oo7's own comment at `server/src/collection/mod.rs` agrees:

```rust
// Remove any existing items with the same attributes
if replace {
```

The implementation, however, matches on subset rather than equality.

## Reproducer

Against released **0.6.0** (`oo7-cli 0.6.0`), using an isolated keyring file so nothing real is at risk:

```console
$ echo "pw-alice" | oo7-cli -k repro.keyring -s $PW store "Alice entry" service=demo user=alice
$ echo "pw-bob"   | oo7-cli -k repro.keyring -s $PW store "Bob entry"   service=demo user=bob

$ oo7-cli -k repro.keyring -s $PW list
[Alice entry]
secret = pw-alice
attributes = {"user": "alice", "service": "demo"}
[Bob entry]
secret = pw-bob
attributes = {"service": "demo", "user": "bob"}

# store a third, unrelated credential under a COARSER attribute set
$ echo "pw-coarse" | oo7-cli -k repro.keyring -s $PW store "Coarse entry" service=demo

$ oo7-cli -k repro.keyring -s $PW list
[Coarse entry]
secret = pw-coarse
attributes = {"service": "demo"}
```

`pw-alice` and `pw-bob` are gone. No prompt, no warning, no count of what was removed.

### Control

Replacing with the *same* attribute set behaves correctly — only the intended item is replaced:

```console
$ echo "pw-alice2" | oo7-cli -k control.keyring -s $PW store "Alice v2" service=demo user=alice
$ oo7-cli -k control.keyring -s $PW list
[Bob]
secret = pw-bob
attributes = {"user": "bob", "service": "demo"}
[Alice v2]
secret = pw-alice2
attributes = {"service": "demo", "user": "alice"}
```

So the defect is specific to storing under a *coarser* attribute set, not to `replace` in general.

## Root cause

`matches_attributes` (`client/src/file/mod.rs`) iterates only the **query's** keys and requires each to be present and equal on the item. It never requires the item to lack additional attributes, so *query ⊆ item* counts as a match:

```rust
attributes.search_attributes().iter().all(|(k, value)| {
    item_attrs.get(k.as_str()).map(|v| v.as_ref()) == Some(value.as_str())
})
```

That predicate then backs the destructive path in both backends:

- **File backend** — `create_item(replace=true)` calls `keyring.remove_items(attributes)` (`client/src/file/unlocked_keyring.rs`), which does `self.items.retain(|e| !e.matches(attributes, key))` (`client/src/file/api/mod.rs`) — plural, every match.
- **Server / D-Bus** — `create_item` calls `search_items_with_key(&attributes)` and deletes every returned item (`server/src/collection/mod.rs`), which is the path real Secret Service clients hit.

## Why this is easy to miss

Subset matching looks deliberate, and for lookup it *is*: #233 reported that search failed when the searched attributes were a subset of the stored ones, and PR #234 changed matching to subset accordingly. That is the right semantics for `SearchItems` and I'm not suggesting reverting it.

The issue is that one predicate now serves two callers with opposite requirements. Lookup should be permissive — returning a superset of candidates is harmless. Deletion should be strict — over-matching destroys data. Widening the predicate for the first silently widened it for the second.

The existing `item_replacement_behavior` test doesn't catch it because every `create_item` call in it passes an *identical* attribute set; the coarse-vs-fine case is never exercised. (Minor: that test's comment `// Test replace=true on empty attributes (should just add)` doesn't match the code beneath it, which uses `unique_attrs` rather than empty attributes.)

## Suggested direction

Use exact attribute-set equality on the `replace` path while leaving `SearchItems` subset-based — e.g. a separate `matches_attributes_exact` (equal key sets *and* equal values) used by `remove_items` when invoked from `create_item`'s replace branch.

## Adjacent, possibly intentional

`Keyring::delete(attributes)` and `oo7-cli delete` share the same predicate, so `oo7-cli delete service=demo` empties every item under `service=demo` (verified). "Delete by query removes all matches" is a defensible design, so I'm flagging it rather than calling it a bug — but it's silent about how many items it removed, which makes an over-broad query hard to notice.

## Versions

- Reproduced with `oo7-cli 0.6.0` (distribution package).
- The relevant code is unchanged on `main` at `06b9cfe` — `matches_attributes`, `remove_items`, and both `create_item` replace paths are identical there.

julian-corbet:
`CreateItem(replace=true)` currently reuses the subset matcher needed by `SearchItems`. As a result, creating an item with a coarser attribute set removes every more-specific item that contains those attributes.

This change:

- adds exact matching for both encrypted and unencrypted file items
- uses it for single and bulk file-backend replacement
- uses it for the D-Bus server's in-memory item replacement
- keeps subset matching unchanged for searches and explicit query deletion
- adds regressions for the file, bulk, and D-Bus paths, including exact replacement after a coarse item is created

Fixes #541.

### DIFF
--- a/client/src/file/mod.rs
+++ b/client/src/file/mod.rs
@@ -93,6 +93,18 @@ impl Item {
             Self::Locked(locked) => locked.inner.matches(attributes, key),
         }
     }
+
+    /// Check if this item has exactly the given attributes.
+    pub fn matches_attributes_exact(
+        &self,
+        attributes: &impl AsAttributes,
+        key: Option<&Key>,
+    ) -> bool {
+        match self {
+            Self::Unlocked(unlocked) => unlocked.attributes() == &attributes.as_attributes(),
+            Self::Locked(locked) => locked.inner.matches_exact(attributes, key),
+        }
+    }
 }
 
 #[derive(Debug)]
--- a/client/src/file/mod.rs
+++ b/client/src/file/mod.rs
@@ -101,7 +101,7 @@ impl Item {
         key: Option<&Key>,
     ) -> bool {
         match self {
-            Self::Unlocked(unlocked) => unlocked.attributes() == &attributes.as_attributes(),
+            Self::Unlocked(unlocked) => unlocked.matches_exact(attributes),
             Self::Locked(locked) => locked.inner.matches_exact(attributes, key),
         }
     }
--- a/client/src/file/unlocked_item.rs
+++ b/client/src/file/unlocked_item.rs
@@ -62,6 +62,11 @@ impl UnlockedItem {
         &self.attributes
     }
 
+    /// Check whether the attribute maps match.
+    pub fn matches_exact(&self, attributes: &impl AsAttributes) -> bool {
+        self.attributes == attributes.as_attributes()
+    }
+
     /// Retrieve the item attributes as a typed schema.
     ///
     /// # Example
--- a/client/src/file/unlocked_keyring.rs
+++ b/client/src/file/unlocked_keyring.rs
@@ -465,10 +465,10 @@ impl UnlockedKeyring {
         let item = {
             let key = self.derive_key().await?;
             let mut keyring = self.keyring.write().await;
-            if replace {
-                keyring.remove_items(attributes, key.as_deref())?;
-            }
             let item = UnlockedItem::new(label, attributes, secret);
+            if replace {
+                keyring.remove_items_exact(item.attributes(), key.as_deref())?;
+            }
             let encrypted_item = item.encrypt(key.as_deref())?;
             keyring.items.push(encrypted_item);
             item
@@ -540,10 +540,10 @@ impl UnlockedKeyring {
         let _span = tracing::debug_span!("bulk_create", items_to_create = items.len());
 
         for (label, attributes, secret, replace) in items {
-            if replace {
-                keyring.remove_items(&attributes, key.as_deref())?;
-            }
             let item = UnlockedItem::new(label, &attributes, secret);
+            if replace {
+                keyring.remove_items_exact(item.attributes(), key.as_deref())?;
+            }
             let encrypted_item = item.encrypt(key.as_deref())?;
             keyring.items.push(encrypted_item);
         }
--- a/server/src/collection/mod.rs
+++ b/server/src/collection/mod.rs
@@ -308,7 +308,7 @@ impl Collection {
         // Remove any existing items with the same attributes
         if replace {
             let existing_items = self
-                .search_items_with_key(&attributes, key.as_deref())
+                .search_items_exact_with_key(&attributes, key.as_deref())
                 .await?;
             if !existing_items.is_empty() {
                 let mut items = self.items.lock().await;
@@ -524,6 +524,26 @@ impl Collection {
 
             // Use the oo7::file::Item's matches_attributes method
             if file_item.matches_attributes(attributes, key) {
+                matching_items.push(item_wrapper.clone());
+            }
+        }
+
+        Ok(matching_items)
+    }
+
+    async fn search_items_exact_with_key(
+        &self,
+        attributes: &HashMap<String, String>,
+        key: Option<&oo7::Key>,
+    ) -> Result<Vec<item::Item>, ServiceError> {
+        let mut matching_items = Vec::new();
+        let items = self.items.lock().await;
+
+        for item_wrapper in items.iter() {
+            let inner = item_wrapper.inner.lock().await;
+            let file_item = inner.as_ref().unwrap();
+
+            if file_item.matches_attributes_exact(attributes, key) {
                 matching_items.push(item_wrapper.clone());
             }
         }
--- a/server/src/collection/mod.rs
+++ b/server/src/collection/mod.rs
@@ -541,7 +541,9 @@ impl Collection {
 
         for item_wrapper in items.iter() {
             let inner = item_wrapper.inner.lock().await;
-            let file_item = inner.as_ref().unwrap();
+            let Some(file_item) = inner.as_ref() else {
+                continue;
+            };
 
             if file_item.matches_attributes_exact(attributes, key) {
                 matching_items.push(item_wrapper.clone());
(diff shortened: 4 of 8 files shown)


julian-corbet:
Validation update:

- The full fork CI run passes Check, Rustfmt, Clippy, Test Suite (all feature/backend combinations), and Meson: https://github.com/julian-corbet/oo7/actions/runs/31764506620
- `cargo-deny` is the only failing job. An untouched `main` baseline has the identical failure: https://github.com/julian-corbet/oo7/actions/runs/31764541659
- Both report the existing wildcard dependency, duplicate `syn` versions, and RUSTSEC-2026-0244; this patch does not touch dependencies.

The upstream PR workflows are still awaiting first-time-contributor approval.

client/src/file/mod.rs
```diff
@@ -93,6 +93,18 @@ impl Item {
...
     }
+
+    /// Check if this item has exactly the given attributes.
+    pub fn matches_attributes_exact(
+        &self,
+        attributes: &impl AsAttributes,
+        key: Option<&Key>,
+    ) -> bool {
+        match self {
+            Self::Unlocked(unlocked) => unlocked.attributes() == &attributes.as_attributes(),
```
bilelmoussaoui: I would add a similar helper to UnlockedItem for the sake of consistency

client/src/file/unlocked_keyring.rs
```diff
@@ -465,10 +465,10 @@ impl UnlockedKeyring {
         let item = {
             let key = self.derive_key().await?;
             let mut keyring = self.keyring.write().await;
+            let item = UnlockedItem::new(label, attributes, secret);
```
bilelmoussaoui, line 468: This change is unnecessary no?

client/src/file/unlocked_keyring.rs
```diff
@@ -540,10 +540,10 @@ impl UnlockedKeyring {
         let _span = tracing::debug_span!("bulk_create", items_to_create = items.len());
 
         for (label, attributes, secret, replace) in items {
```
bilelmoussaoui, line 542: Same here

server/src/collection/tests.rs
```diff
@@ -226,6 +226,73 @@ async fn create_item_with_replace() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+#[tokio::test]
+async fn create_item_with_replace_matches_attributes_exactly()
```
bilelmoussaoui: Remove the exactly from the name

bilelmoussaoui:
Both CI failures are unrelated to this one, so ignore them

client/src/file/unlocked_item.rs
```diff
@@ -62,6 +62,11 @@ impl UnlockedItem {
         &self.attributes
     }
 
+    /// Check if this item has exactly the given attributes.
+    pub fn matches_attributes_exact(&self, attributes: &impl AsAttributes) -> bool {
```
bilelmoussaoui: For encryptrd item, you usef matches_exact but not here. Make them consistent please. I would use matches_exact in both

server/src/collection/mod.rs
```diff
@@ -531,6 +531,26 @@ impl Collection {
...
+        &self,
+        attributes: &HashMap<String, String>,
+        key: Option<&oo7::Key>,
+    ) -> Result<Vec<item::Item>, ServiceError> {
+        let mut matching_items = Vec::new();
+        let items = self.items.lock().await;
+
+        for item_wrapper in items.iter() {
+            let inner = item_wrapper.inner.lock().await;
+            let file_item = inner.as_ref().unwrap();
```
bilelmoussaoui: I would avoid the unwrap here

client/src/file/mod.rs
julian-corbet: Done, thank you. `Item::matches_attributes_exact` now delegates to `UnlockedItem::matches_exact`, matching `EncryptedItem::matches_exact`.

client/src/file/unlocked_keyring.rs
julian-corbet: This ordering is needed because `UnlockedItem::new` adds the implicit `xdg:content-type` attribute. Replacement must compare the final attribute map that will be stored; otherwise, a repeated `replace = true` call without an explicit content type would not replace the previous item. The regression test covers that case.

client/src/file/unlocked_keyring.rs
julian-corbet: The same applies to bulk creation: the candidate is constructed first so replacement compares its final stored attributes, including the implicit content type.

server/src/collection/tests.rs
julian-corbet: Done, thank you. The test is now `create_item_with_replace_matches_attributes`.

client/src/file/unlocked_item.rs
julian-corbet: Done, thank you. Both concrete item types now use `matches_exact`.

server/src/collection/mod.rs
julian-corbet: Done, thank you. The helper now handles a missing inner item without unwrapping.

julian-corbet:
Review revision validation: [fork CI](https://github.com/julian-corbet/oo7/actions/runs/31798699376) passes Check, Rustfmt, Clippy, the full Test Suite, and Meson. `cargo-deny` reports the same wildcard dependency, duplicate `syn`, and RUSTSEC-2026-0244 errors as the [untouched `main` baseline](https://github.com/julian-corbet/oo7/actions/runs/31764541659); this revision introduces no new policy errors.

PR #550, opened 2026-08-16
Port to zgvariant

bilelmoussaoui:


Cargo.toml
```diff
@@ -53,7 +53,8 @@ tempfile = "3.26"
 tokio = { version = "1.52", default-features = false }
 tracing = "0.1"
 tracing-subscriber = "0.3"
-zbus = { version = "5.14.0", default-features = false }
-zbus_macros = { version = "5.11", features = ["gvariant"] }
+zbus = { version = "5.19", default-features = false }
+zgvariant = { version = "1.0", features = ["serde_bytes"] }
+zbus_macros = { version = "5.19", features = ["gvariant"] }
```
A6GibKm, line 58: Do we still need to explicitly declare this?

A6GibKm: It is needed, which is kinda annoying since we do not really use zbus_macros here.

PR #552, opened 2026-08-18
Port to zgvariant

zeenix:
Supersedes #550.

Moves oo7's GVariant usage off the (now-deprecated) `zvariant/gvariant` feature onto the
standalone [`zgvariant`](https://crates.io/crates/zgvariant) crate:

- `client` uses `zgvariant` for the keyring **file** format, and reaches the D-Bus zvariant
  through `zbus::zvariant` (the same instance ashpd re-exports).
- The pam module ⇄ daemon socket protocol is an internal wire format (not real D-Bus), so it now
  uses the GVariant encoding via `zgvariant` on both ends — the pam module and the daemon's
  `pam_listener`. This drops zvariant from the pam module entirely; it now depends only on
  `zgvariant`.

Requires zvariant 5.15.0 / zvariant_utils 4.2.0 — the coexistence fix that lets zvariant's D-Bus
format and zgvariant's GVariant format share one build — now published, so the tree resolves
purely from crates.io.

Verified locally: the whole workspace builds, the pam and daemon `pam_listener` round-trip tests
pass (they exercise the socket wire format end-to-end), and clippy + rustfmt are clean.

Generated by Claude Opus 4.8 (claude-opus-4-8).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1158,6 +1158,7 @@ dependencies = [
  "zbus",
  "zbus_macros",
  "zeroize",
+ "zgvariant",
  "zvariant",
 ]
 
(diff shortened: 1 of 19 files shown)


bilelmoussaoui:
> pam speaks the D-Bus wire format only (Context::new_dbus), so it keeps a direct,

pam can be switched to use gvariant format

zeenix:
> pam can be switched to use gvariant format

Done.

bilelmoussaoui:
lgtm, will merge once there is a release with the changes.

zeenix:
> lgtm, will merge once there is a release with the changes.

It's already there.

PR #558, opened 2026-08-24
server: detect graphical sessions under systemd --user compositors

### LINKED ISSUE #557: Secret Service falls back to the CLI prompter under compositors run as a systemd --user service
## Summary

`oo7-daemon` picks its prompter backend (GNOME/Plasma GUI vs CLI) based on `SessionType`, which is resolved via `logind`'s `GetSessionByPID` on the D-Bus peer's PID. Under a compositor that follows the systemd `--user` integration model (niri and sway both document this as the recommended setup), this always resolves to `Unspecified`, and every client gets the CLI prompter, even in a fully graphical session. In practice this makes the keyring unusable: `secret-tool`, `seahorse`, and anything else that expects a GUI prompt fails outright.

## Environment

- Arch Linux, `ly` as greeter, niri as compositor, started via `systemctl --user start niri.service` (niri's documented systemd integration path)
- oo7 0.7.0.alpha

## Reproduction

```
$ secret-tool store --label="test" service test-oo7 account test
Password:
secret-tool: org.freedesktop.DBus.Error.Failed: CLI prompter failed:
org.freedesktop.DBus.Error.NameHasNoOwner: Destination does not exist
```

The CLI prompter is attempted and fails because there is no controlling terminal available to it in this context (the client is a D-Bus service call, not an interactive shell), which surfaces as a `NameHasNoOwner` error rather than anything mentioning session detection.

## Root cause

`logind`'s `GetSessionByPID` only resolves a PID if it lives inside an actual login session's cgroup scope (`session-<N>.scope`). A compositor started as a systemd `--user` service is placed under `user@<uid>.service`'s own cgroup tree instead, outside of any session scope, and so is everything it spawns. `GetSessionByPID` fails for every such peer, so `SessionType::from_logind` returns `None` and the daemon falls back to treating the session as non-graphical, no matter what is actually running.

This is not specific to niri; it affects any compositor or session manager that uses the same systemd `--user` service pattern rather than being launched from inside a `session-<N>.scope` (e.g. via a display manager that execs it directly into the login session).

TheSkyentist:
Fixes #557

## Fix

`SessionType::detect` now falls back past `logind`, in order:

1. The peer's own `/proc/<pid>/environ`, since compositors export `WAYLAND_DISPLAY`/`DISPLAY` to the clients they spawn. This is often blocked by Yama's `ptrace_scope` (default 1 on many distros), since the daemon is never an ancestor of its D-Bus peers.
2. The systemd `--user` manager's own exported environment, which compositors following this integration model populate on startup (e.g. via `dbus-update-activation-environment`) and which isreadable live over D-Bus with no `ptrace_scope` restriction. This is coarser, since it is session-wide rather than peer-specific, so it is only used once `logind` gives no answer at all for the peer.

Each step is only consulted if the previous one could not place the peer in a session at all, so a definite non-graphical answer from `logind` (e.g. a real tty/SSH login) is still trusted as-is.

## Testing

Reproduced and confirmed fixed on Arch Linux with niri run as a systemd `--user` service and `ly` as greeter. Added unit tests for `from_environ`. `cargo test`, `cargo clippy --all-targets` clean.


server/src/session.rs
```diff
@@ -49,6 +49,75 @@ impl SessionType {
             _ => Self::Unspecified,
         })
     }
+
+    /// Fall back to the peer's own environment, for compositors run as a
+    /// systemd `--user` service (e.g. niri, sway), which `logind` can't
+    /// place in a session at all. Often blocked by Yama's `ptrace_scope`,
+    /// since the daemon isn't an ancestor of its peers; see
+    /// [`Self::from_systemd_user_environment`] for the last resort.
+    pub async fn from_environ(pid: u32) -> Option<Self> {
```
bilelmoussaoui: Doesn't need to be pub.

server/src/session.rs
```diff
@@ -49,6 +49,75 @@ impl SessionType {
...
+        let environ = tokio::fs::read(format!("/proc/{pid}/environ")).await.ok()?;
+        let has_non_empty_var = |name: &str| {
+            let prefix = format!("{name}=");
```
bilelmoussaoui: You could pass WAYLAND_DISPLAY=directly and avoid the allocation here?

server/src/session.rs
```diff
@@ -49,6 +49,75 @@ impl SessionType {
...
+    /// `dbus-update-activation-environment`), and it's readable over
+    /// D-Bus with no `ptrace_scope` restriction. Coarser than the other
+    /// checks since it's session-wide rather than peer-specific, so it's
+    /// only consulted once `logind` can't place the peer anywhere.
+    pub async fn from_systemd_user_environment() -> Option<Self> {
+        let connection = zbus::Connection::session().await.ok()?;
+        let manager = SystemdManagerProxy::new(&connection).await.ok()?;
+        let environment = manager.environment().await.ok()?;
+
+        let has_non_empty_var = |name: &str| {
```
bilelmoussaoui: This is duplicated between the two, just split it into a little helper and add some minimal tests for it?

server/src/session.rs
TheSkyentist: done in [80df49c](https://github.com/linux-credentials/oo7/pull/558/commits/80df49c2cfe3367e7ab135a5bb20397dc2ba94a9)

server/src/session.rs
TheSkyentist: done in [80df49c](https://github.com/linux-credentials/oo7/pull/558/commits/80df49c2cfe3367e7ab135a5bb20397dc2ba94a9)

server/src/session.rs
TheSkyentist: done in [80df49c](https://github.com/linux-credentials/oo7/pull/558/commits/80df49c2cfe3367e7ab135a5bb20397dc2ba94a9)

Issue #548, opened 2026-08-13
Design: preserve complete keyring state across daemon restarts


caniko:
## Problem

I hit a lifecycle gap while using oo7 with PAM.

PAM unlocked the Login collection, and `oo7-daemon-login` delivered the
password through its sealed memfd. Later, a package update restarted
`oo7-daemon` without ending the desktop session. The helper had already exited,
and no systemd credential was configured, so the new daemon loaded the
collection locked. The user had not logged out or requested `Lock()`.

I previously suggested retaining the helper memfd in systemd's file descriptor
store in [#506](https://github.com/linux-credentials/oo7/issues/506#issuecomment-5172034242).
After reviewing the wider lifecycle, I no longer think oo7 should retain the raw
login password for the whole session. The fdstore still looks like a suitable
optional transport, but it should carry keyring-specific restart state instead.

## Proposed ownership

Credential acquisition should remain outside oo7. PAM, the desktop,
credentiald, or another provider supplies a secret or key. oo7 validates it,
unlocks its own keyrings, and retains only the state needed to resume those
keyrings after its process restarts.

The split would be:

- The file backend owns validated key-based unlocking and should retain derived
  keys rather than source passwords.
- The server owns a versioned snapshot of persistent keyring and lock state.
- An optional Linux/systemd backend carries that snapshot through the service
  fdstore.
- Desktop and distro integration owns the logout boundary.

Non-systemd behavior should remain unchanged.

## First release contract

I think the first release advertising restart continuity should implement the
complete state model rather than a Login-only or collection-only subset:

- Every persistent encrypted and unencrypted file keyring is represented.
- Collection and per-item lock state is restored exactly.
- The raw PAM or prompt password is discarded after key derivation.
- Graceful and crash-triggered daemon restarts restore state.
- Explicit locks remain effective after restart.
- Malformed, stale, mismatched, or oversized state fails locked without partial
  restoration.
- Single-keyring and multi-keyring password changes have crash-consistent
  recovery.
- Restoration finishes before `org.freedesktop.secrets` is acquired.
- An explicit stop, logout, user-manager exit, or reboot discards state.

The work can still land as small preparatory PRs. The shipped service should not
enable continuity until the complete contract is tested.

## Key-based file API

`UnlockedKeyring` currently retains the source `Secret` and derives its `Key`
lazily. I propose deriving the key during unlock and retaining only the key.

The narrow API needed for this is:

[...]

PR #559, opened 2026-08-25
client: support already-derived file keys

caniko:
## Summary

Add file-backend APIs for callers that already hold derived key material:

- `LockedKeyring::{validate_key, unlock_with_key}`
- `UnlockedKeyring::{load_with_key, open_with_key, open_at_with_key}`

This also makes `oo7::Key` publicly nameable for those APIs.

## Behavior

Keys are checked against the active crypto backend's required key length before use.

`validate_key` mirrors `validate_secret`: it returns true for an empty keyring or when at least one item authenticates. `unlock_with_key` retains the existing corruption thresholds, so a partially corrupted keyring can validate but still return `PartiallyCorruptedKeyring`.

Non-empty unencrypted keyrings reject a direct key. Key-based `open` and `open_at` operate only on v1 paths and do not inspect or migrate v0 files; secret-based open retains legacy migration.

An exact-length `Key::new(...)` value is treated as caller-supplied direct key material, so the caller is responsible for its entropy.

## Scope

This PR does not add restart state, fdstore transport, PAM or server changes, D-Bus readiness changes, migration recovery, a generic credential provider, or an unchecked key API.

## Testing

Tested with both crypto backends, the complete Tokio client suites, async-std builds, workspace checks and clippy, rustfmt, rustdoc, cargo-deny, and the affected server and package builds. A disposable external async-std consumer also exercised key-only read, write, reload, and wrong-key rejection with both crypto backends without Tokio or `oo7-daemon` in its feature graph.

Refs #548.


client/src/file/locked_keyring.rs
```diff
@@ -100,52 +136,80 @@ impl LockedKeyring {
...
-                    n_valid_items += 1;
+            Some(Arc::new(key))
+        } else {
+            None
+        };
+
+        Ok(self.into_unlocked(key, Some(Arc::new(secret))))
+    }
+
+    fn validate_items(keyring: &api::Keyring, key: &Key) -> Result<(), Error> {
```
bilelmoussaoui: Should this be moved to api::Keyring type itself?

client/src/file/unlocked_keyring.rs
```diff
@@ -227,6 +233,17 @@ impl UnlockedKeyring {
         }
     }
 
+    async fn open_with_key_paths(
+        v1_path: PathBuf,
+        v0_path: PathBuf,
+        key: Key,
+    ) -> Result<Self, Error> {
+        if !v1_path.exists() && v0_path.exists() {
```
bilelmoussaoui: the error here is wrong

client/src/file/unlocked_keyring.rs
```diff
@@ -243,6 +260,18 @@ impl UnlockedKeyring {
         Self::open_with_paths(v1_path, v0_path, secret).await
     }
 
+    /// Open a named current-format keyring with an already-derived key.
+    ///
+    /// Unlike [`Self::open`], this cannot migrate a legacy keyring because the
```
bilelmoussaoui: why it would open a v0 if it cannot migrate it? so just skip that part no?

client/src/file/unlocked_keyring.rs
```diff
@@ -289,6 +318,21 @@ impl UnlockedKeyring {
         Self::open_with_paths(v1_path, v0_path, secret).await
     }
 
+    /// Open a named current-format keyring at a specific data directory with
+    /// an already-derived key.
+    ///
+    /// This does not migrate legacy keyrings; see [`Self::open_with_key`].
+    #[cfg_attr(feature = "tracing", tracing::instrument(skip(key), fields(data_dir = ?data_dir.as_ref())))]
+    pub async fn open_at_with_key(
```
bilelmoussaoui, line 308: same comment

client/src/key.rs
```diff
@@ -3,7 +3,11 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 
 use crate::{crypto, file};
 
-/// A key.
+/// Cryptographic key material.
+///
+/// File-keyring APIs accept already-derived values constructed with
+/// [`Self::new`]. The same type also exposes the low-level key-exchange helpers
```
bilelmoussaoui: I would drop the zvariant part from the docs, it is an implementation detail

client/src/file/unlocked_keyring.rs
caniko: Removed `LegacyMigrationRequiresSecret` and its display and test contract.

client/src/file/unlocked_keyring.rs
caniko: `open_with_key` now resolves only the v1 path and calls `load_with_key`; it no longer inspects v0.

client/src/file/locked_keyring.rs
caniko: Moved the pure item-counting and corruption-threshold logic into `api::Keyring`; `LockedKeyring` now delegates while retaining caller-specific diagnostics.

client/src/key.rs
caniko: Removed the zvariant wording; the type documentation now describes key material, file-key construction, and redacted debug output.

client/src/file/unlocked_keyring.rs
caniko: `open_at_with_key` now follows the same v1-only path and does not inspect v0.

